PR #373, opened 2011-05-27
Current attributes in models become previous attributes too early

curvedmark:
The resetting of previous attributes seems to happen too early, i.e., as soon as all event listeners are called, current attributes become previous attributes, which really should happen the next time set() is called, otherwise the following code fails:

```
var Model = Backbone.Model.extend({
    foo: function() {
        this.set({key: 'val2'});
        this.trigger('fooevent');
    }
});

var model = new Model({key: 'val1'});

model.bind('fooevent', function() {
    console.log(model.previous('key'));
});

model.foo();
```

The output should be `val1` and in backbone 0.3.3 it's `val2`

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -136,7 +136,7 @@
     this.cid = _.uniqueId('c');
     this.set(attributes, {silent : true});
     this._changed = false;
-    this._previousAttributes = _.clone(this.attributes);
+    this._attributes = _.clone(this.attributes);
     if (options && options.collection) this.collection = options.collection;
     this.initialize(attributes, options);
   };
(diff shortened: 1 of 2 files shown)


harshal-dhumal:
The problem still persist if we have more than 1 model properties. Please see below example.
```

var Model = Backbone.Model;
var model = new Model({key: 'val1', key1: 'key1'});
model.on('change', function(){
    console.log(this.previous('key'));
    console.log(this.previous('key1'));
});
model.set({key: 'val2'});
model.set({key1: 'key2'});

Actual output
val1
key1
val2
key1

Expected output
val1
key1
val1
key1
```

**Backbone version 1.1.2**

PR #1291, opened 2012-05-08
Removed the harmful instanceof operator in preference for duck typing. Fixes #1288

### LINKED ISSUE #1288: Multiple backbones inside Node makes instanceof fail
This just consumed the last 4 hours of my life... and boy was it the most crazy difficult thing to debug.

If we have two modules:
- Module A provides custom backbone.js models
- Module B provides custom backbone.js collections

Then we have our app which:
- Requires Module A
- Requires Module B
- Creates a model instance using a custom model from Module A
- Creates a collection instance using a custom collection from Module B
- Adds the model instance to the collection instance

Then the sky falls down and burns all the horrible plagues at once. Why? I know the answer, but I'll give you a gold star if you can guess it.

Here's the answer: because our lovely `model instanceof Model` inside `Backbone.Model.prototype._prepareModel` fails, as the models are using one backbone.js and the collections are using another!!!! Despite them being the same backbone.js version.

ARGH! That line needs to be better, perhaps check for the presence of a `this.isBackboneModel` variable defined by the `Backbone.Model.prototype`.

ARGHHHH. For now, it seems the only solution is to pass the Modules the Backbone.js module that we want them to use, rather than having them use their own local copies. Which is a real PIA. Which imho, the flag idea is a much better one.

Thoughts?

balupton:
- Fixes #1288
- Forked from the 0.9.2 tag
- Added `isBackboneWhatever: true` to prototypes, uses them for checks
- Changed jQuery instanceof check to `element.jquery` presence check
- No new tests, existing browser and rake tests pass
- Repo to reproduce the problem is here: https://github.com/balupton/backbone-instanceof-bug
- Repo to prove this fixes the problem is here: https://github.com/balupton/backbone-instanceof-bug/tree/ducktyped


mbrevda:
@jashkenas is there another way to confirm that something is a Backbone.X? Im having this issue with collections

akre54:
@mbrevda try `col instanceof Backbone.Collection`.

mbrevda:
@akre54 thats what I was trying, but being that I have two instances of Backbone (from two different modules) that wont work

akre54:
Well for starters... why do you have two instances of Backbone?

What are you attempting to accomplish with this check?

mbrevda:
Two modules: A, and B. Both require backbone. A creates a collection and passes it to B. It seems that `a.backbone.collection != b.backbone.collection`, hence I cannot do an instanceof test in be to ensure that a Backbone.Colleciton was passed in.

See also the ticket referenced above (#1288)

akre54:
Well that's the thing... did you actually read #1288? This is a dependencies issue, not a Backbone one. Fix your dependencies and the situation will resolve itself. Anything else is playing with fire.

mbrevda:
I did, I was just asking if there was another *safe* way about. How can I "fix my dependencies"?

akre54:
I'm not in a position to help debugging at the moment, but try googling or using stackoverflow. It's not as onerous a process as you might think. The core gist of it is that moduleA should be reliant on the same Backbone as moduleB, which means usually means making sure the version ranges in package.json are compatible. Clear your `node_modules` folder and try reinstalling.

mbrevda:
Thanks @akre54. I did that, I believe they're the same. According to #1288 absolute paths matter here, and without peer dependencies there is no way around that.

mbrevda:
As an aside (and for the sake of anyone else that ends up here), it would seem that [npm 3.0](http://blog.npmjs.org/post/110924823920/npm-weekly-5) will mitigate this issue

PR #1630, opened 2012-09-06
Allow an individual router to be stopped

mxriverlynn:
Hi all,

In my Backbone.Marionette add-on framework, I have the concept of modules that can be started and stopped. These modules represent areas of the application that may or may not need to be running at a given time, hence the ability to start and stop them. I also make heavy use of multiple routers - typically one router per functional area of the app, which often boils down to very small routers within my modules.

I've run in to a problem recently, where stopping one of my modules does not stop the routes from that module. Presently, there is no way for a router to be stopped, hence there is no way for me to stop the routes for a given module. This creates issues with routes being handled when they shouldn't be, etc.

This pull request is a basic implementation of allowing individual routers to be stopped, by adding a `stop` method to the router prototype, and by adding a `removeRoute` method to the Backbone.History prototype. This functionality will allow my routers to be started and stopped with my modules, allowing my application to behave correctly.

Things to note in this pull request:
- It's incomplete. I have not accounted for routes that are added via `router.route(...)` calls, yet
- It's slow (I think). I have not optimized the performance of removing an individual route from Backbone.history - but I also don't have any performance benchmarks on it yet.

I'm guessing, based on the work I've done already, that the way routes are stored in routers and in history is going to have to change slightly, to optimize the removal of routes for performance. I'm not 100% sure of this, though, as I would need to build some benchmarks for performance to see how slow this is.

I want to get feedback on this before I head down that path and put in all the rest of the work, though. Is a viable piece of functionality to include in the core of Backbone? What design considerations do I need to keep in mind when fixing the remaining items for this? Am I missing any items that need to be accounted for in my short list? 

Thanks

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -906,7 +906,7 @@
     //     });
     //
     route: function(route, name, callback) {
-      if (!_.isRegExp(route)) route = this._routeToRegExp(route);
+      if (!_.isRegExp(route)) route = routeToRegExp(route);
       if (!callback) callback = this[name];
       Backbone.history.route(route, _.bind(function(fragment) {
         var args = this._extractParameters(route, fragment);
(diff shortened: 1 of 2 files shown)


AndrewEastwood:
How can I handle such problem? I need to cancel some routes according to internal states.

akre54:
@AndrewEastwood you can `return false` from [`Router#execute`](http://backbonejs.org/#Router-execute)

PR #2340, opened 2013-03-06
Backbone assumes that Backbone.ajax will be asynchronous.

caiges:
Backbone assumes that Backbone.ajax will be an asynchronous method. In the case of a user supplied Backbone.ajax method that caches requests in memory (and returns cached results) or performs some other logic that returns immediately, the 'sync' event will be fired before the 'request' event. It seems logical that the 'request' event should be fired before the request is actually performed. As documented, the current implementation makes sense unless the scenario as described occurs. This might suggest that the 'request' event is indeed correct in its current form and that another event 'beforeRequest' might be more appropriate.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1434,8 +1434,8 @@
     }
 
     // Make the request, allowing the user to override any Ajax options.
-    var xhr = options.xhr = Backbone.ajax(_.extend(params, options));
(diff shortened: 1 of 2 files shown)


feast107:
123123

feast107:
123123

PR #2362, opened 2013-03-12
Allow array of event handlers in View#delegateEvents.

alexgb:
It can be useful to define multiple independent event handlers for a single event + selector combination. This helps promote reusable event handler methods, for example:

``` javascript

Backbone.View.extend({
  events: {
    "mouseover .map":    ["showTip", "moveTip"],
    "mousemove .map":    "moveTip",
    "mouseout .map":     "hideTip"
  }
});
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1297,8 +1297,9 @@
     //
     //     {
     //       'mousedown .title':  'edit',
-    //       'click .button':     'save'
-    //       'click .open':       function(e) { ... }
+    //       'click .button':     'save',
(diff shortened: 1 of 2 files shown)


jfbrennan:
I want to avoid this:
```
collection.on('sync', callback1)
collection.on('sync', callback2)
```
and this:
```
collection.on('sync', function() {
  callback1()
  callback2()
})
```
and just have this beauty right here:
```
collection.on('sync', [callback1, callback2])
```

PR #2378, opened 2013-03-15
Fix #2376 - Update docs for changedAttributes.

### LINKED ISSUE #2376: changedAttributes can unexpectedly return false
Per the documentation:

> changedAttributes model.changedAttributes([attributes])
> Retrieve a hash of only the model's attributes that have changed. Optionally, an external attributes hash can be passed in, returning the attributes in that hash which differ from the model. This can be used to figure out which portions of a view should be updated, or what calls need to be made to sync the changes to the server.

I was expecting to call `_.isEmpty` on the outcome of this for flow control.

braddunbar:
This updates the documentation, but for what it's worth I think that returning `{}` is a good deal less surprising.  Thoughts?

### DIFF
--- a/index.html
+++ b/index.html
@@ -1419,10 +1419,11 @@ book.on("change", function() {
     <p id="Model-changedAttributes">
       <b class="header">changedAttributes</b><code>model.changedAttributes([attributes])</code>
       <br />
-      Retrieve a hash of only the model's attributes that have changed. Optionally,
-      an external <b>attributes</b> hash can be passed in, returning
-      the attributes in that hash which differ from the model. This can be used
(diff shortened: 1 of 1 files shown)


armaanahluwalia:
This is really unexpected. In terms of API design this seems to be a poor choice since as mentioned before, changedAttributes implies it returns attributes that have changed. Anyone extending backbone has to handle around this weird edge case. Anyone looking to guard against the changedAttributes being empty should be a doing a call to hasChanged() first.

Consider someone iterating over the changedAttributes with something like _.each. They would have to add code to check if its false. 

Behaviour should be changed IMHO.

Issue #2345, opened 2013-03-08
.save() should return a rejected promise object for an invalid model


yuku:
Currently `Model#save` returns `false` if it fails validation.

``` javascript
var model = new MyModel({invalid: 'attribute'});
model.save().done(function() {...});
//=> Uncaught TypeError: Object #<error> has no method 'done' 
```

It should return a rejected promise object instead of `false` if it is available. (FYI: Zepto does not provide promise object by default)

In addition, the way to notice whether a fail callback is called because of invalid or error is needed.

``` javascript
model.save()
  .done(function () {
    // success
  })
  .fail(function (model, resp, options) {
    if (options.isInvalid) {
      // invalid
    }
  })
  .fail(function (model, resp, options) {
    if (!options.isInvalid) {
      // error
    }
  });
```

If this proposal makes sense, I will send a pull request.

PR #2489, opened 2013-04-16
Return a rejected promise instead of false on invalid model save

akre54:
From the discussion in #2345, returning false from a failed save prevents a clean chaining API (`model.save().done(...).fail(...)`). Using deferreds, we can reject a promise instead to avoid the discrepancy.

This pull doesn't introduce any new options vars (i.e. `isValid`), instead suggesting a separate event listener for the model's `invalid` event or querying `model.isValid()` directly within the failed handler. The `invalid` event will always be fired before the deferred is rejected and returned. The fail handler is fired for both failed validation as well as any ajax errors which makes it difficult to mutate the options object or keep the two callbacks' arguments consistent.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -519,7 +519,7 @@
 
       if (this.isNew()) {
         options.success();
-        return Backbone.Deferred().reject();
+        return Backbone.Deferred().resolve();
       }
       wrapError(this, options);
 
--- a/backbone.js
+++ b/backbone.js
@@ -519,7 +519,7 @@
 
       if (this.isNew()) {
         options.success();
-        return false;
+        return Backbone.Deferred().reject();
       }
       wrapError(this, options);
 
--- a/backbone.js
+++ b/backbone.js
@@ -445,7 +445,7 @@
     // If the server returns an attributes hash that differs, the model's
     // state will be `set` again.
     save: function(key, val, options) {
-      var attrs, method, xhr, attributes = this.attributes;
+      var attrs, method, xhr, dfd = new $.Deferred, attributes = this.attributes;
 
       // Handle both `"key", value` and `{key: value}` -style arguments.
       if (key == null || typeof key === 'object') {
@@ -461,9 +461,9 @@
       // `set(attr).save(null, opts)` with validation. Otherwise, check if
       // the model will be valid when the attributes, if any, are set.
       if (attrs && !options.wait) {
-        if (!this.set(attrs, options)) return false;
+        if (!this.set(attrs, options)) return dfd.rejectWith(this);
       } else {
-        if (!this._validate(attrs, options)) return false;
+        if (!this._validate(attrs, options)) return dfd.rejectWith(this);
       }
 
       // Set temporary attributes if `{wait: true}`.
--- a/backbone.js
+++ b/backbone.js
@@ -1206,7 +1206,9 @@
 
   // Set the default promises library to proxy to `$`.
   // Override this if you'd like to use a different library.
-  Backbone.Deferred = Backbone.$.Deferred;
+  Backbone.Deferred = function() {
+    return Backbone.$.Deferred.apply(Backbone.$, arguments);
+  }
 
   // Backbone.Router
   // ---------------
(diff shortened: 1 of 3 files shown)


Artazor:
I still think that this PR is independent of #3582 
The only thing that we should agree that 
`save` and `destroy` should return thenables.

Thus if we combine this PR and #3582 it will look like:
```JavaScript

Backbone.newDestroyed = function(model) {
    return Backbone.Deferred().resolve(model);
}

Backbone.notValidated = function(model) {
    return Backbone.Deferred().reject(model.validationError);
}
```

There also a question, should the obtained xhr be also wrapped into Backbone.Defered().resolve(xhr)? To unify all thenables?

jridgewell:
I see this as solving the root problem brought up by #3582. Additionally, it only adds one method onto the `Backbone` namespace, and it's a well-defined class (`deferred`s are easy to create from Promise libraries).

> There also a question: should the obtained xhr be also wrapped into Backbone.Defered().resolve(xhr)? To unify all thenables?

So that all thenable return values are the same class, yes we should. Imagine overriding `Backbone.Deferred` to use native promises:

```javascript
Backbone.Deferred = function() {
    var deferred = {};
    deferred.promise = new Promise(function(resolve, reject) {
        deferred.resolve = resolve;
        deferred.reject = reject;
    });
    return deferred;
};
```

I'd hate for some return values to be jQuery promises and some to be native promises.

- - -

As a side note, I don't think we should `return deferred.resolve()`. `#resolve` and `#reject` return a non-standard value (the deferred, vs `undefined`), and that the deferred object has `#then` is non-standard.

Artazor:
So why Deferred? Why not Backbone.Promise?
And use pretty sandard Backbone.Promise.resolve(value) and Backbone.Promise.reject(reason)
? 

There is widespread belief that you should avoid using deferreds at all.

jridgewell:
> So why Deferred? Why not Backbone.Promise?

Because jQuery provides a Deferred class, so we don't have to write a ton of extra code.

> There is widespread belief that you should avoid using deferreds at all.

Hm?

jashkenas:
Opening for more discussion ... but we'd still need a real patch+proposal.

Artazor:
I would strongly advocate for the `Backbone.Promise` that should provide only two static methods `resolve` and `reject`

```TypeScript
declare module Backbone {
    ...
    export interface Thenable<T> {
        then<U>(
             fulfilled: (result: T) => U | Thenable<U>,
             rejected: (reason: any) => U | Thenable<U>
        ): Thenable<U>
    }
    ...
    export class Model<TSync> {
          save(...): Thenable<TSync>
          fetch(...): Thenable<TSync>
          destroy(...): Thenable<TSync>
          sync(...): Thenable<TSync>
    }

    export class Collection<TModel extends Model, TSync> {
          fetch(...): Thenable<TSync>
          sync(...): Thenable<TSync>
    }

    export function sync<TModel, TOpts, TSync>(
          method: string, 
          model: TModel, 
          options: TOpts
     ): Thenable<TSync>

    export class Promise {
         // lets rely only on those two static methods:
         static resolve<T>(value: T: Thenable<T>): Thenable<T>;
         static reject<T>(reason: any): Thenable<T>;
    }
}
```
Inside you can use promises created by jQuery.Deferred by default or the same jqXHR
but Backbone itself should rely only on two standard static methods that are place under Backbone.Promise:  `resolve` and `reject`.

My motivation:
* Those who never want to override methods will get the same jqXHR and JQueryPromises by the default implementation. Note that we do not need to implement promises in Backbone, we will reuse jQuery Deferred.promise() or similar things
* Those who wnat to override this, probably want to use Promises/A+ implementation and would be able to do it with single:
`Backbone.Promise = MyFavoritePromiseClass`

akre54:
I agree fully with @Artazor here (and thanks for the detailed interface). Do you propose we wrap `sync` (or `ajax`) in `Backbone.Promise` or should the implementing user's `ajax` method be required to return a thenable?

It'd be awesome to see a pull with this implementation if you have time.

Issue #2627, opened 2013-06-20
Give model.set a parse option


etler:
The collection counterpart of set allows you to provide a "parse" option to run the parser function before processing the collection object. This is handy when you have bootstrap data that you want to build a collection from, as it wont go through the fetch function.

Is there a reason the model object doesn't provide the same option to parallel collection? It seems inconsistent to me that it is omitted.

Edit: Upon further inspection, reset does still respects the parse option.

PR #2636, opened 2013-06-25
Give model.set a parse option

etler:
Implementing #2627

To get the parse option into `model.set` I also created `options.defaults`  so the defaults can be applied after parsing. I'm not sure if we would want to support that, although I can see a use case for wanting to provide fallback values after creating a model the first time.

To prevent `options.defaults` from causing the side effect of re-applying defaults on merge in the `collection.set` method, I opted to extend the model constructor options when adding the defaults option. As a result of this, `_prepareModel` can no longer rely on the options side effect from `_validate` to access `options.validationError`, so I repopulate it instead. I think it's best to remove options side effects where possible as they're acting as a way to return extra data from a function, and as a result adds complexity to nested method calls and to the flow of data.

By moving parsing into `model.set`, I was also able to remove the `options._attrs` side effect hack for merging models. `options._attrs` was being used to return the post parsed attributes to `collection.set` for merging, but now that parsing is done within `model.set`, that is no longer necessary, and we can just send `model.set` the original copy of the attributes, and have it re-parse them there.

It would be better to find a way to do this without `options.defaults`, as that is out of the scope of this issue, but by doing things this way I was able to remove `options._attrs`, which to me seems worse than having `options.defaults`.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -431,7 +431,7 @@
       var model = this;
       var success = options.success;
       options.success = function(resp) {
-        if (!model.set(model.parse(resp, options), options)) return false;
+        if (!model.set(resp, options)) return false;
         if (success) success(model, resp, options);
         model.trigger('sync', model, resp, options);
       };


jclif:
:+1:

burak-son-sociomantic:
:+1:

burakson:
:+1:

PR #2816, opened 2013-10-16
Add `set` event triggered once when collection models are added or removed

jdudek:
Events `add` and `remove` are a great way to keep a view in sync with collection. A view can observe when collection's content changes and re-render itself. A model can be added only in the client or loaded from the server and this is transparent to the view.

However, `add` and `remove` are triggered for every single model added/removed, which results in render being called multiple times. This causes performance issues, as rendering can be a heavy operation.

I'm not the first one to encounter this problem:
* http://backbonefu.com/2013/08/handling-multiple-collection-add-models-in-a-batch/
* http://blog.pamelafox.org/2013/06/improving-backbone-app-performance.html

My suggestion is to add a new event `set` that is triggered every time models are added or removed, but just once during one operation. E.g. calling `add` with multiple models will trigger `add` event for every model and `set` event only once.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -634,24 +634,11 @@
 
     // Remove a model, or a list of models from the set.
     remove: function(models, options) {
-      var singular = !_.isArray(models);
+      var singular = !_.isArray(models), removed;
       models = singular ? [models] : _.clone(models);
       options || (options = {});
-      var i, l, index, model;
-      for (i = 0, l = models.length; i < l; i++) {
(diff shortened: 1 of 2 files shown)


jashkenas:
Going to merge ... but call it 'updated'.

methyl:
:+1:



Lucjan

caseywebdev:
Glad to see this eventually made it :wink:

PR #2931, opened 2013-12-19
Type check Collection#get arguments.

### LINKED ISSUE #2922: Model with id "undefined" returned by collection.get(obj)
@caseywebdev The `Collection::get` method defined by 01d9f0f9b1e85cb21a2940902bc8ba4280c8bca9 has this odd behavior when the collection contains a model with the id `"undefined"`: `collection.get(obj)` will return that model whenever `obj.id` is `undefined`. An odd edge case, I know, but believe it or not I ran into this in a production application... among other things, it means that

```
new Backbone.Collection([{id: 'undefined'}, {id: 'foo'}, {id: 'bar'}])
```

will create a collection with length `1`.

What was wrong with the [Backbone 1.0.0 definition](https://github.com/jashkenas/backbone/blob/1.0.0/backbone.js#L786) of `get`, which explicitly checked for this case?

```
return this._byId[obj.id != null ? obj.id : obj.cid || obj];
```

Related to #2847.

braddunbar:
Fix #2922.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -799,8 +799,8 @@
     // Get a model from the set by id.
     get: function(obj) {
       if (obj == null) return void 0;
-      if (_.isObject(obj)) return this._byId[obj.id] || this._byId[obj.cid];
-      return this._byId[obj];
+      if (_.isNumber(obj) || _.isString(obj)) return this._byId[obj];
+      return this._byId[obj.id] || this._byId[obj.cid];
     },
 
     // Get the model at the given index.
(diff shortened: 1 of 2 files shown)


dylanlingelbach:
@tnbKristi and I hit this today.  Would you be open to an approach that prevented this by checking on `set` vs `get`?  Given an id of `'undefined'` completely breaks the collection after an item with a key of `undefined` is added it seems ok to say ids of `'undefined'` are not allowed to be used on a model.

If that sounds promising I can code up a PR

PR #3049, opened 2014-03-07
prevent memory leaks with 'stopListening' and 'off'

smelnikov:
This PR fixes potential leaks when using `stopListening` and `off`, it addresses the following:

```javascript
var a = _.extend({}, Backbone.Events);
var b = _.extend({}, Backbone.Events);

b.on('y', fn);
a.listenTo(b, 'x', fn);
a.stopListening(b, 'x');
b = null; // b can't be garbage collected until a.stopListening() is called
```

```javascript
var a = _.extend({}, Backbone.Events);
var b = _.extend({}, Backbone.Events);
var c = _.extend({}, Backbone.Events);

a.listenTo(c, 'x', fn);
b.listenTo(c, 'x', fn);
c.off('x');
c = null; //c can't be garbage collected until 'stopListening()' is called on a and b
```

this implementation has a minimal impact on performance as it uses a counter to determine when a listener link can be safely broken.


smelnikov:
I've updated this pull request with a merge from master. I've actually ended up reverting the changes (but kept the tests) made in order to support #3226 as they were already fixed with this pull request (With no duplication in `listenToOnce` and `eventApi` required). This actually made this pull request have more line deletions than insertions within backbone.js.  

Is there a reason why we're not moving forward with this PR? Any concerns that I can address ?

jashkenas:
@braddunbar How do you feel about this now?

backbone.js
```diff
@@ -106,46 +95,50 @@
...
         var remaining = [];
         for (var j = 0, k = events.length; j < k; j++) {
           var event = events[j];
+          var listener = this._listeners && this._listeners[event.listenId];
           if (
             callback && callback !== event.callback &&
             callback !== event.callback._callback ||
-            context && context !== event.context
+            context && context !== event.context ||
+            name && name !== eventName
```
jridgewell: This `name && name !== eventName` condition should be taken care of by [L111](https://github.com/jashkenas/backbone/pull/3049/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR111)

smelnikov: Yup, good catch.

backbone.js
```diff
@@ -112,42 +101,46 @@
 
       // Remove all callbacks for all events.
       if (!name && !callback && !context) {
-        this._events = void 0;
+        var listenerIds = _.keys(this._listeners);
+        for (var i = 0, l = listenerIds.length; i < l; i++) {
```
akre54: On a related note... Should we be avoiding `_.keys` in `_.each`? What if we branched at `nativeKeys` vs `for ... in` in `each` to avoid the double loop that @stephanebachelier pointed out? @jdalton

jdalton: The advantage to using the util libs methods is it includes iteration fixes that you wouldn't get otherwise.

braddunbar:
I've been a bit under the weather this week, but I'll do my best to get around to this today!

jridgewell:
Closing due to https://github.com/jashkenas/backbone/pull/3455

PR #3056, opened 2014-03-11
Better documentation around the behaviour of `once` when multiple events are passed in

suprememoocow:
`EventEmitter.once` will fire once per event -- as opposed to once -- when multiple events are passed in using the space-separated syntax. This can be fairly counter-intuitive. This PR updates the documentation to make not of this.


akre54:
Merged in 24a514436751b9302bf01839b5089cc047f25463

PR #3073, opened 2014-03-18
Remove remaining jQuery from View tests

akre54:
Now that #3003 is merged, we need an easy way for alternative View implementations to run against our View conformance test. This pulls jQuery out of the View tests except in for checking `view.$el instanceof Backbone.$` (which all jQuery-like apis should support).

### DIFF
--- a/test/view.js
+++ b/test/view.js
@@ -289,29 +289,25 @@
     document.body.removeChild(el);
   });
 
-  test("custom events", 2, function() {
-    var count = 0;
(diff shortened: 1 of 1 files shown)


akre54:
@braddunbar want to take one final pass through this and then either merge or close? I'm fine either way.

wyuenho:
@akre54 Has PhantomJS 2 fixed this click event issue?

PR #3115, opened 2014-04-09
Fix minor typo: `messsages` --> `messages`

gasi:
See title :wink:


sderickson:
This typo lives on. Maybe needs to go into the master branch instead? Is that the source?

https://github.com/jashkenas/backbone/blob/gh-pages/index.html#L3396

gasi:
@sderickson Yes, that is correct. Should I submit another patch?

PR #3129, opened 2014-04-23
workaround for a bug in the android browser.

Breton:
When leaving a backbone app, and returning via the back button,
Android's browser fails to properly restore the value of
Backbone.history.history, turning it into a dangerous pointer into
apparently freed memory.
This addition patches the one part of Backbone most likely to run afoul
of this bug while hopefully not harming other browsers.


akre54:
Closing absent a use case. We can reopen if there's something here.

PR #3165, opened 2014-05-27
Only trigger `sort` event when Collection order changes.

### LINKED ISSUE #3158: Collection.sort always trigger event
When I call `myCollection.sort()` (after a model change in my collection for example), `myCollection` will emit a `sort` event, even if the order didn't change.

I don't think this event should be trigger if the order didn't change. 
But before sending a PR, I would like to understand why it could be use full to keep this event or if it could be an improvement.

akre54:
Fixes #3158

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -853,7 +853,7 @@
       if (!this.comparator) throw new Error('Cannot sort a set without a comparator');
       options || (options = {});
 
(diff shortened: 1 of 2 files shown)


akre54:
Closed in favor of #3381.

Issue #3169, opened 2014-06-01
Binary Search for Sorted Collections


joshbambrick:
I am considering writing a small binary search function method for sorted collections. I would also modify the current `where` and `findWhere` methods in order to take advantage of this feature whenever possible (user search has only one field, which is the same as the comparator). This would offer a very significant reduction for searching already sorted collection to O(lg n). Furthermore, this speedup could be achieved by users without having to modify their code at all.

It may also be worth considering modifying the `Collection.set` method to offer O(n) insertion into sorted collections, as opposed to O(n lg n) currently achieved by re-sorting the entire collection. The main downside to this idea is that it may be slower if inserting many Models, not just one, because the collection is only re-sorted once in that case - so I think I will put this off in the meantime.

If I were to submit a pull request, would this be considered a desirable feature?

jashkenas:
Let's move this conversation over into the ticket(s).

PR #3207, opened 2014-06-25
Added Binary Search for Sorted Collections

joshuabambrick:
Added `search` method to collection prototype as per #3169. This is automatically used by `where` and `findWhere` when it is possible to do so to provide speedups without user needing to do anything.

For `toFind`, the user may provide:
- a comparison function which compares takes one value and returns 1 if the sought value has higher index, -1 if its lower and 0 if the value matches (`comparator` unused)
- a value to search for (the value of the `model[comparator]` sought - requires `comparator` to be a string)
- a model to search for (where `comparator` is a function)

Users may also request that it is the index of the model that is returned or that the model of highest/lowest index which matches the comparison is sought.

When no match is found, `undefined` is returned, or if `returnIndex` is true then `-1` is returned instead.

One point that may need to be considered is that the behaviour when a model is input as `toFind` but the `comparator` is a string is different from that when `comparator` is a function. Furthermore, the name of this method may also need some more thought to emphasise that it is for _binary_ searches or that it only works on sorted collections so `binarySearch` or the more user-friendly `sortedSearched` might be preferable.


joshbambrick:
bumped test underscore to 1.7.0 to include @megawac's new _.comparator method for consistency with `Array.prototype.sort` and `_.sortedIndex`
also only have `getMax` as a boolean, and otherwise efficiently fetch the minimum (with no time big-O complexity cost)

megawac:
@joshbambrick that's not 1.7 but the master working branch.

joshbambrick:
you're right. I wasn't quite sure what the protocol was in this situation.

akre54:
@megawac @joshbambrick where are we at with this? Worth adding to 1.2.0 or holding off a bit more?

megawac:
@akre54 I feel we should off on this until underscores methods are improved

akre54:
Ok I'll leave it be for now. Thanks.

joshbambrick:
we could update the pull request we a local copy of (https://github.com/jashkenas/backbone/pull/3207/files)[`_.comparator`], and then update it to use that defined in underscore, when it becomes available.

joshbambrick:
see most recent pull request for version with local copy of comparator (and also reset underscore.js version - oops)

joshbambrick:
could consider renaming to `binarySearch` to be Java-like, or `sortedIndex` since it does a similar thing to the underscore.js method

joshbambrick:
Any updates on the status of this? If the `compare` slows the search down too much, [as for _.comparator](https://github.com/jashkenas/underscore/pull/1939), we could put this back to the original version.

PR #3243, opened 2014-07-26
Sorted collection: use binary search to insert items

### LINKED ISSUE #3242: Poor "add" performance on sorted collections
I am having significant performance problems when adding items separately to a sorted collection.  This appears to be because a whole-list sort happens every time an item is added.  In my case, I cannot roll all of the additions into a single addition operation (which would trigger only one sort).

Performance could be improved by using binary search to find the insertion point (unfortunately not with underscore's sortedIndex, as this does not take a comparator), and inserting the record at the appropriate location instantly.

megawac:
Use `_.sortedIndex` to determine the new position of an element when inserting items in the simple `iterator` case. Continue delegating to `sort` for the comparator case.

Resolves #3242

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -688,7 +688,7 @@
       if (options.parse) models = this.parse(models, options);
       var singular = !_.isArray(models);
       models = singular ? (models ? [models] : []) : models.slice();
-      var id, model, attrs, existing, sort;
+      var id, model, attrs, existing, sort, simpleSort;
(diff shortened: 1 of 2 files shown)


jashkenas:
If it's breaking still, let's leave this be for now...

As I said up top — sortedIndex insert at position is how this used to be implemented. We changed to sort'ing for some reason at some point.

Issue #3070, opened 2014-03-17
Proposal: Backbone.Model::matches


gf3:
Hey, before I create a PR w/ tests I just wanted to test the waters for interest in a `matches` method for `Backbone.Model`. It essentially is a predicate that would take an object of conditions a la `Backbone.Collection::where`:

``` coffee
# model = new Backbone.Model({...})
#
# conditions =
#   key: 'value'
#   lol: 'wat'
#
# model.matches(conditions)

Backbone.Model::matches = (conditions) ->
  _.every conditions, (value, key) =>
    @get(key) == value
```

Is this something worthy of inclusion?

PR #3307, opened 2014-09-11
Add `Model::matches` as special-cased proxy to `_.matches`

gf3:
Discussion → #3070

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -319,6 +319,11 @@
       return this.get(attr) != null;
     },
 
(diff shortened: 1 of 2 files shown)


redexp:
This method should be rewritten on something which not produces proxy function. You not using it in `Collection#where` which is proof of my words, it's bad method for loop callbacks. Code is elegant but performance is just not good enough

jridgewell:
@redexp That could be accomplished with https://github.com/jashkenas/underscore/issues/1985.

PR #3330, opened 2014-10-03
Optimize underscore proxy methods

megawac:
Not sure how the maintainers will feel about this . The greatest benefit will clearly be for smaller collections/models.

You can do away with the code to support for length 2 and 4 if you'd rather (`_.filter(_.keys(_), function(name) {return _[name].length === 2})`)

**JSPerf** http://jsperf.com/backbone-underscore-funcs

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -261,6 +261,36 @@
     }
   };
 
+  // Proxy Underscore methods to a Backbone class' prototype using a
+  // particular attribute as the data argument
+  var addMethod = function(method, attribute) {
+    var _method = _.isString(method) ? method : method[0];
+    switch (_.isString(method) && _[method].length) {
+      case 1: return function() {
+        return _[_method](this[attribute]);
+      }
+      case 2: return function(value) {
+        return _[_method](this[attribute], value);
+      }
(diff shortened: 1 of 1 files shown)


jridgewell:
> The greatest benefit will clearly be for smaller collections/models.
> Doesn't seem terribly persuasive — we should be optimizing for large collections, not small ones, no?

I'm not sure how the size of the model/collection will affect the performance. This seems like a huge win, though, because `slice.call(arguments)` [deoptimizes](https://github.com/petkaantonov/bluebird/wiki/Optimization-killers#32-leaking-arguments) the method.

I made a [second jsperf](http://jsperf.com/backbone-underscore-funcs/3) to try and keep things in perspective (`_.keys` skews the graph, it's **way** faster than a `_.filter` will ever be). It's a pretty damn solid performance gain.

> The format is weird, this creates a dependency on underscore that backbone will have to maintain wrt the # of arguments...

I've opened https://github.com/megawac/backbone/pull/1 on @megawac's PR to sort this out.

jashkenas:
It's really a bummer that JS engines respond well to jumping through these kinds of silly hoops. *sigh*

PR #3338, opened 2014-10-10
Fix a bunch of accessibility issues.

danielnixon:
* Added missing label for "What needs to be done?" text input.
* Added `href` attributes to all `<a>` tags to make them able to be focused with a keyboard.
* Replaced "double click to edit" (which can't be accessed with a keyboard [or from a touch device, for that matter]) with an edit button.
* Associated the TODO item labels with their check boxes.
* Always show the delete button. Showing it only when the parent `<li>` has mouse hover doesn't work with a keyboard (or from a touch device, for that matter).
* Apply the existing `:hover` styles to `:focus` as well. In general you always want to apply styles to both.
* Added the WAI-ARIA `role="button"` attribute to the buttons that are (probably incorrectly) marked up as `<a>` tags. They should really be marked up using `<button>`, but in the interest of treading lightly I left them as `<a>` tags. The fact that the `<a>` tags didn't have `href` attributes (and that they now have `href="#"`) is the clue that they are really buttons and not links.
* Added screen reader only "delete" text to the delete buttons, which were previously empty (with only a background image).

### DIFF
--- a/examples/todos/index.html
+++ b/examples/todos/index.html
@@ -13,6 +13,7 @@
 
     <header>
       <h1>Todos</h1>
+      <label class="sr-only" for="new-todo">What needs to be done?</label>
       <input id="new-todo" type="text" placeholder="What needs to be done?">
     </header>
 
@@ -23,14 +24,10 @@
     </section>
 
     <footer>
-      <a id="clear-completed">Clear completed</a>
+      <a href="#" id="clear-completed" role="button">Clear completed</a>
       <div id="todo-count"></div>
     </footer>
 
-  </div>
-
-  <div id="instructions">
-    Double-click to edit a todo.
   </div>
 
   <div id="credits">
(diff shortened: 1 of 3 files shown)


jashkenas:
This changed the appearance of the demo significantly — degrading it and making it more complex. It shouldn't have been merged.

That said, the demo can sure use a cleanup. It should be a simple, instructive, learning example. (Although perhaps it's too old and well-worn for that at this point.)

danielnixon:
Could you be more specific about the significant appearance change you're referring to? Is it the visible edit and delete buttons or something else? Those two features are currently hidden behind a double click and a mouse hover, respectively. Having them be permanently visible makes things _less_ complex, not more, doesn't it? That the "double click to edit" instructional text is required is telling, isn't it?

Totally agree re "simple, instructive, learning example". Why not have it be accessible (not to mention touch device friendly) as well?

PR #3395, opened 2014-11-25
`clear` should set attributes to defaults

samjoch:
When i defined `defaults` attributes, on `clear` those values are not retrieved.


platinumazure:
This functionality can be achieved pretty easily like so:

    model.clear().set(_.result(model, "defaults"));

To avoid a double-flash in your view, you could add a custom option to the `clear` call and have the view callbacks not perform the update if it's happening during this "clear-then-defaults" call.

akre54:
I think simpler wins out here. Thanks anyways.

PR #3397, opened 2014-11-26
Allow events api callbacks to be method names

### LINKED ISSUE #3333: Allow listenTo callback to be a method name? (string)
Since "The **callback** will always be called with **object** as context", perhaps we could optionally give the string name of a method on **object**? This would more closely mirror the functionality of the events hash.
It does open the doors for an error when that method doesn't exist on **object**, but people can already pass undefined so it's an issue that really already exists.

thejameskyle:
Resolves #3333, moving from #3396

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -245,7 +245,7 @@
     }
 
     // Handle string callbacks.
-    if (callback && _.isString(callback)) {
+    if (typeof callback === 'string') {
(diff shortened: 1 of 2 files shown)


jashkenas:
A wee bit to string-programm-y. You have a reference to the method right there ... just use the reference.

Issue #3283, opened 2014-08-29
Support for opts.context in AJAX callbacks


naganowl:
The [jQuery AJAX docs](http://api.jquery.com/jQuery.ajax/) state that a `context` attribute can be specified on `opts` to specify what `this` should refer to in all AJAX-related callbacks.

For many of the model server calls (such as `save` or `fetch`), the `success` and `error` callbacks are wrapped in such a way that any provided methods already on `opts` will be invoked with the default context, rather than the provided context. See examples of this in the [success](https://github.com/jashkenas/backbone/blob/master/backbone.js#L453) and  [error](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1696) callbacks of `fetch` in the source.

This was somewhat mentioned in [#2192](https://github.com/jashkenas/backbone/issues/2192), but was closed and points to other issues which are only tangentially related to use of a context. I understand there are workarounds for this, but is there a reason why these callbacks don't use `call` or `apply` or even respect `opts.context`?

Thanks!

PR #3423, opened 2014-12-22
Call the success callback with context

jridgewell:
Fixes https://github.com/jashkenas/backbone/issues/3283. The only niggle is `Model#destroy` when `isNew`, since it doesn't go through the usual ajax process.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -464,7 +464,7 @@
       var success = options.success;
       options.success = function(resp) {
         if (!model.set(model.parse(resp, options), options)) return false;
-        if (success) success(model, resp, options);
+        if (success) success.call(options.context, model, resp, options);
         model.trigger('sync', model, resp, options);
       };
       wrapError(this, options);
@@ -514,7 +514,7 @@
         if (_.isObject(serverAttrs) && !model.set(serverAttrs, options)) {
           return false;
         }
-        if (success) success(model, resp, options);
+        if (success) success.call(options.context, model, resp, options);
         model.trigger('sync', model, resp, options);
       };
       wrapError(this, options);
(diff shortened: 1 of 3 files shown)


CoryDanielson:
This code assumes that `options.context` is always defined. This would break callbacks that have already had their context bound via `.bind()`. To fix this, the `.call` should only be done conditionally, like so:

```
if (success) {
    if (options.context)
        success.call(options.context, model, resp, options);
    else 
        success(model, resp, options);
}

if (error) {
    if (options.context)
        error.call(options.context, model, resp, options);
    else
        success(model, resp, options);
}
```

I don't feel that this code is coupled with the jQuery AJAX API. It just happens to implement a similar API.

jridgewell:
> This would break callbacks that have already had their context bound via .bind()

It won't. Doesn't matter what context you call a bound function with, the one supplied to `bind` (whether native or `_.bind`) will be the context of the callback.

Consider that `_.bind` essentially maps to

```javascript
_.bind = function(fn, context) {
    return function() {
        return fn.call(context, arguments);
    };
};

var cb = _.bind(success, model); // => function() { return success.apply(model, arguments) };

cb.call(undefined, 1, 2, 3);
/*
// Maps to
(function() {
    return success.apply(model, arguments);
}).call(undefined, 1, 2, 3)

// Maps to
success.apply(model, [1, 2, 3]);
```

CoryDanielson:
Ahhh, yep :+1:
```
var fn = function () { alert(this); }.bind('A');
fn.call('B');
> A
```

With the code as it is, the only other issue I have is that it would be a minor performance hit. It would only be minor, because these callbacks would not be executed very rapidly, unless maybe websockets were involved.

http://jsperf.com/unnecessary-call-test

PR #3425, opened 2014-12-24
Make it possible to decode a url containing an encodeURIComponent encode...

asparagino:
...d % symbol

We've been having problems with params that have been encoded as URI parts with encodeURIComponent. For example:

'http://foo.com/bar/'+encodeURIComponent('foo%/bar')

Somewhere in the router decodeURI is getting called on the URI, then it's throwing a URIError in  _extractParameters when decodeURIComponent is getting called on an already decoded % character.

This patch fixes our app by falling back on regex substitutions of URI separators not already decoded by decodeURI when a URIError is thrown. It is still necessary to perform replacements of characters decodeURI does not catch.

Confusingly, the unit test doesn't seem to call decodeURI on the url with the checkURL call, so I've left decodeURIComponent in place rather than avoiding it completely. It would break other tests to remove it and I'm not confident I know all the code paths that call _extractParameters.

Signed-off-by: Pete Setchell <pete@crittercism.com>

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1390,7 +1390,38 @@
       return _.map(params, function(param, i) {
         // Don't decode the search params.
         if (i === params.length - 1) return param || null;
-        return param ? decodeURIComponent(param) : null;
+        // decodeURIComponent on a param already partially decoded by decodeURI call causes a URIError.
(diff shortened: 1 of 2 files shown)


jashkenas:
Have you seen this? 

https://github.com/jashkenas/backbone/pull/3434

Does it address your problem?

PR #3427, opened 2014-12-25
Backbone Conf videos

brock:
Adds a section called "Talks". I looked at creating a table with tags, but there were so many good talks, and I guess I didn't feel qualified to eliminate any of them. :grinning:

### DIFF
--- a/index.html
+++ b/index.html
@@ -505,6 +505,10 @@
       <li>– <a href="#examples-tzigla">Tzigla</a></li>
     </ul>
 
+    <a class="toc_title" href="#talks">
+      Talks
+    </a>
+
     <a class="toc_title" href="#changelog">
       Change Log
     </a>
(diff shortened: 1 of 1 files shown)


jashkenas:
I think that a curated table/list with titles and tags would be more valuable then just leaving readers to go wander through on their own ... In particular, it could be a nice complement to the FAQ. Anyone else want to pitch in on this?

jmeas:
I think that it should be a section of the website. Videos are a great reference for developers of all skill levels.

jashkenas:
This is still open for anyone who feels like writing a little bit of HTML.

jashkenas:
Looks like nobody picked this up ... closing.

elgubenis:
I'm late but i'd pick that up, if thats ok

jridgewell:
Sure, mind opening another PR? Bonus points for including @brock's original commit.

PR #3432, opened 2015-01-03
Encourage the use of #listenTo

jridgewell:
No functional changes, just encourages the use of `#listenTo` over `#on`.


braddunbar:
Hey @jridgewell! Thanks for the patch!

While `stopListening` is great for easily preventing memory leaks, I don't think it's necessary in this case. There is a clear point at which `#off` can be called with a specific function and context and thus we shouldn't introduce extra overhead via `#listenTo`.

PR #3434, opened 2015-01-04
Decode unicode escapes without decoding `%25`.

### LINKED ISSUE #3426: URIError parsing url Params encoded with encodeURIComponent
It appears that decodeURI is called on urls before _extractParameters calls decodeURIComponent on params extracted from the url. This can cause a URIError when the param has already been partially decoded to contain a character that wouldn't be legal in a URI.

Example url: 'http://foo.com/bar/'+encodeURIComponent('foo%/bar')

Here's a fix that works for us, though I'm not confident I've tested with router and backbone history options we're not using:
https://github.com/jashkenas/backbone/pull/3425

braddunbar:
Fixes #3426.

### The Problem

By using `decodeURI` to decode unicode escapes, `%25` is also decoded. When included in a url parameter, this creates an invalid escape sequence and causes `decodeURIComponent` to throw.

*Sidebar: Should invalid sequences cause the router to throw in the first place? What else can it do?*

### The Solution

Double escape `%25` as `%2525` before using `decodeURI`.

### Is this another step down a crazy rabbit hole of URI escaping?

Maybe? I'm pretty happy with the outcome though. I like that we can continue to remove escapes if we come across more problems of this nature. Further, adding it as `History#decodeFragment` means users can customize it to patch things easily when needed.

I'm going to do a bit more research and try to figure out why the browser escapes unicode in the first place. It effectively prevents us from knowing what was escaped by the user and what was escaped by the browser. :confused:

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1440,7 +1440,9 @@
       return path === this.root && !this.getSearch();
     },
 
-    // Decode unicode escapes without decoding `%25`.
+    // Unicode characters in `location.pathname` are percent encoded so they're
+    // decoded for comparison. `%25` should not be decoded since it may be part
+    // of an encoded parameter.
     decodeFragment: function(fragment) {
       return decodeURI(fragment.replace(/%25/g, '%2525'));
     },
--- a/backbone.js
+++ b/backbone.js
@@ -1440,6 +1440,11 @@
       return path === this.root && !this.getSearch();
     },
 
+    // Decode unicode escapes without decoding `%25`.
+    decodeFragment: function(fragment) {
+      return decodeURI(fragment.replace(/%25/g, '%2525'));
+    },
+
     // In IE6, the hash fragment and search params are incorrect if the
     // fragment contains `?`.
     getSearch: function() {
@@ -1456,7 +1461,9 @@
 
     // Get the pathname and search params, without the root.
     getPath: function() {
-      var path = decodeURI(this.location.pathname + this.getSearch());
+      var path = this.decodeFragment(
+        this.location.pathname + this.getSearch()
+      );
       var root = this.root.slice(0, -1);
       if (!path.indexOf(root)) path = path.slice(root.length);
       return path.charAt(0) === '/' ? path.slice(1) : path;
@@ -1629,7 +1636,7 @@
       var url = root + fragment;
 
       // Strip the hash and decode for matching.
-      fragment = decodeURI(fragment.replace(pathStripper, ''));
+      fragment = this.decodeFragment(fragment.replace(pathStripper, ''));
 
       if (this.fragment === fragment) return;
       this.fragment = fragment;
--- a/test/router.js
+++ b/test/router.js
@@ -829,6 +829,21 @@
     Backbone.history.start({pushState: true});
   });
 
+  test('unicode pathname with % in a parameter', 1, function() {
+    location.replace('http://example.com/myyjä/foo%20%25%3F%2f%40%25%20bar');
+    Backbone.history.stop();
+    Backbone.history = _.extend(new Backbone.History, {location: location});
+    var Router = Backbone.Router.extend({
+      routes: {
+        'myyjä/:query': function(query) {
+          strictEqual(query, 'foo %?/@% bar');
+        }
+      }
+    });
+    new Router;
+    Backbone.history.start({pushState: true});
+  });
+
   test('newline in route', 1, function() {
     location.replace('http://example.com/stuff%0Anonsense?param=foo%0Abar');
     Backbone.history.stop();


braddunbar:
For some added fun, PhantomJS displays a behavior not found in any actual browser (the tests pass in all the browsers I could find to test). :unamused:

backbone.js
```diff
@@ -1440,6 +1440,11 @@
       return path === this.root && !this.getSearch();
     },
 
+    // Decode unicode escapes without decoding `%25`.
```
akre54: Can you add a bit more prose here for future-usses who might look at this quizzically? It's always nice to have a bit more context in comments than just looking at the code would give us.

akre54: Also, any reason this has to be on the Router prototype instead of living as a standalone helper function?

macgyver:
Works in Chrome 41, Firefox 33, Safari 8, Safari for iOS 7.1, and IE8-10 (the results of extractParams are properly decoded strings when the url is a percent-encoded string, even when including `%25`)

backbone.js
braddunbar: > Can you add a bit more prose here for future-usses who might look at this quizzically?

Happy to!

> Also, any reason this has to be on the Router prototype instead of living as a standalone helper function?

I started out with a standalone helper but then I started to think what would need to be done if someone finds another problematic escape sequence. With the helper, you'd need to override `#getPath` and `#navigate`, which could get ugly really fast. This way, it's a one liner to replace `#decodeFragment`.

braddunbar: Updated in faea842. It's hard to strike a balance between succinctness and completeness so let me know if I should tweak it! :smiley:

jashkenas:
Looks fine to me -- but it would be nice to get someone familiar with Browser-unicode-escaping to chime in ... if such a unicorn exists.

akre54:
praps @mathiasbynens can help per usual?

backbone.js
akre54: excellent. thanks!

braddunbar:
FYI, PhantomJS is failing because it decodes the entire url, including the fragment and all the percent-encoded characters. This is not behavior exhibited by any browser and I'm reticent to add a fix for it. If we can't trust it to behave like a browser then what good does it do us?

akre54:
I'd like to merge this but I'm weary of breaking every Travis build from now on. Do we have a plan around this?

braddunbar:
My thoughts exactly! I'd love some feedback regarding that. PhantomJS is a nice smoke test but I'm not sure what good it does if it's giving false negatives via behavior not exhibited in any real browser.

akre54:
We've had numerous problems with PhantomJS in the past, and it seems due to slow development most of these bugs are here to stay. @megawac created a pull removing PhantomJS from Underscore in jashkenas/underscore#2054, I wonder if he'll be kind enough to do us the favor here too.

megawac:
Was planning on filing an issue with it later today actually about some of
the things that would be involved

akre54:
Excellent. Looking forward.

jridgewell:
I was working on an alternative solution in my [url-encoding](https://github.com/jridgewell/backbone/compare/url-encoding) branch, but never got polished it into a PR. It's working on Chrome, Safari, Firefox, and IE6 (only IE vm I have). I can have it ready tonight, if you'd like.

jridgewell:
Also: this is failing in IE6 IE6 Failure

I've found two issues:

1. The percent encoded "?", and turned that into a percent decoded `location.search`:
  ```javascript
  location = {
    href: "http://example.com/myyjä/foo%20%25%3F%2f%40%25%20bar",
    hash: "",
    host: "example.com:80",
    search: "?/@% bar",
    fragment: "",
    pathname: "/myyjä/foo%20%",
    protocol: "http:"
  }
  ```

  - New issue related to https://github.com/jashkenas/backbone/pull/3152?
2. The pathname is already decoded, so trying to `fragment.replace(/%25/g, '%2525')` never finds anything.

jridgewell:
Follow up: it's because the anchor tag `parser` will decode the pathname. Putting it into the hash, then starting up with `{pushState: true}`, prevents the decode, but requires a bit of boilerplate.

jashkenas:
@braddunbar Any thoughts on @jridgewell's last couple of comments?

jridgewell:
I fixed it by setting `location.pathname` [directly](https://github.com/jashkenas/backbone/pull/3506/files#diff-c6bcc5b8d170df93e5b3223fa590f1f7R834). IE6's _actual_ `location` object doesn't decode the pathname.

jashkenas:
@jridgewell So this is good to merge in your opinion?

jridgewell:
I think so.

jashkenas:
I wouldn't bet money on this change not invoking some hellacious old-IE or old-Firefox decoding problem down the road. But hey — here's to optimism.

akre54:
It looks like this test is failing in Android 4.0.4 and Safari 5 on Win7. 

https://travis-ci.org/jashkenas/backbone/builds/51918113

jashkenas:
> I wouldn't bet money on this change not invoking some hellacious old-IE or old-Firefox decoding problem down the road.

Do I win a prize? ;)

akre54:
I dunno. You said down the road...

braddunbar:
Haha! PhantomJS also failed so my first guess is that it's the same as IE6 and we can get around it by setting `pathname` directly. I'll take a look later on!

jashkenas:
For what it's worth, this patch and new test are currently failing in IE9.

braddunbar:
Hmm…looks ok on BrowserStack.

screen shot 2015-05-13 at 5 45 52 pm

jashkenas:
It does, but it breaks on my local VM. Go figure.

jridgewell:
All tests are passing on mine. Master on IE8 XP

PR #3436, opened 2015-01-05
fix(License): update year and header

gdi2290:
I added a proper LICENSE file header according to MIT site


akre54:
It actually isn't necessary to update the current year at the top of the file. We'll update it when we cut the next release. The name of the license is also unnecessary.

PR #3441, opened 2015-01-07
Update documentation for Backbone.Router

balloob:
The old documentation makes it look `Backbone.history.start()` starts listening for new hash changes and you have to call `Backbone.history.start({pushState: true})` to route the current URL too. This does not match the behavior in the code.

I removed the `,` from the Backbone.Router documentation to clarify this.


akre54:
Nice addition. Thanks.

PR #3442, opened 2015-01-08
Collection url method can include query string

wan2land:
i use backbonejs with cross domain ajax, and collection's url must include query string(token).

`url: items?foo=bar`

in put, delete method

items?foo=bar/item_id -> items/item_id?foo=bar


akre54:
Try passing a `data` param or editing the model's `url` function to do what you want.

Issue #2656, opened 2013-07-04
Optional trailing slash


joshpangell:
When defining a root route, then pushing history onto the stack, then removing it, there is a trailing slash left over. This is not very desirable and would be great as an optional flag when starting history.

The origin of this appears to [reside in the code here](http://backbonejs.org/docs/backbone.html#section-171).
## 

As an example:

``` javascript
Backbone.history.start({
    pushState: true,
    root: "/myroot"
});
```

Opening a page and pushing it onto the history stack works.

``` javascript
var url = "/Page-name";
Router.navigate( url, { trigger: true });
```

Result: **example.com/myroot/Page-name** (this is the desired result)

Then, when closing the page, passing a blank value to remove the page url creates a trailing slash:

``` javascript
var url = "";
Router.navigate( url, { trigger: true });
```

Result: **example.com/myroot/**

Desired result: **example.com/myroot** (no trailing slash)

PR #3443, opened 2015-01-12
History's root

jridgewell:
This is another take on https://github.com/jashkenas/backbone/issues/2656 to allow an optional trailing `/` for root. Instead of normalizing the root to always include the trailing and the baggage that comes with it (all the `slice`s and having to normalize `location.pathname` to include it), it instead only adds the slash when necessary.

I've kept the following:

```javascript
Backbone.history.start({ root: '/root' });
Backbone.history.navigate('?a=b');
console.log(location.href); `.../root?a=b`
```

But it does not apply when a mandatory trailing slash is added:
```javascript
Backbone.history.start({ root: '/root/' });
Backbone.history.navigate('?a=b');
console.log(location.href); `.../root/?a=b`
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1464,7 +1464,7 @@
     // Get the cross-browser normalized URL fragment from the path or hash.
     getFragment: function(fragment) {
       if (fragment == null) {
-        if (this._hasPushState || !this._wantsHashChange) {
+        if (this._usePushState || !this._wantsHashChange) {
           fragment = this.getPath();
         } else {
           fragment = this.getHash();
@@ -1486,7 +1486,8 @@
       this._wantsHashChange = this.options.hashChange !== false;
       this._hasHashChange   = 'onhashchange' in window;
       this._wantsPushState  = !!this.options.pushState;
-      this._hasPushState    = !!(this.options.pushState && this.history && this.history.pushState);
+      this._hasPushState    = !!(this.history && this.history.pushState);
+      this._usePushState    = this._wantsPushState && this._hasPushState;
       this.fragment         = this.getFragment();
 
       // Normalize root to always include a leading slash.
@@ -1514,7 +1515,7 @@
       // Proxy an iframe to handle location events if the browser doesn't
       // support the `hashchange` event, HTML5 history, or the user wants
       // `hashChange` but not `pushState`.
-      if (!this._hasHashChange && this._wantsHashChange && (!this._wantsPushState || !this._hasPushState)) {
+      if (!this._hasHashChange && this._wantsHashChange && !this._usePushState) {
         var iframe = document.createElement('iframe');
         iframe.src = 'javascript:0';
         iframe.style.display = 'none';
@@ -1533,7 +1534,7 @@
 
       // Depending on whether we're using pushState or hashes, and whether
       // 'onhashchange' is supported, determine how we check the URL state.
-      if (this._hasPushState) {
+      if (this._usePushState) {
         addEventListener('popstate', this.checkUrl, false);
       } else if (this._wantsHashChange && this._hasHashChange && !this.iframe) {
         addEventListener('hashchange', this.checkUrl, false);
@@ -1553,7 +1554,7 @@
       };
 
       // Remove window listeners.
-      if (this._hasPushState) {
+      if (this._usePushState) {
         removeEventListener('popstate', this.checkUrl, false);
       } else if (this._wantsHashChange && this._hasHashChange && !this.iframe) {
         removeEventListener('hashchange', this.checkUrl, false);
@@ -1621,7 +1622,7 @@
 
       // Append a trailing slash if needed.
       var root = this.root;
-      if (fragment && this._hasPushState) {
+      if (fragment && this._usePushState) {
         if (root.charAt(root.length - 1) !== '/' && fragment.charAt(0) !== '?') root += '/';
       }
 
@@ -1634,7 +1635,7 @@
       this.fragment = fragment;
 
       // If pushState is available, we use it to set the fragment as a real URL.
-      if (this._hasPushState) {
+      if (this._usePushState) {
         this.history[options.replace ? 'replaceState' : 'pushState']({}, document.title, url);
 
       // If hash changes haven't been explicitly disabled, update the hash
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -1487,11 +1486,12 @@
...
       this._wantsPushState  = !!this.options.pushState;
-      this._hasPushState    = !!(this.options.pushState && this.history && this.history.pushState);
+      this._hasPushState    = !!(this.history && this.history.pushState);
+      this._usePushState    = this._wantsPushState && this._hasPushState;
       this.fragment         = this.getFragment();
 
-      // Normalize root to always include a leading and trailing slash.
-      this.root = ('/' + this.root + '/').replace(rootStripper, '/');
+      // Normalize root to always include a leading slash.
+      this.root = (this.root || '/').replace(/^[^\/]/, '/$&');
```
braddunbar: I think this can be simpler. The default is already handled above. Also, we should remove `rootStripper` if it's not in use. I think we can avoid the regex entirely with the following.

```js
if (this.root.charAt(0) !== '/') this.root = '/' + this.root;
```

jridgewell: Agreed. I had just found the replace `$&` on line 1439 and thought it would be nifty to use again.

braddunbar: I'm partial to it as well! :smiley:

backbone.js
```diff
@@ -1621,11 +1620,12 @@
...
       fragment = this.getFragment(fragment || '');
 
-      // Don't include a trailing slash on the root.
+      // Append a trailing slash if needed.
       var root = this.root;
-      if (fragment === '' || fragment.charAt(0) === '?') {
-        root = root.slice(0, -1) || '/';
+      if (fragment && this._usePushState) {
+        if (root.charAt(root.length - 1) !== '/' && fragment.charAt(0) !== '?') root += '/';
       }
```
braddunbar: I think I'd flip this on it's head and prepend the slash to `fragment` when needed.

```js
if (this._usePushState && fragment.charAt(0) !== '?') fragment = '/' + fragment;
```

backbone.js
```diff
@@ -1436,8 +1436,7 @@
 
     // Are we at the app root?
     atRoot: function() {
-      var path = this.location.pathname.replace(/[^\/]$/, '$&/');
-      return path === this.root && !this.getSearch();
+      return this.location.pathname === this.root && !this.getSearch();
```
braddunbar: I think we'd still need to normalize here because I don't think the result should change based on trailing slash preference.

backbone.js
braddunbar: Ah, but `root` might already have a trailing slash…nevermind.

jridgewell: I would still need to check `fragment !== ''` and `root.charAt(root.length - 1) !== '/'` to avoid adding an unnecessary slash.

```javascript
Backbone.history.start({ root: '/root' });
// Shouldn't have a trailing slash
Backbone.history.navigate('');
location.pathname === '/root';

Backbone.history.start({ root: '/root/' });
// Shouldn't have a double slash
Backbone.history.navigate('test');
location.pathname === '/root/test';

**Comment by braddunbar** [comment]
I'm a bit wary of this because it's a breaking change. I've written code in the past that depends on the fact that `History.root` will always contain a leading and trailing slash. Like #3391, if the change is purely aesthetic I think we should probably leave it alone.

File: backbone.js (hunk 3 above)
Comment (jridgewell) [inline backbone.js]: I felt iffy about this as well. However, I remember reading jashkenas arguing `/root` is not the same resource as `/root/` (but I can't find it!).

I think the only downside to this as-is is: At `/root` and `{ root: '/root/' }` will cause a redirect when transitioning the app to `hashChange`. End url will be `/root/#root`.

Comment (braddunbar) [inline backbone.js]: > I felt iffy about this as well. However, I remember reading jashkenas arguing /root is not the same resource as /root/ (but I can't find it!).

I agree, though I think they should behave similarly regarding a hashChange redirect.

> I think the only downside to this as-is is: At /root and { root: '/root/' } will cause a redirect when transitioning the app to hashChange.

Something like that yea. It's a pretty nasty breaking change.

Comment (jridgewell) [inline backbone.js]: Actually, scratch that downside. It works as-is. :smile:

Comment (jridgewell) [inline backbone.js]: Edit: nope, should have tested it better...

File: backbone.js [hunk 4]
```diff
@@ -1487,11 +1483,12 @@
       this._wantsHashChange = this.options.hashChange !== false;
       this._hasHashChange   = 'onhashchange' in window;
       this._wantsPushState  = !!this.options.pushState;
-      this._hasPushState    = !!(this.options.pushState && this.history && this.history.pushState);
+      this._hasPushState    = !!(this.history && this.history.pushState);
+      this._usePushState    = this._wantsPushState && this._hasPushState;
```
braddunbar: Regardless of where `root` ends up, I like this. I've always thought the way we conflate `_wantsPushState` and `_hasPushState` will eventually cause a bug. :clap:

jridgewell: This is a gentle nudge towards https://github.com/jashkenas/backbone/pull/3394, so that the props actually mean what their name implies.

backbone.js
jridgewell: Should be fixed now.

backbone.js
akre54: :+1:

braddunbar:
Hi @akre54! I'm ok with this functionality generally, but I don't think it should be merged as is because it's a breaking change as I've indicated above. `Backbone.history.root` is a public-ish API that I've depended on before and I'm sure others have as well. I don't think we should break it for an aesthetic option.

akre54:
Ah sorry, I think you're right. I'll back it out.

braddunbar:
Thanks much @akre54! And @jridgewell, I'm still fine with adding the functionality as long as it keeps the leading and trailing slash on `Backbone.history.root`. :smiley:

jridgewell:
> I'm still fine with adding the functionality as long as it keeps the leading and trailing slash on Backbone.history.root.

Meaning you still want the normalized (`/root` -> `/root/`) as `Backbone.history.root`, but an internal `_root = '/root'`? I think at that point, `Backbone.history.root` is very misleading?

braddunbar:
I was thinking more along the lines of `this._trailingSlash = /\/$/.test(options.root);` or similar. Then just use the one check when setting the url.

PR #3449, opened 2015-01-19
Fix #3448. Ensure listenToOnce splits event names.

### LINKED ISSUE #3448: listenToOnce space-separated events only attaching callback once
http://jsbin.com/wuriku/1/edit?js,console

```js
events.once('foo bar', callback);
events.trigger('foo');
events.trigger('foo');
events.trigger('bar');
events.trigger('bar');

console.log(count === 2); // >> true
```

```js
events.listenToOnce(target, 'foo bar', callback);
target.trigger('foo');
target.trigger('foo');
target.trigger('bar');
target.trigger('bar');

console.log(count === 2); // >> false
```

Removing the (seemingly overkill) `_.once` would fix this.

braddunbar:
Calling `listenToOnce('x y', …)` should be functionally equivalent to `listenToOnce('x', …); listenToOnce('y', …)`.

I realize this isn't the most elegant fix, but it works correctly. The difference in argument order for `listenToOnce` makes it hard to implement `eventApi` for both. I'll look into simplifying though.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -184,12 +184,19 @@
         for (var event in name) this.listenToOnce(obj, event, name[event]);
         return this;
       }
-      var cb = _.once(function() {
-        this.stopListening(obj, name, cb);
+      if (eventSplitter.test(name)) {
+        var names = name.split(eventSplitter);
+        for (var i = 0, length = names.length; i < length; i++) {
+          this.listenToOnce(obj, names[i], callback);
+        }
+        return this;
+      }
+      var once = _.once(function() {
+        this.stopListening(obj, name, once);
         callback.apply(this, arguments);
       });
-      cb._callback = callback;
-      return this.listenTo(obj, name, cb);
+      once._callback = callback;
+      return this.listenTo(obj, name, once);
     },
 
     // Tell this object to stop listening to either specific events ... or
--- a/test/events.js
+++ b/test/events.js
@@ -541,11 +541,12 @@
   test("#3448 - listenToOnce with space-separated events", 2, function() {
     var one = _.extend({}, Backbone.Events);
     var two = _.extend({}, Backbone.Events);
-    one.listenToOnce(two, 'x y', function() { ok(true); });
-    two.trigger('x');
-    two.trigger('x');
-    two.trigger('y');
-    two.trigger('y');
+    var count = 1;
+    one.listenToOnce(two, 'x y', function(n) { ok(n === count++); });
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -184,12 +184,19 @@
...
-      var cb = _.once(function() {
-        this.stopListening(obj, name, cb);
+      if (eventSplitter.test(name)) {
+        var names = name.split(eventSplitter);
+        for (var i = 0, length = names.length; i < length; i++) {
+          this.listenToOnce(obj, names[i], callback);
+        }
+        return this;
+      }
+      var once = _.once(function() {
```
braddunbar: I replaced `cb` with `once` because it mirrors the implementation of `#once` and I think it's a more descriptive name.

test/events.js
```diff
@@ -538,4 +538,14 @@
     equal(obj, obj.stopListening());
   });
 
+  test("#3448 - listenToOnce with space-separated events", 2, function() {
+    var one = _.extend({}, Backbone.Events);
+    var two = _.extend({}, Backbone.Events);
+    one.listenToOnce(two, 'x y', function() { ok(true); });
+    two.trigger('x');
+    two.trigger('x');
```
jridgewell: Should we test "x" triggers once and "y" triggers once?

braddunbar: Not a bad idea! I'll see what I can do.

braddunbar: Updated in dc048c6.

akre54:
Yeah I feel like we're pushing `eventsApi` to its limit these days. Might be worth looking into the various use cases and refactoring this chunk.

backbone.js
akre54: :+1:

akre54:
Merging. We can simplify later.

jashkenas:
Can we simplify this?

jridgewell:
> Can we simplify this?

#3455 does this by turning `eventsApi` into a [reducing function](https://github.com/jridgewell/backbone/blob/4018a5683f77038efa43959223b83badc486ce3e/backbone.js#L84-L105) (analogous to `_.reduce`) and [`on`](https://github.com/jridgewell/backbone/blob/4018a5683f77038efa43959223b83badc486ce3e/backbone.js#L137-L144) and [`off`](https://github.com/jridgewell/backbone/blob/4018a5683f77038efa43959223b83badc486ce3e/backbone.js#L182-L218) into reduce iteratees.

Issue #3453, opened 2015-01-22
Memory leak in stopListening?


albinohrn:
Hi,

this issue is some what related to ticket #3226 but it's not the same. If you think they are feel free to close and reference that.

After the fix of #3226 there's still a leak. But this doesn't just apply to the listenToOnce. Take a look at the following [example](http://requirebin.com/?gist=8a50edb1123e063fe875). (The example runs Backbone v1.1.2 but the master works the same way).

The problem is that if the `model` I'm listening to has a listener as well. Then the `view.listeningTo` will not be cleaned up correctly. 

It is because of this [line](https://github.com/jashkenas/backbone/blob/master/backbone.js#L213):

```javascript
    if (remove || _.isEmpty(obj._events)) delete this._listeningTo[id];
```
`obj._events` isn't empty and therefore the `view` will keep its reference to the `model`. Is this correct? In our project this causes problems since the object that is listened to will always have its own listeners and where using `listenToOnce` so the references to the old objects will be kept "forever". But we might have to rethink our solution to workaround this issue?

/Albin

smelnikov:
A pull request #3049 is already open to address this issue. Although there seems to be no desire to merge it at this time.

PR #3455, opened 2015-01-22
Fix listenTo memory leak

jridgewell:
Fixes https://github.com/jashkenas/backbone/issues/3453, and competes with https://github.com/jashkenas/backbone/pull/3049.

I think this aligns more closely with what's currently implemented. :smile:

It also implements `listenApi`, which is `eventsApi` translated to the `listenTo` / `stopListening` method signatures.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -137,7 +137,7 @@
       var listeningTo = this._listeningTo;
       if (!listeningTo) return this;
 
-      var listeneeIds = (obj) ? [obj._listenId] : _.keys(listeningTo);
+      var listeneeIds = obj ? [obj._listenId] : _.keys(listeningTo);
       for (var i = 0, length = listeneeIds.length; i < length; i++) {
         var id = listeneeIds[i];
         var listenee = listeningTo[id];
@@ -234,23 +234,18 @@
     return obj;
   };
 
-  // Maps the normalized event callbacks into onceWrappers.
+  // Maps the normalized event callbacks into once wrappers.
   var onceMap = function(name, callback, offer) {
     return eventsApi(function(map, name, callback, offer) {
-      if (callback) map[name] = onceWrap(name, callback, offer);
+      if (callback) {
+        var once = map[name] = _.once(function() {
+          offer(name, once);
+          callback.apply(this, arguments);
+        });
+        once._callback = callback;
+      }
       return map;
     }, {}, name, callback, offer);
-  };
-
-  // Wraps an event callback, using the `offer` function to off the event
-  // once it's been called.
-  var onceWrap = function(name, callback, offer) {
-    var once = _.once(function() {
-      offer(name, once);
-      callback.apply(this, arguments);
-    });
-    once._callback = callback;
-    return once;
   };
 
   // A difficult-to-believe, but optimized internal dispatch function for
--- a/backbone.js
+++ b/backbone.js
@@ -152,12 +152,13 @@
     stopListening: function(obj, name, callback) {
       var listeningTo = this._listeningTo;
       if (!listeningTo || !listenApi(this, 'stopListening', obj, name, callback)) return this;
-      if (obj) listeningTo = _.pick(listeningTo, obj._listenId);
-      for (var id in listeningTo) {
-        var listenee = listeningTo[id];
+      var listeneeIds = (obj) ? [obj._listenId] : _.keys(listeningTo);
+      for (var i = 0, length = listeneeIds.length; i < length; i++) {
+        var listenee = listeningTo[listeneeIds[i]];
+        if (!listenee) continue;
         listenee.obj.off(name, callback, this);
         var events = offApi(listenee.events, name, callback);
-        if (!events) delete this._listeningTo[id];
+        if (!events) delete this._listeningTo[listeneeIds[i]];
       }
       if (_.isEmpty(this._listeningTo)) this._listeningTo = void 0;
       return this;
--- a/backbone.js
+++ b/backbone.js
@@ -82,9 +82,7 @@
     // the callback to all events fired.
     on: function(name, callback, context) {
       if (!eventsApi(this, 'on', name, [callback, context]) || !callback) return this;
-      this._events || (this._events = {});
-      var events = this._events[name] || [];
-      this._events[name] = events.concat({callback: callback, context: context, ctx: context || this});
+      this._events = onApi(this._events || {}, name, callback, context, this);
       return this;
     },
 
@@ -107,48 +105,7 @@
     // callbacks for all events.
     off: function(name, callback, context) {
       if (!this._events || !eventsApi(this, 'off', name, [callback, context])) return this;
-
-      // Remove all callbacks for all events.
-      if (!name && !callback && !context) {
-        this._events = void 0;
-        return this;
-      }
-
-      var names = name ? [name] : _.keys(this._events);
-      for (var i = 0, length = names.length; i < length; i++) {
-        name = names[i];
-
-        // Bail out if there are no events stored.
-        var events = this._events[name];
-        if (!events) continue;
-
-        // Remove all callbacks for this event.
-        if (!callback && !context) {
-          delete this._events[name];
-          continue;
-        }
-
-        // Find any remaining events.
-        var remaining = [];
-        for (var j = 0, k = events.length; j < k; j++) {
-          var event = events[j];
-          if (
-            callback && callback !== event.callback &&
-            callback !== event.callback._callback ||
-            context && context !== event.context
-          ) {
-            remaining.push(event);
-          }
-        }
-
-        // Replace events if there are any remaining.  Otherwise, clean up.
-        if (remaining.length) {
-          this._events[name] = remaining;
-        } else {
-          delete this._events[name];
-        }
-      }
-
+      this._events = offApi(this._events, name, callback, context);
       return this;
     },
 
@@ -171,27 +128,17 @@
     // listen to an event in another object ... keeping track of what it's
     // listening to.
     listenTo: function(obj, name, callback) {
-      var listeningTo = this._listeningTo || (this._listeningTo = {});
+      if (!listenApi(this, 'listenTo', obj, name, callback)) return this;
       var id = obj._listenId || (obj._listenId = _.uniqueId('l'));
-      listeningTo[id] = obj;
-      if (!callback && typeof name === 'object') callback = this;
+      var listeningTo = this._listeningTo || (this._listeningTo = {});
+      listeningTo[id] || (listeningTo[id] = {obj: obj, events: {}});
       obj.on(name, callback, this);
+      listeningTo[id].events = onApi(listeningTo[id].events, name, callback, this);
       return this;
     },
 
     listenToOnce: function(obj, name, callback) {
-      if (typeof name === 'object') {
-        for (var event in name) this.listenToOnce(obj, event, name[event]);
-        return this;
-      }
-      if (eventSplitter.test(name)) {
-        var names = name.split(eventSplitter);
-        for (var i = 0, length = names.length; i < length; i++) {
-          this.listenToOnce(obj, names[i], callback);
-        }
-        return this;
-      }
-      if (!callback) return this;
+      if (!listenApi(this, 'listenToOnce', obj, name, callback) || !callback) return this;
       var once = _.once(function() {
         this.stopListening(obj, name, once);
         callback.apply(this, arguments);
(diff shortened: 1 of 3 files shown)


test/events.js
```diff
@@ -79,7 +79,7 @@
   test("listenTo and stopListening with event maps", 4, function() {
     var a = _.extend({}, Backbone.Events);
     var b = _.extend({}, Backbone.Events);
-    var cb = function(){ ok(true); };
+    var cb = function(){ console.log('called'); ok(true); };
```
megawac: Whoops :)

jridgewell: :smile:

jridgewell:
As a side: I have a completed [`eventsApi` refactor](https://github.com/jridgewell/backbone/compare/eventsApi-refactor) that cleans up the duplication in `listenApi`. I haven't included it in this PR yet, since it might be polarizing. If the only issue with this PR is the duplication, I'd be glad to put it in as well.

jashkenas:
Cleaning up the duplication would be lovely.

jridgewell:
Ok, it's included now.

backbone.js
```diff
@@ -224,27 +158,99 @@
...
+      if (events) triggerEvents(events, args);
+      if (allEvents) triggerEvents(allEvents, [name].concat(args));
+    }
+    return obj;
+  };
+
+  // Maps the normalized event callbacks into onceWrappers.
+  var onceMap = function(name, callback, offer) {
+    return eventsApi(function(map, name, callback, offer) {
+      if (callback) map[name] = onceWrap(name, callback, offer);
```
akre54: Any reason not to inline `onceWrap` here?

backbone.js
```diff
@@ -205,14 +136,17 @@
...
       if (!listeningTo) return this;
-      var remove = !name && !callback;
-      if (!callback && typeof name === 'object') callback = this;
-      if (obj) (listeningTo = {})[obj._listenId] = obj;
-      for (var id in listeningTo) {
-        obj = listeningTo[id];
-        obj.off(name, callback, this);
-        if (remove || _.isEmpty(obj._events)) delete this._listeningTo[id];
+
+      var listeneeIds = (obj) ? [obj._listenId] : _.keys(listeningTo);
```
akre54: unnecessary parens

backbone.js
```diff
@@ -224,27 +158,99 @@
...
       }
-      return false;
     }
+    return _.isEmpty(events) ? void 0 : events;
+  };
 
-    return true;
+  var triggerApi = function(obj, name, sentinel, args) {
+    if (obj._events) {
+      if (sentinel !== triggerApi) args = [sentinel].concat(args);
```
akre54: What's the sentinel about? I assume something with the `eventsAPI` works?

Some comments here (and elsewhere) would help.

backbone.js
jridgewell: Nope. I'll fix it.

backbone.js
jridgewell: That's because `trigger({ event: value}, 1, 2, 3)` will prepend `value` to the arguments of the listener. Re: https://github.com/jashkenas/backbone/issues/3460

akre54: Makes sense. I had a hard time tracking what's going on with the callbacks here, since the only other place I see this called, `triggerApi` is being used as the sentinel.

jridgewell:
Ping @akre54: I commented, and added a `triggerSentinel` object to prevent further confusion. Let me know you have any other suggestions.

backbone.js
```diff
@@ -159,11 +109,9 @@
     trigger: function(name) {
       if (!this._events) return this;
       var args = slice.call(arguments, 1);
-      if (!eventsApi(this, 'trigger', name, args)) return this;
-      var events = this._events[name];
-      var allEvents = this._events.all;
-      if (events) triggerEvents(events, args);
-      if (allEvents) triggerEvents(allEvents, arguments);
+
+      // Use `eventsApi` to normalize `name` into the proper event names.
```
akre54: Mind clarifying this? I'm not sure what this is getting at.

akre54:
I think a little more of the "why" and the "how" and less "what" in the comments might be helpful.

Maybe a sentence or two at the top briefly describing *how* the many abstracted functions work. Right now I'm on the fence that this consolidation is worth the added indirection. It definitely makes it harder to follow the logic. 

Can you also make a jsperf showing the extra functions aren't too horrible and squash to a single commit and I'll merge? Thanks for your hard work on this. (And no stress about getting it done today... enjoy the snowday!)

jridgewell:
Here's a simple jsperf: http://jsperf.com/eventsapi-refactor (It was hidden in 50adef3's commit message). Safari seems to be the only one taking a hit, but it's still very fast.

akre54:
how about vs master? That's showing all the changes with the extra
callbacks added in.

jridgewell:
Here's the simple case [against master](http://jsperf.com/eventsapi-refactor/3), and another using [space-separated events](http://jsperf.com/eventsapi-refactor/2).

I did the original tests against the `listenTo-memory-leak-immutable-on-events` branch, because that's the one with the minimum code to fix the memory leak. Comparing against `master` seemed apples-to-oranges.

jridgewell:
Actually, the performance hit seems to be coming from https://github.com/jashkenas/backbone/pull/3463, which is also merged into `eventsApi-refactor`. The [simple case](http://jsperf.com/eventsapi-refactor/4) against those two have the exact same speeds.

smelnikov:
Nice, but does not seem to compete with #3049 as it does not address the following: 

```javascript
var a = _.extend({}, Backbone.Events);
var b = _.extend({}, Backbone.Events);
var c = _.extend({}, Backbone.Events);

a.listenTo(c, 'x', fn);
b.listenTo(c, 'x', fn);
c.off('x');
c = null; //c can't be garbage collected until 'stopListening()' is called on a and b
```

Here are the relevant tests

```javascript
 test("off removes object from listeners", 6, function() {                                                                                                                                                                                    
    var a = _.extend({}, Backbone.Events);
    var b = _.extend({}, Backbone.Events);
    var c = _.extend({}, Backbone.Events);
    var d = _.extend({}, Backbone.Events);
    var fn = function() {};    
        
    b.listenTo(a, 'all', fn);  
    c.listenTo(a, 'all', fn);  
    c.listenTo(d, 'x', fn);    
      
    equal(_.size(b._listeningTo), 1);
    equal(_.size(c._listeningTo), 2);
    equal(_.size(a._listeners), 2); 
    a.off();                   
    equal(_.size(b._listeningTo), 0);
    equal(_.size(c._listeningTo), 1);
    equal(_.size(a._listeners), 0); 
  }); 
      
  test("off w/event name removes object from listeners", 8, function() {
    var a = _.extend({}, Backbone.Events);
    var b = _.extend({}, Backbone.Events);
    var c = _.extend({}, Backbone.Events);
    var fn = function() {};

    b.listenTo(a, 'x', fn);
    c.listenTo(a, 'x y', fn);

    equal(_.size(b._listeningTo), 1);
    equal(_.size(c._listeningTo), 1);
    equal(_.size(a._listeners), 2); 
    a.off('x');
    equal(_.size(b._listeningTo), 0);
    equal(_.size(c._listeningTo), 1);
    equal(_.size(a._listeners), 1); 
    a.off('y');
    equal(_.size(c._listeningTo), 0);
    equal(_.size(a._listeners), 0); 
  });
```

jridgewell:
Hmm, let me see what I can do to support that case.

jridgewell:
I've addressed @smelnikov's `#off` memory leak, but the code isn't as pretty as I would have liked.

backbone.js
akre54: @jridgewell the comments are still lacking. Can you do a pass to make sure the logic is clear? I'm finding it hard to follow what is happening and why.

jridgewell: I haven't revised the comments yet, was trying to solve the `off` problem first.

Is there a particular spot you're not able to follow?

backbone.js
```diff
@@ -106,49 +97,19 @@
...
-          delete this._events[name];
+      if (!this._events) return this;
+      this._events = eventsApi(offApi, this._events, name, callback, context);
+
+      var listeners = this._listeners;
+      if (listeners) {
+        var ids = context != null ? [context._listenId] : _.keys(listeners);
+        for (var i = 0, length = ids.length; i < length; i++) {
+          var listener = listeners[ids[i]];
+          if (!listener) break;
```
megawac: Why is this necessary? I would expect a `continue` if anything

backbone.js
akre54: The bits with `triggerSentinel` and the numerous internal functions mostly. 

It almost seems like we should drop the `Events = {on: ..., off: ..., listenTo: ...}` object, and just colocate the different events methods (`Events.on = ...; onApi = ...; Events.off = ...;`.) To follow this logic, I'm having to jump back and forth through the source to even understand where the different pieces are going.

There's so much extra overhead and indirection going on here, it really detracts from the readability of the Events code.

backbone.js
```diff
@@ -171,48 +130,33 @@
...
-        return this;
-      }
-      if (!callback) return this;
-      var once = _.once(function() {
-        this.stopListening(obj, name, once);
-        callback.apply(this, arguments);
-      });
-      once._callback = callback;
-      return this.listenTo(obj, name, once);
+      name = onceMap(name, callback, _.bind(this.stopListening, this, obj));
```
megawac: Should it not also be bound to `name, callback`?

backbone.js
```diff
@@ -223,28 +167,129 @@
...
+    if (name && typeof name === 'object') {
+      // Handle event maps.
+      for (names = _.keys(name), length = names.length; i < length; i++) {
+        events = api(events, names[i], name[names[i]], context, ctx);
       }
-      return false;
+    } else if (name && eventSplitter.test(name)) {
+      // Handle space separated event names.
+      for (names = name.split(eventSplitter), length = names.length; i < length; i++) {
+        events = api(events, names[i], callback, context, ctx);
```
megawac: I feel like the redundancy in these two can be addressed but it can be done in a follow up

backbone.js
jridgewell: When you `listenTo`, context is forced to be `this`. If we're `off`ing, we need to consider two circumstances:

1. No context
    - We'll loop through all our listeners, since any could be affected.
    - This conditional will never be true, so we'll never `break`.
2. Context
    - Only loop through the listeners of the corresponding `_listenId`, since **only** that listener will have the same `context`
    - If we have `listeners`, but not that `_listenId`, then `break`.
      - `ids` is guaranteed to only equal `[context._listenId]`, so there's no use `continue`ing.
    - If we have that `_listenId`, all is good.

backbone.js
```diff
@@ -223,28 +167,129 @@
...
+      events = api(events, name, callback, context, ctx);
+    }
+    return events;
+  };
+
+  // Handles actually adding the event handler.
+  var onApi = function(events, name, callback, context, ctx) {
+    if (callback) {
+      var handlers = events[name] || [];
+      events[name] = handlers.concat({callback: callback, context: context, ctx: context || ctx});
```
megawac: What was the reason you didn't want to use push again, I saw the issue but didnt read it?

backbone.js
jridgewell: Can't, because `name` could be an object or space-separated string. This will be passed in the normalized (just a single event name and it's callback) by the [onceWrapper](https://github.com/jashkenas/backbone/pull/3455/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR286).

backbone.js
```diff
@@ -223,28 +167,129 @@
...
+      name = names[i];
 
-    // Handle space separated event names.
-    if (eventSplitter.test(name)) {
-      var names = name.split(eventSplitter);
-      for (var i = 0, length = names.length; i < length; i++) {
-        obj[action].apply(obj, [names[i]].concat(rest));
+      // Bail out if there are no events stored.
+      var handlers = events[name];
+      if (!handlers) break;
```
megawac: Again, I think this should be a `continue`

backbone.js
```diff
@@ -223,28 +167,129 @@
...
+              callback !== handler.callback._callback ||
+                context && context !== handler.context
+          ) {
+            remaining.push(handler);
+          }
+        }
       }
-      return false;
+
+      // Replace events if there are any remaining.  Otherwise, clean up.
```
megawac: Nit: double space

backbone.js
jridgewell: I agree the source could be better arranged, I'll work on that tonight along with the comments.

> There's so much extra overhead and indirection going on here

I actually thought it was kind of easier to follow without the old `eventsApi` behavior, but then I had to account for `off` cleaning up any listeners. It makes it easier in my mind to think of all the `*Api` functions as special purpose reduce iteratees, with `eventsApi` being `_.reduce`.

backbone.js
jridgewell: Agh. I think we've decided to `slice` in `#trigger` instead of `concat` here. I'll need to update this.

backbone.js
jridgewell: It's the same kind of logic as the other. The only way the conditional will ever be true is if you pass the `name` argument, in which case `names = [name]`. So, no need to continue.

backbone.js
jridgewell: :stuck_out_tongue:

jridgewell:
@akre54 Commented and rearranged.

jashkenas:
I think let's maybe merge this for now -- and then go over it and tweak as need be. (But hopefully need not ;)

PR #3456, opened 2015-01-22
Clarify docs around View#el. Fixes #3454

### LINKED ISSUE #3454: Confusing docs for View.el
The docs for View.el say:

> this.el is created from the view's tagName, className, id and attributes properties, if specified. If not, el is an empty div.

Which is contradicted by the example:

```javascript
var BodyView = Backbone.View.extend({
  el: 'body'
});
```

since here, `el` is assigned a string, which seems to make it equivalent to `tagName`.  It's unclear why you would use `el` in this case instead of `tagName`

akre54:


braddunbar:
:+1: Looks good.

index.html
```diff
@@ -2598,9 +2598,11 @@ <h2 id="View">Backbone.View</h2>
       whether they've already been inserted into the page or not. In this
       fashion, views can be rendered at any time, and inserted into the DOM all
       at once, in order to get high-performance UI rendering with as few
-      reflows and repaints as possible. <tt>this.el</tt> is created from the
-      view's <tt>tagName</tt>, <tt>className</tt>, <tt>id</tt> and <tt>attributes</tt> properties,
-      if specified. If not, <b>el</b> is an empty <tt>div</tt>.
+      reflows and repaints as possible. <tt>this.el</tt> is either a DOM selector
+      string or Element, or created from the view's <tt>tagName</tt>, <tt>className</tt>,
+      <tt>id</tt> and <tt>attributes</tt> properties, if specified. If not,
```
megawac: Can you link these properties where possible? 

http://backbonejs.org/#View-attributes

also that comma there is sort of awkward. I would word this

> `this.el` can be resolved from a DOM selector string or an Element; otherwise it will be created from the view's `tagName`, `className`, `id` and `attributes` properties specified.

thejameskyle:
:+1:

PR #3459, opened 2015-01-23
Collection#set - options.at accepts a string.

braddunbar:
Closes #3353.

I noticed this one still sitting around so I added a patch with a test.


tgriesser:
:+1:

PR #3463, opened 2015-01-24
Do not alter the "on" events array

jridgewell:


akre54:
What does this give us?

jridgewell:
The test was a false positive. "bind does not alter callback list" and "unbind does not alter callback list".

akre54:
Couldn't we also `slice` on [L163](https://github.com/jashkenas/backbone/blob/85885130a8f290a85bb838281849741d2c3d6b79/backbone.js#L163) (or in triggerEvents, etc)? What's the benefit to re-assigning the events hash?

jridgewell:
> Couldn't we also slice on L163

Yup. I think the real point of these tests was to make sure `off`ing didn't cause an array out of bounds error.

akre54:
Totally, and agreed. Can we just add the tests without making changes at the moment, or are you saying the tests would be broken too? If that's the case my personal preference would be to slice over the array in trigger, rather than in on, since it just *feels* more correct. `this._events` isn't immutable, and we really shouldn't be reassigning it every time `on` is called.

braddunbar:
> Yup. I think the real point of these tests was to make sure offing didn't cause an array out of bounds error.

Close, but not quite. The point of the test is to make sure `off` doesn't mutate the array in a way that causes specific event handlers to be skipped (a hard fought debugging session!). Adding events to the end of the handler list is fine. I imagine this would have some fairly serious performance ramifications as well.

> Couldn't we also slice on L163 (or in triggerEvents, etc)? What's the benefit to re-assigning the events hash?

I'd prefer this approach if it's an issue. It was once done this way and I don't remember why it changed. :confused:

jridgewell:
The tests would be broken if this were merged without the `concat`. Though, that would be because the test's purpose isn't what it is actually testing.

jridgewell:
> The point of the test is to make sure off doesn't mutate the array in a way that causes specific event handlers to be skipped (a hard fought debugging session!).

That would make sense too :smile:. Let me look into this again.

braddunbar:
Coolio. Thanks @jridgewell! :smiley:

jridgewell:
Ping. Let me know how you feel about it now.

Issue #3460, opened 2015-01-23
Trigger with object


jridgewell:
Currently, calling `#trigger` with an object will prepend the prop's value to the arguments that get triggered. Is this the desired functionality? I can't find any tests for this behavior.

```javascript
obj.on('event', function() { console.log(arguments); });

obj.trigger({ event: 'first arg' }, 1, 2, 3);
// Logs ['first arg', 1, 2, 3]
```

braddunbar:
Hey @jridgewell! Thanks for pointing it out. I think the docs do cover this, though it's a bit removed from `trigger`.

> All Backbone event methods also support an event map syntax, as an alternative to positional arguments:

It doesn't seem to make much sense for `trigger` though. :confused:

jridgewell:
> It doesn't seem to make much sense for trigger though

I agree. But, my issue wasn't with the object syntax, but that the value is prepended to the arguments.

I'm working on refactoring `eventsApi` (and `listenApi` post https://github.com/jashkenas/backbone/pull/3455), and this is one of the gotchas I noticed. There are two others, but I'll open separate Issues for those.

akre54:
Do we actually need this? I'm not sure there's ever a situation where we'd want to trigger using an object, right?

I think the reason it works is more a result of using `eventsApi` under the hood, no reason to make this official codified behavior.

jridgewell:
> I think the docs do cover this, though it's a bit removed from trigger.

Actually, the [`#trigger` docs](http://backbonejs.org/#Events-trigger) only list single or space-separated events as supported.

> I think the reason it works is more a result of using eventsApi under the hood, no reason to make this official codified behavior.

With the old `eventsApi`, it was indeed just an implementation detail. With my new reducing `eventsApi`, it's something I've had to [hack in](https://github.com/jashkenas/backbone/blob/master/backbone.js#L294-L296).

I'd love to see this behavior removed, and #3465 reverted.

jashkenas:
Feel free to remove and revert it in your reduced API.

jridgewell:
Closed with https://github.com/jashkenas/backbone/pull/3499.

PR #3465, opened 2015-01-25
Test trigger arguments behavior

jridgewell:
To be clear, I think the trigger with object behavior is silly. But I needed tests to ensure that my [`eventsApi` refactor](https://github.com/jridgewell/backbone/tree/eventsApi-refactor) kept the current functionality.

Re: https://github.com/jashkenas/backbone/issues/3460


jridgewell:
Re: https://github.com/jashkenas/backbone/issues/3460

Issue #3300, opened 2014-09-08
jquery as deps


sshmyg:
Hey, is jQuery hard deps for BB?

PR #3467, opened 2015-01-29
Fix missing jquery dependency using commonJS

schorfES:


Florian-R:
Dupe of #2997 #3300 gh-3421 and probably a lot of others.

PR #3469, opened 2015-01-29
[ADDED] events object to bind events to model and collection

jjviscomi:
[UPDATING] issue #3468 to correct tests


akre54:
Duplicate of #2240 and others.

delegateEvents is a fundamentally different beast from model / entity events and should be treated differently too.

As a side note, you don't have to re-open pull requests to make changes. Just push your changes to the same branch and github will auto-update the pull for you.

PR #3476, opened 2015-02-05
Implement Events as a linked list

### LINKED ISSUE #3466: During event execution, stopListening call on another object does not prevent handler from executing
If there are two related objects listening to the same event, it appears that during event execution one object can not call stopListening on another object and prevent the handler callback during the same execution. Looks like the events get cached somewhere in the trigger and stopListening does not change that.

Let me illustrate with an example:

    var parent = new Backbone.Model();
    var child = new Backbone.Model();
    var vent = new Backbone.Model();
    
    parent.listenTo(vent, "alert", function(){
        alert("parent");
        child.stopListening();
    });
    
    child.listenTo(vent, "alert", function(){
        alert("child");
    });
    
    vent.trigger("alert");

My expectation is that if parent handler is called first, the child alert should not come up. However both alerts are issued.

http://jsfiddle.net/8y4b1ruv/1/

Backbone.js 1.1.2

jridgewell:
[jsperf](http://jsperf.com/backbone-events-linked-list/18)

Roughly equivalent performance. Using a linked list allows the events to be mutated by trigger, without skipping an event. This is a **breaking change**.

Events used to use a linked list (and switched to an array in #1284). But, it performed [very badly](http://jsperf.com/backbone-events-linked-list/8). I'm fairly certain that was due to [L106](https://github.com/jashkenas/backbone/blob/6c3d3838e10d5a070f0d1f604fb5c99ed34c8c93/backbone.js#L106) recreating the linked list's object every `#on`, and [L138](https://github.com/jashkenas/backbone/blob/6c3d3838e10d5a070f0d1f604fb5c99ed34c8c93/backbone.js#L138) calling `#on` to re-bind callbacks inside `#off`.

This fixes #3466 and closes #3463.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -287,10 +287,10 @@
   // `offer` unbinds the `onceWrapper` after it as been called.
   var onceMap = function(map, name, callback, offer) {
     if (callback) {
-      var once = map[name] = _.once(function() {
+      var once = map[name] = function() {
         offer(name, once);
         callback.apply(this, arguments);
-      });
+      };
       once._callback = callback;
     }
     return map;
--- a/backbone.js
+++ b/backbone.js
@@ -138,7 +138,7 @@
   var onApi = function(events, name, callback, context, ctx) {
     if (callback) {
       var list = events[name] || (events[name] = {});
-      var ev = {callback: callback, context: context, ctx: context || ctx, next: false};
+      var ev = {callback: callback, context: context, ctx: context || ctx, next: false, skip: false};
       var tail = list.tail || list;
       list.tail = tail.next = ev;
     }
--- a/backbone.js
+++ b/backbone.js
@@ -146,7 +146,7 @@
   // The reducing API that adds a callback to the `events` object.
   var onApi = function(events, name, callback, context, ctx, listening) {
     if (callback) {
-      var list = events.lists[name] || (events.lists[name] = {name: name, tail: void 0, next: void 0});
+      var list = events.lists[name] || (events.lists[name] = {tail: void 0, next: void 0});
       if (!list.next) events.count++;
       if (listening) listening.count++;
 
--- a/backbone.js
+++ b/backbone.js
@@ -303,7 +303,7 @@
     var args = Array(length);
     for (var i = 0; i < length; i++) args[i] = arguments[i + 1];
 
-    var alreadyTriggering = events.alreadyTriggering;
+    var alreadyTriggering = events.triggering;
     events.triggering = true;
 
     eventsApi(triggerApi, events.lists, name, void 0, args);
--- a/backbone.js
+++ b/backbone.js
@@ -191,7 +191,7 @@
       // listening to obj. Break out early.
       if (!listening) break;
 
-      if (listening.obj) listening.obj.off(name, callback, this);
+      listening.obj.off(name, callback, this);
       if (!listening.count) delete listeningTo[id];
     }
 
--- a/backbone.js
+++ b/backbone.js
@@ -315,11 +315,11 @@
   // Handles triggering the appropriate event callbacks.
   var triggerApi = function(lists, name, cb, args) {
     var list = lists[name];
-    var allList = lists.all;
     if (list) {
       triggerEvents(list, args);
       if (!list.next) delete lists[name];
     }
+    var allList = lists.all;
     if (allList) {
       triggerEvents(allList, [name].concat(args));
       if (!allList.next) delete lists.all;
--- a/backbone.js
+++ b/backbone.js
@@ -246,30 +246,28 @@
   // the callback is invoked, it will be removed.
   Events.once =  function(name, callback, context) {
     // Map the event into a `{event: once}` object.
-    var events = onceMap(name, callback, _.bind(this.off, this));
+    var events = eventsApi(onceMap, {}, name, callback, _.bind(this.off, this));
     return this.on(events, void 0, context);
   };
 
   // Inversion-of-control versions of `once`.
   Events.listenToOnce =  function(obj, name, callback) {
     // Map the event into a `{event: once}` object.
-    var events = onceMap(name, callback, _.bind(this.stopListening, this, obj));
+    var events = eventsApi(onceMap, {}, name, callback, _.bind(this.stopListening, this, obj));
     return this.listenTo(obj, events);
   };
 
   // Reduces the event callbacks into a map of `{event: onceWrapper}`.
   // `offer` unbinds the `onceWrapper` after it as been called.
-  var onceMap = function(name, callback, offer) {
-    return eventsApi(function(map, name, callback, offer) {
-      if (callback) {
-        var once = map[name] = _.once(function() {
-          offer(name, once);
-          callback.apply(this, arguments);
-        });
-        once._callback = callback;
-      }
-      return map;
-    }, {}, name, callback, offer);
+  var onceMap = function(map, name, callback, offer) {
+    if (callback) {
+      var once = map[name] = _.once(function() {
+        offer(name, once);
+        callback.apply(this, arguments);
+      });
+      once._callback = callback;
+    }
+    return map;
   };
 
   // Trigger one or many events, firing all bound callbacks. Callbacks are
--- a/backbone.js
+++ b/backbone.js
@@ -152,7 +152,6 @@
         list = events.lists[name] = {
           count: 0,
           triggering: false,
-          shouldDelete: false,
           tail: void 0,
           next: void 0
         };
@@ -249,11 +248,7 @@
       list.tail = tail;
       if (tail === list) {
         events.count--;
-        if (list.triggering) {
-          list.shouldDelete = true;
-        } else {
-          delete lists[name];
-        }
+        if (!list.triggering) delete lists[name];
       }
     }
     return events;
@@ -311,11 +306,11 @@
       var allList = lists.all;
       if (list) {
         triggerEvents(list, args);
-        if (list.shouldDelete) delete lists[name];
+        if (!list.next) delete lists[name];
       }
       if (allList) {
         triggerEvents(allList, [name].concat(args));
-        if (allList.shouldDelete) delete lists.all;
+        if (!allList.next) delete lists.all;
       }
     }
     return obj;
--- a/backbone.js
+++ b/backbone.js
@@ -87,7 +87,7 @@
   // Passes a normalized single event name and callback, as well as the `context`,
   // `ctx`, and `listening` arguments to `iteratee`.
   var eventsApi = function(iteratee, memo, name, callback, context, ctx, listening) {
-    var i = 0, names, length;
+    var i = 0, names;
     if (name && typeof name === 'object') {
       // Handle event maps.
       for (names = _.keys(name); i < names.length; i++) {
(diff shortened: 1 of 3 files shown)


jridgewell:
Note: This'll (hopefully) allow me to get really clever with cleaning up #3455.

jashkenas:
Let's get a better JS perf going.

We don't care about how long it takes to `.on` or `.off` (so much) — we care about how long it takes to trigger.

Try one with a pre-event-ful object with maybe 15 or so different empty functions bound to different events. Then try triggering a half-dozen of them within the test loop. Let's see.

jridgewell:
Here's the jsperf [split into the different components](http://jsperf.com/backbone-events-linked-list/29). The only one that took a hit was `on`, and I think that's just because the `ev` object got bigger.

This PR got a bit muddled, but I think it's still good. In order to support _any_ mutations on the event list, I couldn't delete `_events.eventName`. In order to keep the speed of #2858, I had to then keep track of the number of events that currently have listeners. With that, @smelnikov's listener counting became a more obvious solution (and faster, since two different `events` hash didn't have to be traversed).

jashkenas:
Uh ... it looks like you're triggering `'e' + i` event names ... but you're binding `'' + i` event names. So the trigger test isn't actually triggering anything?

jridgewell:
Well, that's embarrassing.

Seems `trigger` takes a 50% hit :frowning:. Let me see what I can do to fix this.

smelnikov:
Check it out on jsperf real quick and noticed that the event keys are not being cleared when all of events of a certain type are removed. Was this intentional? Not sure if this be classified as a leak since events are typically statically defined but it's not how it behaved before.


```javascript
Backbone.on('a b c d e f g h i j k l m n o p q r s t u v w x y z', function() {}, {});
console.log(String(_.keys(Backbone._events))); //a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z
Backbone.off();
console.log(String(_.keys(Backbone._events))); //undefined

LL.on('a b c d e f g h i j k l m n o p q r s t u v w x y z', function() {}, {}); 
console.log(String(_.keys(LL._events.lists))); //a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z
LL.off();
console.log(String(_.keys(LL._events.lists))); //a,b,c,d,e,f,g,h,i,j,k,l,m,n,o,p,q,r,s,t,u,v,w,x,y,z
```

jridgewell:
> Not sure if this be classified as a leak since events are typically statically defined but it's not how it behaved before.

The list is still present, but any events will be removed.

```javascript
LL.on('a', function cb() {});
LL._events.lists.a.next; // => cb Event Node
LL.off();
LL._events.lists.a.next; // => undefined
```

smelnikov:
> The list is still present, but any events will be removed.

Yeah, I saw that. I meant previously it wasn't keeping around an empty array when all events of a certain type were removed. Was there a use case for that behaviour that this PR will no break and thus introduce memory leaks in same rare circumstances?

jridgewell:
Ah ha! My test was faulty. Every run was adding a few dozen events listeners, which threw off the actual performance testing.

It's fixed now and everything has similar performance, except for `off` which is dramatically faster.

- - -

For giggles, there's also an implementation using a back-linked array. But, it's much slower at just about everything.

jridgewell:
> Was there a use case for that behaviour that this PR will no break and thus introduce memory leaks in some rare circumstances?

There won't be a memory leak, but it could ballon if you create a ginormous amount of event names then off everything. But it's still just an object `{ count: 0, tail: undefined, next: undefined}`, so I think you're more likely to hit other limits before this becomes a problem.

jashkenas:
> There won't be a memory leak, but it could ballon if you create a ginormous amount of event names then off everything

We should amend this PR to fix this before looking at the JSPerf and evaluating.

Some folks have Backbone apps that they leave open for days and days and days ... and it is a valid use case to add dynamically-generated event names. We shouldn't leak bookkeeping objects if possible.

jridgewell:
Updated. It adds a bit of checking, since we can't delete the list while triggering.

Imagine you off all events during a listener then add a new one. The `triggerEvents` loop would need to back track all the way to the List's object, then start again to grab the new listener. If we instead deleted during the trigger, it the new event would be added to a new List and the loop would skip it.

jashkenas:
Okay, so now the problem is this:

image

We have a new implementation that is not-quite-but-almost-as-fast as the current one, that adds a considerable amount of code and complexity, that solves a thing you probably shouldn't be doing in the first place.

I'm all for experimentation and improvement to our events system, but I'm not sure that this is, on balance, an improvement.

jridgewell:
Understandable. The code bloat is only partly due to the linked-lists, though. The rest is from refinements of the `eventsApi` refactor. For reference, here's _only_ the [refinements](https://github.com/jridgewell/backbone/compare/eventsApi-refinements).

eventsApi-refinements

I've got to hand it to @smelnikov, the listener counting system was a great idea. I wish I had taken that route, originally.

smelnikov:
> I've got to hand it to @smelnikov, the listener counting system was a great idea. I wish I had taken that route, originally.

hey thanks! I agree but then again I might be a little biased ;)

Maybe we can revisit #3049. Revert #3455 and then @jridgewell can apply some of his commits within #3455 that increased performance/readability. 

I also gave a bit of thought to mutating events within a trigger and while this is a good implementation, I don't think it's a problem that should be solved.

I think that if you bind say an event listener to an event `'foo'` while `'foo'` is already in the process of triggering, than you have already missed the occurrence of the `'foo'` event and it makes sense for your newly bound event listener not to fire until the next occurrence. It's easier to reason about anyway.

`off` is a bit trickier but I think we could apply the same concept there as well. 

Whatever utility might be gained by mutating events within a trigger is not going to be worth the very significant increase of complexity to the programmers mental model (the one working with backbone, not on it.).

 Besides, Backbone events communication system already has a bit of a reputation for being hard to reason about, this will just kick it up a few notches.

thejameskyle:
Re: Number of additions/deletions

**Before 713c288475fdbb4a56e7bcd5296f3a69ba5b9ad3**
6.96KB gzipped (20.42KB uncompressed)

**After ae56ec5b3ca03bee77ca02d9640d75d05f2f0903**
7.17KB gzipped (20.98KB uncompressed)

210 bytes doesn't seem like a huge deal

jridgewell:
> Maybe we can revisit #3049. Revert #3455 and then @jridgewell can apply some of his commits within #3455 that increased performance/readability.

That's essentially what the [refinements branch](https://github.com/jridgewell/backbone/compare/eventsApi-refinements) is.

> I think that if you bind say an event listener to an event 'foo' while 'foo' is already in the process of triggering, than you have already missed the occurrence of the 'foo' event and it makes sense for your newly bound event listener not to fire until the next occurrence. It's easier to reason about anyway.

Except when you bind other events. IE, you can add to the `all` event (unless it's already empty!), or to another event that happens later in the same trigger (`trigger('event event2')`).

I think the code complexity is a valid criticism, but the mental model is actually simplified this way. The `all` edge cases (it applies to `off` as well) go away, and we're left with just mutable event listeners in every circumstance. Callbacks will always be called in the order the were bound and unbinding stops them from being called.

smelnikov:
> That's essentially what the refinements branch is.

Ah, nice!

> Except when you bind other events. IE, you can add to the all event (unless it's already empty!), or to another event that happens later in the same trigger (trigger('event event2')).

I don't think `trigger('event event2')` is a problem as it can be seen as sugar for `trigger('event').trigger('event2')`. 

The issue with `all` does sound like a bug though, but it looks like it can be fixed by passing in a max length limit to `triggerEvents` since events are always added to the end and when removed it's a new reference anyway.

disruptek:
> I think the code complexity is a valid criticism, but the mental model is actually simplified this way.

Strongly agree.  I've found several bugs in projects using Backbone where even seasoned Backbone users were caught out by the immutable-listeners issue.

When using events, one should not need to think about how Backbone implements iteration over the listeners and whether that mechanism is mutable.  The simple model is the one documented, but it doesn't withstand all use-cases, so subtle bugs appear which are apparently difficult even for pros to puzzle out.

smelnikov:
> Strongly agree. I've found several bugs in projects using Backbone where even seasoned Backbone users were caught out by the immutable-listeners issue.

Do you believe this because of the edge cases and bugs that surround the current implementation, or do you think that having events mutate is they are triggering is easier to reason about than having an immutable set of events at a point in time that are all guaranteed to trigger for any given trigger call?

disruptek:
I don't believe it's an issue of edge cases and bugs, but on the other hand, when reviewing Backbone's implementation I take it on faith that current behavior is in such widespread use as to defy a "bug" label.  I'm pretty lazy, but I don't see the campaigning involved to "fix" what I consider to be design flaws (at best) to be a worthwhile expenditure of time.  That's why I chimed in here; it looks like a breaking change may have some traction, and I just want to provoke more support if at all possible.

My opinion is that the documentation is that which we must impose upon the user to reason with, not the implementation.  If we were to make the documentation complex enough to explain the current behavior, then fewer bugs would likely be created.  I'd prefer not to document this complexity; therefore, I'd prefer not to implement this complexity.

To be clear, it's not just my aforementioned laziness; I actually have no idea how to reason about immutable-listeners in any sufficiently complex system.  Here's my documented demonstration of a similar design problem: marionettejs/backbone.wreqr#35

The output in this tiny example isn't just tricky to anticipate, it's much more complex than we want to document, and the design in that case is simpler.  The situation goes downhill fast when you start building complex Backbone apps, and again, neither auditing of your code nor reading Backbone's documentation is liable to be helpful, because the complexity arises purely from the event system implementation.

jashkenas:
In theory, I'd like to merge this — but I'm a bit too nervous about mucking with events internals to this extent without one of the two following things:

  * A simpler implementation that can be read through easily. (Especially when, conceptually, it seems like a simple implementation should be possible with a bit of cleverness.)

 * Other Backbone contributors looking at this patch and signing off on what its doing / the general approach. @braddunbar? @tgriesser? @megawac?

jridgewell:
This PR is in need of a serious cleanup. I'll get to that tonight.

braddunbar:
> Using a linked list allows the events to be mutated by trigger

Can you explain this a bit with a simple use case? I'm not sure what you mean or when you'd need to. This is a fairly large change but I'm not seeing the immediate benefits.

tgriesser:
I think @braddunbar would have a better opinion on this than I, but to be honest after looking through the tickets and the added tests I don't know I fully follow the actual issue.

I don't necessarily agree that the behavior of #3466 should be guaranteed, mainly because it feels like a huge anti-pattern to rely on order of execution guarantees in event behavior like that.

Also, the first thing I think when I hear "Allowing events to be mutated by a trigger", is big red flag and source of incidental complexity, not feature.

So while implementing events with a linked list sounds generally nice, I'm not sold on the reasons behind it and I'm not sure it warrants the amount of noise it adds to the internals.

Maybe a better real world case of what this is trying to solve would change my mind?

jridgewell:
Closing this. I was planning on using this to build "shortcut" linked lists, so that `#stopListening` could iterate an object's  the listener `EventNodes`, updating `ev.next` and the next node's `ev.prev` so that `#stopListening` only had to do one pass of events. The speed was actually much worse. :confused:

PR #3480, opened 2015-02-08
Fix #3477. Clarify Model.fetch docs

### LINKED ISSUE #3477: Model.fetch docs
The docs for Model.fetch are misleading.  They say that fetch() should reset a model's attributes from the server request, but if attribute X exists on the model, and is not sent by the server, the model retains X.  This is not a reset.  It's a set.

wettowelreactor:
Modified the docs for Model.fetch to describe what happens to attributes set on 
the model but not sent by the server. This should address BUG #3477

Before:
Fetch the model from the server. If the server's representation of the
model differs from its current attributes, they will be overridden,
triggering a `"change"` event.

After:
Fetch the model from the server. If the models differ the two are merged. 
Conflicting attributes are overridden by the server's model, triggering a 
`"change"` event.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -518,9 +518,7 @@
       return _.clone(this._previousAttributes);
     },
 
-    // Fetch the model from the server. If the server's representation of the
-    // model differs from its current attributes, they will be overridden,
-    // triggering a `"change"` event.
+    // Fetch the model from the server. If the models differ the two are merged.
     fetch: function(options) {
       options = options ? _.clone(options) : {};
       if (options.parse === void 0) options.parse = true;
--- a/index.html
+++ b/index.html
@@ -1219,7 +1219,7 @@ alert(JSON.stringify(artist));
     <p id="Model-fetch">
       <b class="header">fetch</b><code>model.fetch([options])</code>
       <br />
-      Resets the model's state from the server by delegating to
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -518,9 +518,9 @@
       return _.clone(this._previousAttributes);
     },
 
-    // Fetch the model from the server. If the server's representation of the
-    // model differs from its current attributes, they will be overridden,
-    // triggering a `"change"` event.
+    // Fetch the model from the server. If the models differ the two are merged.
```
akre54: Probably don't even need the "if the models differ" part. It's merged from the server response regardless.

akre54:
Thanks for this. You might want to also alter the `index.html` docs too since it also references the "overrides" terminology.

wettowelreactor:
What about something like:
https://github.com/wettowelreactor/backbone/compare/patch-1

backbone.js
```diff
@@ -518,9 +518,7 @@
...
```
akre54: Try:

> Fetch the model from the server, merging the response with the model's local attributes. Any changed attributes will trigger a "change" event.

akre54:
Cool, thanks. Will fix up in a sec.

PR #3484, opened 2015-02-10
Include latest options in sync callbacks (in progress)

### LINKED ISSUE #3478: Collection#create doesn't properly pass final options to success callback
In https://github.com/jashkenas/backbone/blob/master/backbone.js#L986 `Collection#create` wraps the `success` callback (in order to handle `wait: true`). But the wrapper does not take the third argument, `options` -- when it calls the wrapped callback it instead just sends it the options `Collection#create` received. This breaks any custom sync code that might be modifying the options further downstream (for instance, it breaks https://github.com/nilbus/Backbone.dualStorage, which adds `dirty: true` to the options if the sync to server failed but succeeded to localStorage instead).

I don't know if this type of signaling via `options` is supported, but I can't see why it would hurt for the wrapper callback here to actually take the options the callback was called with and pass those along, instead of short-circuiting with the original options. Would a PR for this change be acceptable?

platinumazure:
Fixes #3478.

Please do let me know if I can improve the tests. (I was a bit constrained by the fake AJAX non-implementation. I'm too used to sinon.js/buster.js unfortunately.)


jridgewell:
I think this will need to apply to the other `options.success` functions to be accepted.

[L527-L532](https://github.com/platinumazure/backbone/blob/include-latest-options-in-create/backbone.js#L527-L532), [L571-L582](https://github.com/platinumazure/backbone/blob/include-latest-options-in-create/backbone.js#L571-L582), etc...

platinumazure:
I have no problem with making the behavior consistent by any means, but I was trying to fix the reported issue.

I'll add some commits to fix the other parts later on.

PR #3486, opened 2015-02-10
collection::count()

redexp:
Often we need exactly count of models, not array of them. This method will take less time and memory to do such job. Also it more semantic then `.where({...}).length`

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -848,6 +848,18 @@
       return this.where(attrs, true);
     },
 
+    // Return count of models with matching attributes.
+    count: function (attrs) {
+      var count = 0;
+      this.forEach(function (model) {
+        for (var key in attrs) {
+          if (attrs[key] !== model.get(key)) return;
+        }
+        count++;
+      });
+      return count;
+    },
+
     // Force the collection to re-sort itself. You don't need to call this under
     // normal circumstances, as the set will maintain sort order as each item
     // is added.
--- a/test/collection.js
+++ b/test/collection.js
@@ -547,7 +547,7 @@
     equal(JSON.stringify(col), '[{"id":3,"label":"a"},{"id":2,"label":"b"},{"id":1,"label":"c"},{"id":0,"label":"d"}]');
   });
 
-  test("where and findWhere", 8, function() {
+  test("where, findWhere and count", 14, function() {
     var model = new Backbone.Model({a: 1});
     var coll = new Backbone.Collection([
       model,
@@ -562,6 +562,12 @@
     equal(coll.where({b: 1}).length, 0);
     equal(coll.where({b: 2}).length, 2);
     equal(coll.where({a: 1, b: 2}).length, 1);
+    equal(coll.count({a: 1}), 3);
+    equal(coll.count({a: 2}), 1);
+    equal(coll.count({a: 3}), 1);
+    equal(coll.count({b: 1}), 0);
+    equal(coll.count({b: 2}), 2);
+    equal(coll.count({a: 1, b: 2}), 1);
     equal(coll.findWhere({a: 1}), model);
     equal(coll.findWhere({a: 4}), void 0);
   });


platinumazure:
This would probably do better to take advantage of flexible callbacks. (For example, why can't this also support a `_.filter`-style callback?)

Also, I wonder if this should maybe be proposed into Underscore instead and then let Backbone consume it?

jridgewell:
This should more likely be an PR against underscore (it'd be more useful there). However, [two](https://github.com/jashkenas/underscore/pull/1240) [issues](https://github.com/jashkenas/underscore/issues/1509) have already been brought up, both closed.

Side note: I think Ruby's functional methods are the :bomb:. Might not hurt to match its [`Enumerable#count` API](http://www.ruby-doc.org/core-2.2.0/Enumerable.html#method-i-count).

megawac:
If there was an easier way to mixin methods this could also be implemented as

```js
Backbone.Collection.mixin({
   size: _.compose(_.size, _.filter)
});
```

Florian-R:
> If there was an easier way to mixin methods this could also be implemented as

I miss something or a good ol' extend does the trick?

```js
var BaseCollection = Backbone.Collection.extend({
  size: _.compose(_.size, _.filter)
});
```

megawac:
No, because you need to get collection.models first

You would need to compose with bb.prototype.filter

redexp:
@platinumazure unit tests are in pr
@jridgewell it can't be in underscore, because it backbone specific like `collection.where()` 
@megawac @Florian-R my main idea is to create method which will not create array and extra loops

jridgewell:
It definitely can: `Collection#where` is just `Collection#filter` with `first` sugar, and `#filter` comes from [underscore](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1078-L1093).

platinumazure:
@redexp My mistake, I see them now.

@jridgewell I assume @redexp is referring to [this definition][1] in the Backbone source. Even so, I see no reason this couldn't be implemented there, even if we end up needing to re-implement in Backbone for whatever reason.

  [1]: https://github.com/jashkenas/backbone/blob/956138fe5f17625dd850237a4519c3e60c86bc5b/backbone.js#L921-926

redexp:
@jridgewell you mean something like this?
```javascript
count: function (attrs) {
  return _.count(this.models, function (model) {
    return model.matches(attrs);
  });
}
```
or it can accept both object or function
```javascript
count: function (cb) {
  if (_.isObject(cb)) {
    var attrs = cb;
    cb = function (model) {
      return model.matches(attrs);
    };
  }
  return _.count(this.models, cb);
}
```

platinumazure:
I forget if `_.iteratee` is exposed or not yet, but if I were you I would just use that.

megawac:
In 1.7 iteratee is exposed, so the object check would not be necessary

It also wouldn't work because a function is an object :)

jridgewell:
@platinumazure Exactly. Using `_.iteratee` will create a matches function for you if `cb` is an object, or it will optimize calling `cb` if it's a function and there's a context.

Master has a psuedo dependency on 1.7 since https://github.com/jashkenas/backbone/pull/3307. @megawac, why wan't it bumped in package.json?

megawac:
What do you mean @jridgewell it got bumped to 1.6 cause of that pr

jridgewell:
Whoops, I thought `_.matches` was a 1.7 feature. :smile:

redexp:
`where(...).length` 5 times slower `count(...)` and `count` (which uses `model::matches`) is 2 times slower then `where(...).length`
http://jsperf.com/backbone-where-length-vs-count
`matches` is :-1: 
+1 for jashkenas/underscore#1985

akre54:
Unless we decide to add a `mixin` function like @megawac described, this type of thing should live in your Base Collection class. It's definitely not a common use case.

PR #3487, opened 2015-02-10
model::equal()

redexp:
Sometimes it's useful to compare model with hash for example in some custom filters like
```javascript
// where (a = 1 and b = 1) or (a = 2 and b = 2)
collection.filter(function (model) {
  return model.equal({a: 1, b: 1}) || model.equal({a: 2, b: 2});
});
```
or when you need to set to attribute like
```javascript
showModelsByQuery: function (query) {
  this.forEach(function (model) {
    model.set('hidden', !model.equal({a: 1, b: 1}));
  });
})
```
You can't do such thing with `.where()` only if you will loop over models twice

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -440,6 +440,14 @@
       return _.clone(this._previousAttributes);
     },
 
+    // Returns `true` if all attributes is equal to model attributes
+    equal: function (attrs) {
+      for (var key in attrs) {
+        if (attrs[key] !== this.get(key)) return false;
+      }
+      return true;
+    },
+
     // Fetch the model from the server. If the server's representation of the
     // model differs from its current attributes, they will be overridden,
     // triggering a `"change"` event.
@@ -835,10 +843,7 @@
     where: function(attrs, first) {
       if (_.isEmpty(attrs)) return first ? void 0 : [];
       return this[first ? 'find' : 'filter'](function(model) {
-        for (var key in attrs) {
-          if (attrs[key] !== model.get(key)) return false;
-        }
-        return true;
+        return model.equal(attrs);
       });
     },
 
--- a/test/model.js
+++ b/test/model.js
@@ -351,6 +351,15 @@
     equal(model.get('name'), undefined);
   });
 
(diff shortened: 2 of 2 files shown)


jridgewell:
I think https://github.com/jashkenas/backbone/pull/3307 covers this.

redexp:
@jridgewell yes, but my version is just 3 lines longer but much less memory usage. I think method name `matches` is better than `equal`

backbone.js
```diff
@@ -835,10 +843,7 @@
     where: function(attrs, first) {
       if (_.isEmpty(attrs)) return first ? void 0 : [];
       return this[first ? 'find' : 'filter'](function(model) {
-        for (var key in attrs) {
```
akre54: It looks like your code here is out of date, try pulling down the latest.

redexp:
Looks like it useless if you prefer `.matches()` but imho this method uses more memory and not helps to make `collection#where` method shorter

akre54:
This use case is already covered by `matches` as @jridgewell pointed out. If we find that it's too slow we'll fix any perf up there.

redexp:
@akre54 `equal` 5 times faster
http://jsperf.com/backbone-mathces-vs-equal

jridgewell:
Like [I said](https://github.com/jashkenas/backbone/pull/3307#issuecomment-73952459), this can be accomplished (with performance boosts) if we move https://github.com/jashkenas/underscore/issues/1985 forward.

backbone.js
```diff
@@ -440,6 +440,14 @@
       return _.clone(this._previousAttributes);
     },
 
+    // Returns `true` if all attributes is equal to model attributes
+    equal: function (attrs) {
+      for (var key in attrs) {
+        if (attrs[key] !== this.get(key)) return false;
```
megawac: This isn't sufficiently equal anyway, its possible that `this.attributes` has more attributes in it still

redexp: agree, `matches` is better name for this method

PR #3488, opened 2015-02-11
Include updated options in sync callbacks

### LINKED ISSUE #3478: Collection#create doesn't properly pass final options to success callback

platinumazure:
Better version of #3484. Fixes #3478.

### DIFF
--- a/test/model.js
+++ b/test/model.js
@@ -565,10 +565,48 @@
     equal(this.ajaxSettings.url, '/collection/42');
   });
 
+  test("save will pass extra options to success callback", 1, function () {
+    var SpecialSyncModel = Backbone.Model.extend({
+      sync: function (method, model, options) {
+        _.extend(options, { specialSync: true });
+        return Backbone.Model.prototype.sync.call(this, method, model, options);
+      },
+      urlRoot: '/test'
+    });
+
+    var model = new SpecialSyncModel();
+
+    model.save(null, { success: onSuccess });
+    this.ajaxSettings.success();
+
+    function onSuccess(model, response, options) {
+      ok(options.specialSync, "Options were passed correctly to callback");
+    }
+  });
+
   test("fetch", 2, function() {
     doc.fetch();
     equal(this.syncArgs.method, 'read');
     ok(_.isEqual(this.syncArgs.model, doc));
+  });
+
+  test("fetch will pass extra options to success callback", 1, function () {
+    var SpecialSyncModel = Backbone.Model.extend({
+      sync: function (method, model, options) {
+        _.extend(options, { specialSync: true });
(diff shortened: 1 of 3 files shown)


test/model.js
```diff
@@ -580,6 +618,25 @@
...
+      },
+      urlRoot: '/test'
+    });
+
+    var model = new SpecialSyncModel({ id: 'id' });
+
+    model.destroy({ success: onSuccess });
+    this.ajaxSettings.success();
+
+    function onSuccess(model, response, options) {
```
akre54: No named functions. Use `onSuccess = function() { ... }` or put this inline.

platinumazure: Not even in a test?

This is an asynchronous bit, but in order for the test to read more naturally, I always like to put assertions after actions. This technique takes advantage of function hoisting to allow this to work. `var onSuccess = function() { /* ... */ }` will not work-- the declaration is hoisted, but the assignment doesn't happen until the line is reached, so I would have to move that function up. I would never do this in actual code.

That said, if there's no flexibility forthcoming on this, I can certainly make some tweaks. I just wanted to explain my rationale first and make sure it was understood I didn't do it randomly and/or without knowing the potential dangers.

akre54: > Not even in a test?

Nah, we want the source code and its tests to have a consistent style.

> This technique takes advantage of function hoisting to allow this to work

Right. Just move the `onSuccess` above the `destroy` call or inline it in the `options` object. Function hoisting is a bad part of JS and serves no real point here.

platinumazure: As I tried to explain, the "real point" is readability in the test: arrange, act, assert, from top to bottom. But I will push commits to fix this.

platinumazure:
@akre54 Better?

test/model.js
akre54: No I understand what you're saying, I just disagree it reads easier ;)

Thanks for fixing that up.

akre54:
Much. Thanks.

Issue #2920, opened 2013-12-13
Collisions between id and cid within collection cause mayhem


TrevorBurnham:
Because both `id` and `cid` are used as the keys of `_byId`, there's an implicit assumption that no model's `id` will equal the `cid` of any other model in the same collection. Breaking this assumption causes very strange behavior:

```
var model1 = new Backbone.Model()  // model1.cid = 'c1'
var model2 = new Backbone.Model({id: 'c1'})
var collection1 = new Backbone.Collection([model1, model2])
var collection2 = new Backbone.Collection([model2, model1])
```

In this case, `collection1` will only contain `model1`, and `collection2` will only contain `model2`.

This doesn't seem like an unrealistic edge case to me. I think the `_byId` hash should be broken up into separate `_byId` and `_byCid` hashes, with the `Collection::get` behavior made explicit to favor ids over cids:

```
return this._byId[obj] || this._byCid[obj] || this._byId[obj.id] || this._byCid[obj.cid];
```

PR #3490, opened 2015-02-11
Customise model cid prefix

yousefcisco:
This is a really simple change that will allow users to customise the cid prefix to prevent clashes with backend model ids and resolve issues like https://github.com/jashkenas/backbone/pull/3356 and https://github.com/jashkenas/backbone/issues/2920.

I know that this is an edge case but given the small amount of code change required I hope it won't be a problem.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -338,7 +338,7 @@
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
-    this.cid = _.uniqueId('c');
+    this.cid = _.uniqueId(this.cidPrefix);
     this.attributes = {};
     if (options.collection) this.collection = options.collection;
     if (options.parse) attrs = this.parse(attrs, options) || {};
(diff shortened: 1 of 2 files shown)


akre54:
How about putting it on the prototype instead? 

```diff
diff --git a/backbone.js b/backbone.js
index 7a11fdf..d4f1c8d 100644
--- a/backbone.js
+++ b/backbone.js
@@ -338,7 +338,7 @@
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
-    this.cid = _.uniqueId('c');
+    this.cid = _.uniqueId(this.cidPrefix);
     this.attributes = {};
     if (options.collection) this.collection = options.collection;
     if (options.parse) attrs = this.parse(attrs, options) || {};
@@ -361,6 +361,10 @@
     // CouchDB users may want to set this to `"_id"`.
     idAttribute: 'id',
 
+    // The prefix used to identify models locally. You may want to
+    // override this if you're experiencing name clashes.
+    cidPrefix: 'c',
+
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
     initialize: function(){},
```

yousefcisco:
I ran into this issue the other day when we noticed that some products were missing on a client's product listings page. The page integrated directly with a product merchandising service that used a character at the beginning of ids to distinguish between different product types. Products with 'c' at the beginning were going missing when the collection attempted to update the model with a cid that matches the id instead of adding the new model.

I'll update the PR today so the cidPrefix is on the prototype. Thanks for the feedback!

akre54:
Can you squash down to a single commit and I'll merge? Thanks.

yousefcisco:
Done. Thank you

PR #3491, opened 2015-02-12
Fix the collection.prototype.modelId method to don't crash when the model property is undefined

gartz:
Create generic collections where has no model property defined, should not crash on the method `modelId`, it just should search for the default `idAttribute`, and keep working.

Tests included.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1006,7 +1006,7 @@
 
     // Define how to uniquely identify models in the collection.
     modelId: function (attrs) {
-      return attrs[this.model.prototype.idAttribute || 'id'];
+      return attrs[this.model && this.model.prototype.idAttribute || 'id'];
     },
 
     // Private method to reset all internal state. Called when the collection
--- a/test/collection.js
+++ b/test/collection.js
@@ -1414,6 +1414,13 @@
     equal(StoogeCollection.prototype.modelId({_id: 1}), 1);
   });
 
+  test("modelId with no model in the collection", function() {
+    var StoogeCollection = Backbone.Collection.extend({model: undefined});
+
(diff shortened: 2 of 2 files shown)


akre54:
Sorry -- what's the use case for this? You almost always want `model` to be defined. If you're doing something funky, just override `modelId` yourself.

gartz:
Old versions this doesn't broken, so it's breaking on Backbone updating. For huge project just if we patch the backbone lib all the time, that is bad.

akre54:
To quote [from the docs](http://backbonejs.org/#FAQ-extending):
> Many JavaScript libraries are meant to be insular and self-enclosed, where you interact with them by calling their public API, but never peek inside at the guts. Backbone.js is not that kind of library. ... If you catch yourself adding methods to Backbone.Model.prototype, or creating your own base subclass, don't worry — that's how things are supposed to work.

It's a good idea to have your own "base" model and collection classes just for this reason. It becomes a simple and straightforward change once you do.

jashkenas:
Isn't a `model` property defined by default?

gartz:
it should, but if `model` is required by a collection and must be a `Model` constructor, it should be test and throw a error that drive the developer to the source of the problem.
This bug just happens because a requirement wasn't filled and there isn't a check of feedback for the requirement. 

I agree with the quote from @akre54  and I still think this shouldn't generate the current exception in the `modelId` method.

akre54:
But it works correctly out of the box. If *you* do something nonstandard like override model to be `undefined` in your app then it shouldn't be a surprise that this breaks some things. `modelId` is exposed specifically to allow you to override it in cases like this.

Issue #2365, opened 2013-03-12
There is no 'change' event on 'collections'


nitinhayaran:
I have this case, where I think there should be some sort of 'change' event on 'collections' as well.

I presently have a collection of photos which are shown on a page. When the user navigate to other page, this collection is cached. Now if the user comes back to previous page, it is rendered from cached collection, In addition to that a fetch request fired to check if any new image is uploaded during this period. 

Now here, its really difficult to figure out if there is any changes (addition/deletion) occurred in collection. I can somehow work by putting checks in 'add' / 'remove' listener, but it'll be complex. Depending upon the data, there would be multiple 'add'/'remove' function calls, and it won't be very effective to check 'collection change' in these functions. 

I think there should be a simpler way to achieve this.

Issue #2268, opened 2013-02-14
Collection.update leads to a large number of `add` and `remove` events


ivanzhaowy:
Here is the thing:
var collection = new Backbone.Collection({...});

collection.on('add', function() {
    // Operate DOMs
});

collection.on('remove', function() {
    // Operate DOMs
});

collection.update({...});

When the `update` function trigger a large number of events will lead to a performance problem. Is there any way too solve this?

Issue #2265, opened 2013-02-13
added array in the add event callback for collection


wanderer:
When adding multiple items to a collection it would be helpful have the array of models that were added in the `add` callback. Maybe in options? The following is how it might work

``` javascript
someRandomCollection.on("add",callback);

function  callback(model, collect, options) {
        //last model added do something
       if(model === _.last(options.added)){
           dosomething()
       }
}
```

fastest963:
Not to revive a dying thread.. but I just opened #3492 which relates to this. The problem with using `debounce` is that all the events might not have been added at the end during your debounce time. We're doing real-time syncing of collections between tabs and we'd have to do a debounce while also keeping track of all the indexes which can get difficult. The changes I proposed are very small and hopefully resolve this issue and the many others raised on this topic.

PR #3492, opened 2015-02-12
Added `adds` and `removes` events

fastest963:
I'm adding the `adds` and `removes` events which fire an aggregated event with all the added/removed models. This is useful for batch changes that need to be listened for. Without this you'd have to listen to `add` and use debounce to collect all the changes within a certain period of time. This is difficult since you could collect changes with different `at`'s specified and you'll need to keep track of those.

I did change what `remove()` returns. It will only return the models that were actually removed, which could be undefined, if you passed in a singular model that wasn't in the collection. I added a test to back up this new behavior since I didn't see any tests backing up the current way it was working. The docs also weren't clear which way it was supposed to work but since `removes` will only give you the models actually removed, I thought `remove()` should work the same way.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -744,7 +744,10 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       options || (options = {});
-      for (var i = 0, length = models.length; i < length; i++) {
+      // In order to fire removes we're going to rewrite models
+      // as we go and j is going to keep our position. If a model
+      // is invalid and not actually removed, it won't be written.
+      for (var i = 0, length = models.length, j = 0; i < length; i++) {
         var model = models[i] = this.get(models[i]);
         if (!model) continue;
         var id = this.modelId(model.attributes);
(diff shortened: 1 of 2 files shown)


akre54:
For reference: #2063 #2365 #2268 #2265 and others. What about reordering? What's so difficult about reconciling what models you've already rendered and just iterating over the list?

fastest963:
So I talked with @akre54 on gitter and stated that I'd like the `adds`/`removes` so we can get batched change events for efficiently rendering large collections (like your queue or library). Instead of getting 50 `add` events and hitting the DOM for each one, you could get one `adds` event and render all 50 (songs in our case) at once.
The other reason we'd like it is for real-time queue syncing. We can listen for `adds`,`removes`,`reset` and forward those events over the wire instead of having to forward potentially tons of the single events.

In response to the question above, we *could* just reconcile after every event (if I'm understanding you correctly) but some people have thousands of songs in their library/collection and always doing that loop (even when its not necessary) seems kinda crappy. I know looping isn't very expensive but if we can do it just based on change events I'd rather do that.

I also wouldn't be asking for this if the change to backbone was large or complicated. The change is relatively small (9 lines including a bunch of comments) but saves us a bunch of code on our end.

fastest963:
Talked to @akre54 on gitter and we decided that not enough people would benefit from this for it to be worth adding to core. I'm stripping out the `coll.remove` changes though and making a new PR with just those.

PR #3493, opened 2015-02-13
request:[method] events.

CoryDanielson:
- `request:create`, `request:read`, `request:update`, `request:patch`, `request:delete`
- These events are triggered before the `request` event, similar to `change:[attribute]` events.

### DIFF
--- a/index.html
+++ b/index.html
@@ -863,6 +863,7 @@ view.stopListening(model);
       <li><b>"change:[attribute]"</b> (model, value, options) &mdash; when a specific attribute has been updated. </li>
       <li><b>"destroy"</b> (model, collection, options) &mdash; when a model is <a href="#Model-destroy">destroyed</a>. </li>
       <li><b>"request"</b> (model_or_collection, xhr, options) &mdash; when a model or collection has started a request to the server. </li>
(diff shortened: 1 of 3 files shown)


index.html
```diff
@@ -2421,7 +2422,7 @@ <h2 id="Sync">Backbone.sync</h2>
     </p>
 
     <ul>
-      <li><b>method</b> – the CRUD method (<tt>"create"</tt>, <tt>"read"</tt>, <tt>"update"</tt>, or <tt>"delete"</tt>)</li>
+      <li><b>method</b> – the CRUD method (<tt>"create"</tt>, <tt>"read"</tt>, <tt>"update"</tt>, <tt>"patch"</tt> or <tt>"delete"</tt>)</li>
```
CoryDanielson: Updated the docs here to include that `patch` can also be passed into `Backbone.Sync` under the `method` parameter.

CoryDanielson:
Motivation behind this is to provide more granular convenience events that can be used in order to display information to the user in accordance with the request being made.

I've often found myself needing/wishing that I could easily differentiate between different request types, but haven't found an elegant way to do so without adding a thick layer of cruft around sync calls.

akre54:
#2742 and others.

PR #3495, opened 2015-02-16
Add documentation about collection.remove `models` arg.

### LINKED ISSUE #3494: Acceptable values for `models` arg of collection.remove()
The current documentation implies that only model objects are acceptable, but under the hood it uses `collection.get()`, which will take a model object, `id`, or `cid`. Is it supposed to be restricted to just model objects or can it be documented as accepting the same values as `get()`?

jmm:
Fix #3494.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1777,7 +1777,10 @@ ships.add([
     <p id="Collection-remove">
       <b class="header">remove</b><code>collection.remove(models, [options])</code>
       <br />
-      Remove a model (or an array of models) from the collection, and returns them.
+      Remove a model (or an array of models) from the collection, and returns
+      them. <tt>models</tt> (or each of its elements if an array) can be any
+      value acceptable as the <tt>id</tt> argument of
+      <a href="#Collection-get"><tt>collection.get</tt></a>.
       Fires a <tt>"remove"</tt> event, which you can use <tt>silent</tt> to suppress.
       The model's index before removal is available to listeners as
       <tt>options.index</tt>.


jmm:
Is some specific line-length limit supposed to be maintained in this file?

akre54:
Usually about 80 chars. No hard limit though.

index.html
```diff
@@ -1777,7 +1777,10 @@ <h2 id="Collection">Backbone.Collection</h2>
     <p id="Collection-remove">
       <b class="header">remove</b><code>collection.remove(models, [options])</code>
       <br />
-      Remove a model (or an array of models) from the collection, and returns them.
+      Remove a model (or an array of models) from the collection, and returns
```
akre54: "Remove" doesn't agree with "returns" here (not your fault but let's fix this to either "Removes... returns them" or "Remove ... return them").

index.html
```diff
@@ -1777,7 +1777,10 @@ <h2 id="Collection">Backbone.Collection</h2>
...
+      them. <tt>models</tt> (or each of its elements if an array) can be any
```
akre54: How about `The model or models passed to <b>remove</b> can be either Models, ids, or plain objects, anything that <tt>get</tt> can use.` Or something.

index.html
jmm: Agreed, I noticed that too but left it alone.

akre54: Feel free to fix here. Thanks

index.html
jmm: I think it's preferable to keep this DRY and refer to the `id` arg of `get()` for the acceptable values instead of duplicating them here. Your call.

I think it's better to be explicit that `models` can be a a single item or an array, so I feel that replacing "a model (or an array of models)" with "The model or models" reduces clarity.

How about?:
```
Remove model(s) and return a model, or array of models (if `models` is an
array). `models` can be any value acceptable as the <tt>id</tt> argument of
<a href="#Collection-get"><tt>collection.get</tt></a>, or an array of those values.
```

jmm:
> Usually about 80 chars. No hard limit though.

Ok, thanks.

index.html
akre54: Personally I feel it's a little more confusing to refer to `get`. I have a slight pref to enumerate the options here.

Conversely, I'm not sure you need to spell out the "if models is an array" part twice. The plural kind of implies that.

jmm: > Personally I feel it's a little more confusing to refer to `get`. I have a slight pref to enumerate the options here.

It would be a link, the most basic building block of the web. People that are confused by that are going to have bigger problems using Backbone. Namely they're going to be confused about browsing the web let alone building it. If it's duplicated here that means that if acceptable values for `get()`'s `id` change someone will need to remember they're duplicated here and actually update them. Are you sure you don't want to keep this simple and just link to the `get()` documentation?

> Conversely, I'm not sure you need to spell out the "if models is an array" part twice. The plural kind of implies that.

It's better to be specific. For example to avoid any surprise here:

```js
// Returns model
collection.remove(model);

// Returns array
collection.remove([model]);
```

akre54: Sorry -- that should've said it'd be confusing referring *just* to get. How about enumerating the options but also adding in the link?

akre54:
Merging. I'll fixup.

PR #3497, opened 2015-02-17
Remove the micro-opt of caching array length

akre54:
[It's not actually effective](http://mrale.ph/blog/2014/12/24/array-length-caching.html) and it makes the code uglier. 

JSPerf is down at the moment but I seem to remember the benefits in older browsers not even being that great. @jdalton, what say you?

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -90,12 +90,12 @@
     var i = 0, names, length;
     if (name && typeof name === 'object') {
       // Handle event maps.
-      for (names = _.keys(name), length = names.length; i < length; i++) {
+      for (names = _.keys(name); i < names.length; i++) {
         memo = iteratee(memo, names[i], name[names[i]], context, ctx);
       }
     } else if (name && eventSplitter.test(name)) {
       // Handle space separated event names.
-      for (names = name.split(eventSplitter), length = names.length; i < length; i++) {
+      for (names = name.split(eventSplitter); i < names.length; i++) {
         memo = iteratee(memo, names[i], callback, context, ctx);
       }
     } else {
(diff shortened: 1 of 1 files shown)


megawac:
The only case I can see this making a difference is a custom collection which creates a custom object to dynamically compute the `length`. Shitty example

```
function createCollection(items) {
   var result = {};
   var size = items.size;
   items.forEach((item, index) => result[index] = item + 5);
   result.length= {
     valueOf() {
         return size;
     }
   }
}
```

jdalton:
Modern engines will handle the `length` hoisting, this is optimized in IE in Windows 10 preview too. 
Caching a length is handy when you don't want array mutations to affect the iteration count.

akre54:
I think in each of these cases you really shouldn't be mutating the array objects in the first place, right?

jdalton:
> I think in each of these cases you really shouldn't be mutating the array objects in the first place, right?

Nothing appears to be able to access the arrays to mutate them while it's iterating so it should be ok.

jashkenas:
I'm sure that this has gone back-and-forth several times before already ... but whatever ;)

akre54:
@jdalton is this worth [changing in CoffeeScript](http://coffeescript.org/#try:eat%20food%20for%20food%20in%20foods) too?

megawac:
I wouldn't change it there it could break things if the user changes length
i n the loop

akre54:
You probably shouldn't regardless. But fair 'nuff.

jashkenas:
It ain't worth changing in CoffeeScript — there it's a nice micro-optimization that you get for free. The programmer never has to deal with or worry about it.

PR #3499, opened 2015-02-17
Remvoe trigger with object's value prepending

jridgewell:
https://github.com/jashkenas/backbone/issues/3460#issuecomment-74762412

Backbone still supports triggering with an object, but the args passed to `#trigger` will not be prepended with the object's value.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -279,22 +279,13 @@
     if (!this._events) return this;
     var args = slice.call(arguments, 1);
 
-    // Pass `triggerSentinel` as "callback" param. If `name` is an object,
-    // it `triggerApi` will be passed the property's value instead.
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -279,22 +279,13 @@
     if (!this._events) return this;
     var args = slice.call(arguments, 1);
 
-    // Pass `triggerSentinel` as "callback" param. If `name` is an object,
-    // it `triggerApi` will be passed the property's value instead.
-    eventsApi(triggerApi, this, name, triggerSentinel, args);
+    eventsApi(triggerApi, this, name, void 0, args);
```
akre54: Excellent.

jridgewell: :grin: That was way to complicated to explain.

PR #3501, opened 2015-02-18
Update index.html

rjatkinson2:
Would be very helpful to know that the callback function has access to the object.

image


akre54:
I'm not sure that's true.

```js
obj.on('evt', function() {
  console.log(arguments);
});

obj.trigger('evt');
// logs `[]`
```

The object will be passed as the context of the calling function...

```js
obj.on('evt', function() {
  console.log(this === obj); // true
});
```

... provided you don't pass your own `ctx` argument:

```js
obj1.on('evt', function() {
  console.log(this === obj1); // false
  console.log(this === obj2); // true
}, obj2);

obj1.trigger('evt');
```

PR #3505, opened 2015-02-19
Run test suite using karma

megawac:
Me and @akre54 were discussing this yesterday on this commit https://github.com/megawac/backbone/commit/bea414e6af6e1f3c303d7ed2e8a4055d2457dedc

This commit makes travis test using `Phantom`, `Chrome` and `FF`. Later we can do sauce as well

A later commit can omit `phantomjs` from the dev dependencies (which would be sweet!) and assume the user has firefox installed. This assumption would allow people to omit the 13MB buggy install every time and if they dont have ff they can use chrome or opera or whatever else they want no probs.

We did this a couple days ago on underscore as well https://github.com/jashkenas/underscore/pull/2054


akre54:
Chrome failed on Travis. Any ideas?

Also I removed the test-amd.html file since it's pretty superfluous. If you pull in master you should be up to date.

megawac:
@akre54 just https://github.com/karma-runner/karma/issues/1144 had to run in sandbox apparently

braddunbar:
I'm all for using a real browser for testing instead of phantom. :+1:

We still need to swap out the `npm test` script and add `karma-cli` to `package.json` right?

Florian-R:
> We still need to swap out the npm test script and add karma-cli to package.json right?

No need to add `karma-cli` as a dependencie, 
```json
"scripts": {
   "test": "./node_modules/karma/bin/karma start --single-run"
}
```
in the package.json does the trick.

akre54:
Awesome. Let's get to fixing #3434 now.

PR #3506, opened 2015-02-19
Fragment Encoding

jridgewell:
This is another take on #3434. Instead of trying to decode the fragments, which leads to weird double decoding issues, we can just maintain them in the encoded state.

How can we do this? Because it's easy to determine if the fragment has been decoded (either from a browser bug, or from the dev who never encoded it) since we can just check for a decoded character. If we find one, we can assume that the fragment has been run through `decodeURI()` and just invert it with `encodeURI()`. Now we can match encoded fragments with encoded routes, and everything works perfectly. Static parts of the route match, because the route is encoded _and_ the fragment is. And params / splats will be mapped through `decodeURIComponent()`, guaranteeing it'll never throw (cause it's encoded!).

However, we can't do the opposite. Say we check to see if there are any decoded characters. If there are, then we treat it as already decoded, if not, we `decodeURI()`. The problem is, we don't know if that will leave the fragment in a state that will throw when we later `decodeURIComponent()`. Thus, we're back at #3434.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1499,7 +1499,7 @@
 
   // Cached regex for detecting invalid percent-encoded octets. Will match any
   // "%" that is not followed by a hexadecimal number.
-  var invalidPercentOctets = /%(?![a-f0-9]{2}|[A-F0-9]{2})/;
+  var invalidPercentOctets = /%(?![a-fA-F0-9]{2})/;
 
   // Detect if fragment needs to be URI encoded.
   var shouldEncode = function(fragment) {
--- a/backbone.js
+++ b/backbone.js
@@ -1442,7 +1442,7 @@
     // Convert a route string into a regular expression, suitable for matching
     // against the current location hash.
     _routeToRegExp: function(route) {
-      route = route.replace(escapeRegExp, '\\$&')
+      route = encodeFragment(route).replace(escapeRegExp, '\\$&')
                    .replace(optionalParam, '(?:$1)?')
                    .replace(namedParam, function(match, optional) {
                      return optional ? match : '([^/?]+)';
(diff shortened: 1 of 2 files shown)


jashkenas:
Interesting. @braddunbar — any opinion?

backbone.js
```diff
@@ -1493,6 +1493,25 @@
   // Cached regex for stripping urls of hash.
   var pathStripper = /#.*$/;
 
+  // Cached regex for detecting any character that is invalid in an encoded URI.
+  // https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI
+  var invalidURIEncoded = /[^;,/?:@&=+$a-zA-Z0-9-_.!~*'()#%]/;
+
+  // Cached regex for detecting invalid percent-encoded octets. Will match any
+  // "%" that is not followed by a hexadecimal number.
+  var invalidPercentOctets = /%(?![a-f0-9]{2}|[A-F0-9]{2})/;
```
jridgewell: I'd like to get your thoughts on this. While `%Ab` is technically a valid percent-encoding, I imagine a respectable browser always using the same casing. In particular, requiring the same case could catch "encoded" strings like `1.5%Fat` (cause I can't think of a good example...).

braddunbar:
This *is* interesting. Thanks @jridgewell! Here's a couple of initial thoughts.

This is a big breaking change for anyone who depends on `Backbone.history.fragment`. I know that I have in the past and I suspect others do as well. The fragment used to be kept encoded and we swapped because it made matching much easier, among other things @jashkenas may remember.

Also, I don't see any added benefit over #3434. Decoding everything but `%25`, while maybe not ideal, *does* fix #3426 with minimal API change and impact on functionality. I know it breaks PhantomJS but I think a move to testing with real browsers has been met with unanimous approval.

As for whether this is a better approach in general, I'd need to know what benefits we're shooting for and why. That's my 2¢! :smile_cat:

backbone.js
braddunbar: If it's a valid percent-encoding someone will eventually use it and ask us why we don't accept it. :smile:

jridgewell:
> This is a big breaking change for anyone who depends on Backbone.history.fragment.

I suppose we can keep `history.fragment` decoded. But it's going to be ugly to maintain both an encoded private fragment, and a decoded public one.

These undocumented, public history properties are [killing](https://github.com/jashkenas/backbone/pull/3443#issuecomment-74770349) my refactors. :stuck_out_tongue:

> Also, I don't see any added benefit over #3434.

I see #3434 as a ugly hack.

jashkenas:
Backbone.history.fragment wasn't ever part of the real public API — so if we really think this is an improvement, I think it's okay to change it.

braddunbar:
> I see #3434 as a ugly hack.

Why is that? It decodes everything that `decodeURI` would encode except for `%25`. The functionality and motivation is explained in comments and there's good test coverage. What's hacky about it, objectively?

> so if we really think this is an improvement, I think it's okay to change it.

This is my point. I don't understand what this refactor buys us, in terms of user facing API or reduction of complexity.

jridgewell:
Actually, #3434 does add one thing this can't: `myyjä/foo%20%25%3F%2f%40%25%20bar`. If you want to use a encoded component, the entire fragment must be encoded. I still think it's a hack, but it works...

PR #3511, opened 2015-02-20
Test IE7 & IE8

megawac:
IE6 is still creating issues. Not sure what was wrong with these 2 browsers earlier

### DIFF
--- a/karma.conf-sauce.js
+++ b/karma.conf-sauce.js
@@ -15,9 +15,8 @@ var sauceBrowsers = _.reduce([
   ['internet explorer', '11', 'Windows 8.1'],
   ['internet explorer', '10', 'Windows 8'],
   ['internet explorer', '9', 'Windows 7'],
-  // Currently do not work with Karma.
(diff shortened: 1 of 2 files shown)


karma.conf-sauce.js
```diff
@@ -15,9 +15,8 @@ var sauceBrowsers = _.reduce([
   ['internet explorer', '11', 'Windows 8.1'],
   ['internet explorer', '10', 'Windows 8'],
   ['internet explorer', '9', 'Windows 7'],
-  // Currently do not work with Karma.
-  // ['internet explorer', '8', 'Windows 7'],
-  // ['internet explorer', '7', 'Windows XP'],
+  ['internet explorer', '8'],
```
jridgewell: This doesn't need a OS?

jridgewell:
:clap:

karma.conf-sauce.js
megawac: Nah, none of them do actually

PR #3512, opened 2015-02-21
coll.remove only returns models that were removed

fastest963:
This is taken from #3492.

`coll.remove` will only return the models that were actually removed, which could be undefined, if you passed in a singular model that wasn't in the collection. I added a test to back up this new behavior since I didn't see any tests backing up the current way it was working. The docs also weren't clear which way it was supposed to work but if a model wasn't actually removed it makes sense not to return it.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -763,7 +763,10 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       options || (options = {});
-      for (var i = 0; i < models.length; i++) {
+      // We're going to rewrite models as we go and j is going to keep our
+      // position. If a model is invalid and not actually removed, it won't
+      // be rewritten.
+      for (var i = 0, j = 0; i < models.length; i++) {
         var model = models[i] = this.get(models[i]);
         if (!model) continue;
         var id = this.modelId(model.attributes);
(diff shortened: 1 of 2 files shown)


test/collection.js
```diff
@@ -98,7 +98,7 @@
   test('get with "undefined" id', function() {
     var collection = new Backbone.Collection([{id: 1}, {id: 'undefined'}]);
     equal(collection.get(1).id, 1);
-  }),
```
akre54: Good catch. 74fdd620d75db9737a21f7a3d8d2f566f447a482

backbone.js
```diff
@@ -750,8 +753,12 @@
           options.index = index;
           model.trigger('remove', model, this, options);
         }
+        models[j++] = model;
         this._removeReference(model, options);
       }
+      // We only need to slice if models array should be smaller, which is
+      // caused by some models not actually getting removed.
+      if (models.length !== j) models = models.slice(0, j);
       return singular ? models[0] : models;
```
akre54: Won't this break when you pass in, say, 1 model but it removes more than one?

fastest963: Hmm I'm not sure. The loop above is looping over `models`, which would only have 1 thing so I don't see how it could remove more than 1 model. Can you clarify? I might just be missing something. Plus, you can't have the same model in the collection more than once, right?

fastest963:
Any update on this? Thanks!

fastest963:
https://github.com/jashkenas/backbone/commit/bc73e25c5e710a4edade2b5827957dee55cc9a56 is almost 100% my commit, would've been nice just to have merged this instead of copying/pasting and commiting yourself :P

Issue #3345, opened 2014-10-17
Custom comparators?


cbroome:
Backbone 'set' relies on the underscore isEqual method to determine whether attributes have changed, however this becomes an issue when an attribute contains another model or collection.  I'd really like for there to be a way to specify a custom comparator on an attribute that could be consulted instead of _.isEqual if present. Overriding 'set' gets to be a pain.

MidnightWonderer:
@cbroome Why you need to revert isEqual back after you override it?
In megawac's solution you can put as many condition check as you like to make sure there is no conflict not just instanceof check.

cbroome:
@MidnightWonderer Let's say I trust underscore's isEqual to be safer than anything my fellow coders or myself would cook up. We use isEqual in other portions of the app, and the idea of running through complex conditional logic every time is not appealing. It strikes me as bad practice. 

But to bring this back to my original point: I really think it would be beneficial for objects to define their own equality checks.

akre54:
@cbroome let's test this out in a PR. Want to whip one up?

cbroome:
sorry, PR? I'm unfamiliar with the acronym.

akre54:
"PR" == "pull request" on Github.

cbroome:
Ahhh, duh. Yeah, I'll give it a go in the next couple days.

akre54:
Thanks :)

cbroome:
Pull Request: https://github.com/jashkenas/backbone/pull/3515

PR #3515, opened 2015-02-24
#3345 - move attribute comparisons into their own method on the model

cbroome:
This should allow models to provide custom logic to compare atributes. For a set, each attribute will still be compared to every other attribute between the two models. I'm not going to try to change that behavior. This should be enough wiggle room to allow for more powerful comparisons.

I don't know about naming conventions but `attributesAreEqual` does not appear in any Backbone projects on Github. I'm not wedded to the name.

Also, I wasn't able to get the tests to work on Firefox without moving jQuery to the top of the stack.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -463,8 +463,8 @@
       // For each `set` attribute, update or delete the current value.
       for (attr in attrs) {
         val = attrs[attr];
-        if (!_.isEqual(current[attr], val)) changes.push(attr);
-        if (!_.isEqual(prev[attr], val)) {
+        if (!this.attributesAreEqual(current[attr], val)) changes.push(attr);
+        if (!this.attributesAreEqual(prev[attr], val)) {
           this.changed[attr] = val;
         } else {
           delete this.changed[attr];
@@ -493,6 +493,12 @@
       this._pending = false;
       this._changing = false;
       return this;
+    },
+    
+    // Compare two attributes, return true if they are found to be equal.
+    // (3345)
+    attributesAreEqual: function( attrA, attrB ) {
+        return _.isEqual( attrA, attrB );
     },
 
     // Remove an attribute from the model, firing `"change"`. `unset` is a noop
@@ -526,7 +532,7 @@
       var val, changed = false;
       var old = this._changing ? this._previousAttributes : this.attributes;
       for (var attr in diff) {
-        if (_.isEqual(old[attr], (val = diff[attr]))) continue;
+        if (this.attributesAreEqual(old[attr], (val = diff[attr]))) continue;
         (changed || (changed = {}))[attr] = val;
       }
       return changed;
--- a/backbone.js
+++ b/backbone.js
@@ -463,8 +463,8 @@
       // For each `set` attribute, update or delete the current value.
       for (attr in attrs) {
         val = attrs[attr];
-        if (!this.attributesAreEqual(current[attr], val)) changes.push(attr);
-        if (!this.attributesAreEqual(prev[attr], val)) {
+        if (!this.attributesEqual(current[attr], val, attr)) changes.push(attr);
+        if (!this.attributesEqual(prev[attr], val, attr)) {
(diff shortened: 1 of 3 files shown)


akre54:
Let's add in a few more tests for the expected behavior of this method (how are `Model#set` and `Model#changedAttributes` affected once you've overwritten them? Show a good example of the use case of this change.)

How about `isEqual`, `equals`, or `attributesEqual` instead? The name is a bit verbose.

cbroome:
I like "attributesEqual". I'll update in the next few days with some more tests.

cbroome:
Updated with more verbose unit tests. Also looking at the code again I think it'd be helpful to also pass the name of the attribute. I added it as the last parameter for attributesEqual as it could be an 'optional' parameter, but it could be moved to the first too.

jashkenas:
How about a real use case?

cbroome:
I'm sorry, you mean write a different (better) test or to restate the issue driving the request?

jashkenas:
> This should be enough wiggle room to allow for more powerful comparisons.

More powerful comparisons like what? When and how would this be useful? Why is doing it in this way a better idea than overriding _.isEqual in your project?

cbroome:
This would aid in comparisons of objects where some data may be extraneous and should be skipped, or where "equality" between objects may need some complex logic. This seems to be a pretty common case in applications. You're not the first to suggest overwriting _.isEqual. In the ticket leading to this pull request and on StackOverflow and the like, it's cited as common practice. At my place of business, our Backbone application currently does this as well. We catch 'set' operations at a global level and swap out isEqual with a custom method.

I'm not in favor of the approach for a number of reasons. First and foremost, I don't like overwriting good existing code in a separate namespace. Doing so in this manner, however brief, means that the _.isEqual is extended for the whole application (at least with how we use RequireJS). If a dev forgets to restore it back to its original state it could lead to side effects elsewhere.

I like the idea of the model being able to define how its attributes are compared. Any exception to the standard isEqual methodology would probably be very specific to the model itself, most would probably utilize the default behavior. This could allow the few models that need to override isEqual behavior the means to do so. Of course getting such a method into the prototype chain would mean that child classes could build on their parent's attribute comparison logic.

jashkenas:
> We catch 'set' operations at a global level and swap out isEqual with a custom method.

Can you show us what that custom method looks like?

cbroome:
Can't show the exact code, of  course, but this is a close approximation. 

The overriding: 
```javascript
  custom.namespace.BackboneOverride.set = function( key, val, options ) {
    var rv;

    _.isEqual = custom.namespace.BackboneOverride.isEqual;
    rv = custom.namespace.BackboneOverride.original.set.call( this, key, val, options );
    _.isEqual = custom.namespace.BackboneOverride.original.isEqual;

    return rv;
  };
```
Custom isEqual: 
```javascript
  custom.namespace.BackboneOverride.isEqual = function( a, b ) {
    if( ( a instanceof Backbone.Model && b instanceof Backbone.Model ) ||  ( a instanceof Backbone.Collection && b instanceof Backbone.Collection ) ) {
      if ( ( a instanceof Backbone.Model ) && ( a === b ) ) {
        return _.size( a.changedAttributes() ) === 0;
      }

      return custom.namespace.BackboneOverride.original.isEqual( a.toJSON(), b.toJSON() );
    }

    return custom.namespace.BackboneOverride.original.isEqual( a, b );
  };
```

cbroome:
Also as an aside, if there's a discussion thread for suggestions on how to implement service layers in Backbone for dealing with complex APIs, I'd love to know about it :)

jashkenas:
This is an interesting idea, but the particular implementation falls a little short...

The override here is still per-attribute — while for your use case, it seems like the desirable thing would be to have the override method comparing two complete models — so that you can short-circuit and bail out quickly for potentially complex cases. Basically — encompassing the entire `for attr in attrs` loop, instead of overriding within it.

Closing this one, but feel free to try again...

PR #3522, opened 2015-02-28
explicitly pass options with .call to ensure consistent objects

morganrallen:
The rational behind this commit is that the `initialize` method in some situations is unable to override `silent`.

This occurs when the user instantiating a collection does not pass in any options. In this case, the `Collection`
constructor creates an empty object ( https://github.com/jashkenas/backbone/blob/master/backbone.js#L722 ).

Then the user override initialize is called using `.apply` ( https://github.com/jashkenas/backbone/blob/master/backbone.js#L726 )
with `arguments`, then on the next line the `options` object is passed. If the user provided `options` it's all good, because `arguments[1]`
and `options` are the same object. But if they do not provide it, `arguments[1]` is `undefined` and `options` is the new created empty object.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -723,7 +723,7 @@
     if (options.model) this.model = options.model;
     if (options.comparator !== void 0) this.comparator = options.comparator;
     this._reset();
-    this.initialize.apply(this, arguments);
+    this.initialize.call(this, models, options);
     if (models) this.reset(models, _.extend({silent: true}, options));
   };
 
--- a/test/collection.js
+++ b/test/collection.js
@@ -1571,4 +1571,23 @@
     deepEqual(collection.pluck('id'), [1, 3, 2]);
   });
 
+  asyncTest("silence override", function() {
+    var C = Backbone.Collection.extend({
+      initialize: function(models, options) {
+        options.silent = false;
+
+        this.on("add", this.onAdd, this);
+
+        Backbone.Collection.prototype.initialize.call(this, models, options);
+      },
+
+      onAdd: function() {
+        equal(this.length, 1);
+
+        start();
+      }
+    });
+
+    var c = new C(["a"]);
+  });
 })();


morganrallen:
For what its worth, the tests are passing locally for me, and Travis is failing on a different test.

morganrallen:
Weeee! passing tests. Any thoughts on this? I need to know how I'll proceed on my work based on this being accepted or not.

jashkenas:
Good catch.

Ideally, we'd be able to do a version of this patch that still allows for any number of arguments to be passed to the constructor/initialize ... and not limit it to just two.

akre54:
Really? I feel like you shouldn't be allowed to mutate options passed into `initialize`.

What's the use case for doing this in `initialize` instead of where you've invoked the constructor?

jashkenas:
No -- it's not about mutation. It's about arguments[1] being undefined in that apply, instead of an empty object. Right?

jridgewell:
>  It's about arguments[1] being undefined in that apply, instead of an empty object.

akre54:
Right I get that. But what's the use case for needing the same `options` object in your `initialize` if not to mutate it?

If you didn't pass anything down from the instantiation why would you need the `options` object that the constructor creates for itself?

jashkenas:
On second thought, @akre54 is right.

If you don't pass in any options, then suddenly getting an empty options object in your `initialize` would be very weird.

morganrallen:
A couple facts still remain here. 

If the user doesn't pass `options` while instantiating, `.initialize` and `.reset` receive different arguments.

And while I don't see an issue with one constructor mutating the options in the prototype chain, if that is not desired, some other method of overriding `{silent: true}` needs to be available. And saying 'make the user pass in that option' isn't acceptable. I have a collection that MUST trigger events, always and forever, that shouldn't be a user choice. Just as it is forced to be silent, there needs to be a method override that is a child constructor.

morganrallen:
Also, I don't see the rational behind not emitting events in `var c = new Collection([{},{}])` while emitting them when using `c.add(....)`

morganrallen:
@jashkenas As far as the object 'suddenly appearing', every constructor has `options || (options = {})`, so that is already happening.

PR #3523, opened 2015-03-03
Sync test scripts with karma config.

braddunbar:
`setup/dom-setup.js` depends on jQuery so it must be included after.


jridgewell:
Test failure was fixed by https://github.com/jashkenas/backbone/pull/3524. Thanks for fixing up local testing!

PR #3524, opened 2015-03-03
Set the pathname explicitly for phantomjs.

braddunbar:
This fixes the test in phantomjs by setting the pathname to what it would be anyway in a browser.


akre54:
Excellent. Thanks Brad.

PR #3528, opened 2015-03-07
Use global jquery if it exists for CommonJS and UMD

megawac:
Carrying on discussion from #2862, I feel this is the most practical and non-invasive way to support CommonJS and jQuery. jQuery is often loaded via CDN on websites (including those using browserify) or exposed globally for plugin support.

This enables developers using browserify to avoid having to do in many cases (though they would be expected to load jQuery before backbone. They would still be able to select a different version of jQuery for their package ofc by setting BB.$

```
Backbone.$ = global.$ = require('jquery');
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -6,25 +6,25 @@
 //     http://backbonejs.org
 
 (function(root, factory) {
+  var $ = (root.jQuery || root.Zepto || root.ender || root.$);
 
   // Set up Backbone appropriately for the environment. Start with AMD.
   if (typeof define === 'function' && define.amd) {
-    define(['underscore', 'jquery', 'exports'], function(_, $, exports) {
+    define(['underscore', 'jquery', 'exports'], function(_, jQuery, exports) {
       // Export global even in AMD case in case this script is loaded with
       // others that may still expect a global Backbone.
-      root.Backbone = factory(root, exports, _, $);
+      root.Backbone = factory(root, exports, _, jQuery);
     });
 
   // Next for Node.js or CommonJS. jQuery may not be needed as a module.
   } else if (typeof exports !== 'undefined') {
     var _ = require('underscore');
-    factory(root, exports, _);
+    factory(root, exports, _, $);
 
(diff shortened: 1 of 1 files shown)


jashkenas:
Any opinions on this change from the experts?

@jrburke, perhaps?

jrburke:
Extracting the `root.jQuery || root.Zepto || root.ender || root.$` to a variable at the top seems fine. 

For the AMD branch, I would not modify it since asking for 'jquery' as a dependency means it will be loaded/resolved, so no need to change that section. The change just confuses the story there.

For the CommonJS branch, someone with better knowledge of how Backbone is used with CJS would be better to comment. Usually, if there is a dependency, it should be explicitly expressed via `require('jquery')`. However if backbone is used often enough in node without jQuery present, some other steps or guidance might be needed there.

akre54:
Webpack has [`externals`](http://webpack.github.io/docs/library-and-externals.html) to handle cases like this. RequireJS has `paths`, and browserify has browserify-shim. I consider that part of the deal when you're mixing globals and modules in your build tool.

I'm not sure Backbone should be wading into the mixed use territory though. That's a bit outside its purview.

jridgewell:
@megawac Good to close with https://github.com/jashkenas/backbone/pull/2862?

megawac:
I'm not sure, I'd still kind of like to see this added because CDN loading of jquery (and browserify for everything else) is pretty common case - though it's easy enough to work around with browserify-shim

martinoppitz:
Sorry but that is not a solution for Backbone Marionette. Try it by yourself.

npm install backbone.marionette

note: in my case the backbone dep was not installed automatically. Lookup before use.

So. Ich used var Marionette = require(back... mario...);

Backbone.$ undefined.

Why is the factory call in Backbone. Js in exports case different to requirejs and global?

Where are the mistake?

Backbonejs or Backbone.Marionette

jashkenas:
Closing b/c #2862

Issue #2976, opened 2014-01-29
Collection length does not match models length, when fetching and parsing data with duplicate ids


eschwartz:
### Duplication Steps
- Create a model, which parses it's `id` from an arbitrary response object attribute
- Call fetch on a collection for the model
  - API returns multiple models with the same id
- Call fetch a second time
  - API returns multiple models with the same id
- Backbone throws error at 719:46 `TypeError: Cannot read property 'cid' of undefined`

``` javascript
set: function(models, options) {
      // ....

      // Remove nonexistent models if appropriate.
      if (remove) {
        for (i = 0, l = this.length; i < l; ++i) {
          if (!modelMap[(model = this.models[i]).cid]) toRemove.push(model);
          // TypeError; this.models.length !== this.length
        }
        if (toRemove.length) this.remove(toRemove, options);
      }

      // ...
}
```

See [Spec (failing) for mismatched collection length](https://github.com/eschwartz/backbone/commit/5f3f73f6d49cd90e209f3db774ba3e676c93c3e8) for a failing spec, duplicating this issue.
### Expected Behavior

The root issue is clearly that the API is returning duplicate objects. However, I'm wondering if there might be a better way for Backbone to handle this situation, which would make the problem easier to trace. 

What is the default behavior when adding a model to a collection with the same id as an existing model in the collection? I would expect an error to be thrown.

PR #3529, opened 2015-03-10
Fixed duplicate models with parsed ids being added multiple times

fastest963:
Brought up with @akre54 on getter. I think we're waiting for @caseywebdev's input.

Basically the issue is that the attributes are not `parse`d before being sent to `get` so `modelId` doesn't return the right properties. You *could* make `modelId` do the same thing as parse but its not clear you'd have to do that and then you're duplicating functionality that you have to remember to update whenever properties change.

### DIFF
--- a/test/collection.js
+++ b/test/collection.js
@@ -1431,6 +1431,25 @@
     equal(c.models.length, 1);
   });
 
+  test('Do not allow duplicate models (with parsed id) to be `add`ed or `set`', function() {
+    var Stooge = Backbone.Model.extend({
+      idAttribute: '_id',
+      parse: function(a) {
+        var attrs = {};
+        attrs._id = a.__id;
+        return attrs;
+      }
+    });
+    var StoogeCollection = Backbone.Collection.extend({model: Stooge});
+    var c = new StoogeCollection();
+    c.add([{__id: 1}, {__id: 1}], {parse: true});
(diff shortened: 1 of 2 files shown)


test/collection.js
```diff
@@ -1431,6 +1431,25 @@
...
+      idAttribute: '_id',
+      parse: function(a) {
+        var attrs = {};
+        attrs._id = a.__id;
+        return attrs;
+      }
+    });
+    var StoogeCollection = Backbone.Collection.extend({model: Stooge});
+    var c = new StoogeCollection();
+    c.add([{__id: 1}, {__id: 1}], {parse: true});
```
akre54: Can we clean up these tests to use, e.g. a nested resource instead of a weird `id` example? This is kinda hard to follow.

akre54:
Argh the dreaded double-parsing strikes again. It definitely feels weird to force a model's attributes to be parsed in two places (see discussion in #2976), but isn't this better handled in the Collection's parse? We don't tend to like to create temporary models just to merge them in later.

akre54:
After thinking about this a bit more we definitely don't want to be creating temporary models just to set their attributes later on. Agreed that the logic duplication between `modelId` and `parse` isn't ideal, but this isn't the right fix. Let's keep discussing it.

PR #3537, opened 2015-03-16
Update travis configuration

megawac:
This should make travis builds quite a bit faster - this makes use of our 10 concurrent sessions (we had 3 before) and the new `sudo: false` travis option that makes it faster to spin up builds. Builds complete 1-3 minutes faster and boot much faster.

Also adds Safari 8 to the matrix. TODO fix the bug on karma that makes this necessary :/

-----------------

Also if anyone has an extra VPS they want to donate, it'd be nice to set up testing of pull requests for underscore and backbone (https://github.com/jashkenas/underscore/issues/2085). Send me an email or message me on IRC

### DIFF
--- a/karma.conf-sauce.js
+++ b/karma.conf-sauce.js
@@ -4,13 +4,14 @@ var _ = require('underscore');
 var sauceBrowsers = _.reduce([
   ['firefox', '35'],
   ['firefox', '30'],
-  ['firefox', '20'],
+  ['firefox', '21'],
   ['firefox', '11'],
   ['firefox', '4'],
 
   ['chrome', '40'],
-  ['chrome', '35'],
-  ['chrome', '28'],
+  ['chrome', '39'],
+  ['chrome', '31'],
+  ['chrome', '26'],
 
   ['internet explorer', '11', 'Windows 8.1'],
   ['internet explorer', '10', 'Windows 8'],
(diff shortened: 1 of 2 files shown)


karma.conf-sauce.js
```diff
@@ -72,14 +80,11 @@ module.exports = function(config) {
...
 
-    // TODO(vojta): remove once SauceLabs supports websockets.
-    // This speeds up the capturing a bit, as browsers don't even try to use websocket.
-    transports: ['xhr-polling'],
     captureTimeout: 120000,
-    customLaunchers: sauceBrowsers
+    customLaunchers: sauceBrowsers,
 
     // Browsers to launch, commented out to prevent karma from starting
     // too many concurrent browsers and timing sauce out.
```
akre54: remove this comment?

karma.conf-sauce.js
```diff
@@ -22,15 +23,22 @@ var sauceBrowsers = _.reduce([
   ['opera', '12'],
   ['opera', '11'],
 
+  ['android', '5'],
+  ['android', '4.4'],
   ['android', '4.3'],
   ['android', '4.0'],
 
-  ['safari', '8'],
+  ['safari', '8.0', 'OS X 10.10'],
```
akre54: Do we have mobile safari as an option at all? To catch JIT bugs etc?

(I know there's a bug with Object.create and use strict that we might want to look out for)

karma.conf-sauce.js
megawac: Its still an issue - probably should just recomment it out

karma.conf-sauce.js
megawac: Nah, IOS has been kinda buggy on sauce

karma.conf-sauce.js
akre54: is this blocking a merge?

megawac: Nah, its just an annoyance; it just means you have to list the browsers when you start the process (see travis script)

PR #3539, opened 2015-03-17
change child.__super__ to child.super_

manjun-han:
When inheritance structure become more complicated and deep , in  subClass's method we want to call superClass's method , it's  quite long to write the code .

    var app.BaseView = Backbone.View.extend({
        render:function(data){
        // to do
        }
    });

    var app.SlidePanel = app.BaseView.extend({
        render:function(data){
        // call super
        app.BaseView.__super__.render.call(this,data);

        // to do
       }
    })

    var app.CommentSlidePanel = app.SlidePanel.extend({
        render:function(data){
        // call super
        app.SlidePanel.__super__.render.call(this,data);

        // to do
       }
    });

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1803,7 +1803,7 @@
 
     // Set a convenience property in case the parent's prototype is needed
     // later.
-    child.__super__ = parent.prototype;
+    child.super_ = parent.prototype;
 
     return child;
   };


akre54:
`__super__` is mainly accessible for CoffeeScript interoperability and not meant to be used in real code. 

Changing this would break too many CoffeeScript projects. Just call the method on the prototype directly: `SlidePanel.prototype.render.call(this, data)`;

PR #3542, opened 2015-03-20
Wrap commonjs jQuery require in a try-catch

akre54:
Addresses issues from #3421.

* If your app uses jQuery with AMD or global, or you set `Backbone.$` directly, nothing changes.
* If your app uses jQuery with commonjs, you must specify the dependency in your own app (unchanged). However, we now automatically register `Backbone.$` for you so you can remove the setup line from your code.
* If your app doesn't use jQuery and you're using a commonjs loader like webpack or browserify, you must [tell your build tool to exclude or ignore jQuery](https://github.com/jashkenas/backbone/issues/2997#issuecomment-67675558).
* If your app uses a jQuery mock (like Zepto) with commonjs, you need to aliasify (or use ContextReplacementPlugin if Webpack), or set `Backbone.$` directly.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -17,8 +17,9 @@
 
   // Next for Node.js or CommonJS. jQuery may not be needed as a module.
   } else if (typeof exports !== 'undefined') {
-    var _ = require('underscore');
-    factory(root, exports, _);
+    var _ = require('underscore'), $;
+    try { $ = require('jquery'); } catch(e) {}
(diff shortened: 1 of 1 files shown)


jridgewell:
:+1:

dgbeck:
Thx @akre54 for opening this guy and pushing us forward.

samccone:
:+1:

jashkenas:
Let's give it a try.

PR #3543, opened 2015-03-22
Collection.reset should remove options.at to prevent not sorting

fastest963:
Collection.add doesn't sort if you pass options.at but if that somehow gets passed to Collection.reset it passes that along to add which then doesn't sort the collection. options.at in reset is essentially useless
since the collection is emptied before adding.

See failing test for use case. Basically trying to keep 2 collections in sync by listening for `add remove reset sort` and then updating another collection while passing through options (since those might be helpful to things listening on other collection) inadvertently might pass through `options.at` which then causes second collection not to be sorted.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -893,6 +893,8 @@
       }
       options.previousModels = this.models;
       this._reset();
+      // at is not supported for reset and causes sort not to fire in add
+      if (options.at !== undefined) delete options.at;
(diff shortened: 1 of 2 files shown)


fastest963:
This looks like it was a false positive and is actually just a problem that another PR would've introduced. Closing. Sorry for the spam.

PR #3548, opened 2015-03-26
Remove unneeded files from bower package

JustBlackBird:
Files that are useless in bower package should be removed from there.


akre54:
Shouldn't we instead whitelist the files we *do* need?

JustBlackBird:
As far as I know there is no way to list files that should be included in bower package. The only way is ignore unneeded files.

akre54:
Tech.

Issue #2848, opened 2013-11-04
key of a changed attribute


DontShootMe:
hello,

When listening : 

```
model.on("change:attr" , function(model , value , options ){ }) 
```

How to find the attr wich trigger the event ?
why don't pass an object insteadof a value 

```
=> { key : value } or { key : attr , value : value } ?
```

with this enhancement it will be possible to do things like that :+1: 

```
model.on("change:attr" , function(model , changed , options ){

console.log("event fired for : " + changed.key + " with value => " + changed.value);

 })
```

For the moment i don't know how to do that without a big mess !

modification :
https://github.com/jashkenas/backbone/blob/master/backbone.js#L355

```
 - this.trigger('change:' + changes[i], this, current[changes[i]], options);
 + this.trigger('change:' + changes[i], this, { key : changes[i] , value : current[changes[i]] }, options);
```

PR #3549, opened 2015-03-27
Model change:[attribute] event enhancement

vasilgeorgiev:
Set 4th parameter when Backbone.Model triggers event from "change:[attribute]". The parameter is the "attribute" key itself.
The purpose of such change is allowing the ability to create a single method that will handle changes only on relevant attributes and can have options stored per key that has been monitored. 
Such case can be automatic update on specific html parts of the view without rendering the entire view when we are talking for large views that have complicated visual representation (ongoing animations when change or complicated conditional logic for the partial that will increase the overhead of the application).  
I have been using said change for 4 months now in one of my projects.

Example:
```javascript
Backbone.View.extend(
{
    // define object to store actions and callbacks for changes
	components: {},
    // boolean flag for those views that wont use this functionality
	autoUpdate: true,

	initialize: function(){
		this.setAutoUpdates();
		this.trackForUpdate();
	},

	/** Tack for update functionality
	 * use createComponents to create components that need to be updated
	 * 
	 * trackForUpdates will create changes specific for those components
	 * @items - object containing elements for tracking that are not added initially
	 * 
	 * updateComponent is general callback for updates
	 * 
	 * Called when some of the attributes of the model are changed.
	 * The method checks in components object in current view for the key of the changed
	 * attribute and if such key exist takes action
	 *
	 * components - object that contains monitored attributes of the views model
	 * contains objects in a following format {elem: "", action: "", callback: ""}
	 * @elem jQuery object: points to DOM element that needs to change
	 *
	 * @action string/function not required: action is function that is executed and supposed to change the elem.
	 * If string the code assumes that this is jQuery method and will try to directly implement it to the elem with argument the new value.
	 * Example: {elem: $(".my-element"), action: "val"}
	 * 	is equal to $(".my-element").val(value);
	 * If function the codes will execute that function with applied variables
	 * Example: {elem: $(".mu-element"), action: myFunction"}
	 *  is equal to myFunction(new_value, prev_value, element, key, view);
	 *  where
	 *  new_value : the new value of the attribute
	 *  prev_value: the previous value of the attribute
	 *  element : the jQuery element that is supposed to be updated
	 *  key: the key of the attribute changed
	 *  view: reference to the view instance
	 *
	 *  @callback function not required: function that will be executed after the action.
	 *  The function takes new_value, prev_value, element, key, view as arguments. Look at action function for details!
	 */

	trackForUpdate: function(items)
	{
		var forRegistration = {};
		
		if(!items)
		{
			if(this.components && !_.isEmpty(this.components))
			{
				forRegistration = this.components;
			}
		}
		else
		{
			if(_.isObject(items) && !_.isEmpty(items)){
				forRegistration = items;
			}
		}
		
		if(_.isEmpty(forRegistration))
		{
			return;
		}
		
		_.each(forRegistration, function(value, key){
			this.listenTo(this.model, "change:" + key, this.updateComponent);
		}, this);
	},
	/**
	* Single method that calls for specific update behaviour when attributes value changes
	*/
	updateComponent: function(model, value, options, key){
		if(this.components[key] == 'undefined'){
			return;
		}
		
		var element 	= this.components[key].elem,
			action 		= this.components[key].action,
			callback 	= this.components[key].callback,
			prev_value 	= model.previousAttributes()[key];

			if(typeof element == "string"){
				element = this.$el.find(element);
			}
		
			if(typeof action == 'function')
			{
				//execute the function
				action(value, prev_value, element, key, this);
			}
			//else if action is string
			else if(typeof action == 'string')
			{
				//try to execute it
				try
				{
					element[action](value);
				}
				//log error if action is not valid function
				catch(err)
				{
					console.log(action + " is not a function of that object!");
				}
			}

			//if callback is function
			if(typeof callback == 'function')
			{
				//execute the callback
				callback(value, prev_value, element, key, this);
			}
		
	},
	/**
	 * Populates components with elements based on "data-autoupdate" attribute.
	 * This feature will trigger only if 'autoUpdate' property of the view is set to true.
	 * NOTE: "pushComponents" will overwrite auto updated components completely
	 * 
	 * HTML Example:
		<div class="my-container">
			<p data-autoupdate="my_property" >{{ my_property }}</p>
		</div>
	 */
	setAutoUpdates: function(){
		if(!this.autoUpdate == true){
			return;
		}
		
		var $this = this;
		var elements = this.$el.find("[data-autoupdate]");

		elements.each(function(){
			var item = $(this).attr("data-autoupdate");
			if(!$this.components.item){
				$this.components[item] = {"elem": $(this), "action": "html"};
			}
		});
	},
	/**
	* Method to update the components during run time if need be
	*/
	pushComponents: function(push){
		this.components = $.extend(true, {}, this.components, push);
		this.trackForUpdate(push);
	}
});
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -477,7 +477,7 @@
       if (!silent) {
         if (changes.length) this._pending = options;
         for (var i = 0; i < changes.length; i++) {
-          this.trigger('change:' + changes[i], this, current[changes[i]], options);
+          this.trigger('change:' + changes[i], this, current[changes[i]], options, changes[i]);
         }
       }
 


akre54:
Thanks for the detailed pull request. I believe we've covered this in #2848 and others but let me know if I'm mistaken and I'll re-open. 

The gist of the solution is that you should already know the attribute being changed. Either because you've explicitly passed the attribute as the event name (`model.on('change:title', ...)`) or you're looping over a list of attributes and you have access to the attribute in a variable:

```js
_.each(attrs, function(attr) {
  model.on('change:' + attr, function() {
    // do something with the attr variable here
  });
});
```

DontShootMe:
Reading the source i understand that it's better to have option at the end of arguments, so i come back and ask again what i have already asked you for ( [#2848](https://github.com/jashkenas/backbone/issues/2848#) )

> On this demand there is no regression, there is no functionality removing , but adding performance on the "client" side.
> just see how many prop/attr/function exist in Backbone related to retreive the changed attributes !
> But again once, the best thing according to me, will be to provide them !

@akre54 
> The gist of the solution is that you should already know the attribute being changed

yes we know the attribute and we also know the model :
```javascript
 this.listenTo(this.model, "change:" + key, function(model, value, options) {
```
But you provide the complete model !
We have a complete model, a complete options and please provide us a complete attribute, directly from the loop, to augment our performance.

PR #3551, opened 2015-03-29
Add documentation to delegateEvents for cases of rendering child views

thenickcox:
Calling jQuery’s `html()` method in a parent view internally calls
`empty()`, which clears an element’s inner nodes, losing any bound
events on child views. This PR adds documentation to `delegateEvents`
to describe this use case, including how to maintain events on a child
view when the parent view is re-rendered, and includes an example.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2954,6 +2954,36 @@ var DocumentView = Backbone.View.extend({
 });
 </pre>
 
+    <p>
+      Using <b>delegateEvents</b> can additionally be useful when rendering child views
+      into a parent view (e.g., appending <tt>&lt;li&gt;</tt> elements rendered by one view
+      into to a parent <tt>&lt;ul&gt;</tt>, itself rendered by another view). If your parent
+      view responds to changes in the the child <tt>&lt;li&gt;</tt> by calling <tt>render()</tt>,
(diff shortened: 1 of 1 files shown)


thenickcox:
I'm definitely open to feedback on this PR. But it's something I've run into multiple times, and it's never immediately obvious to me why events on child views stop working when the parent view is re-rendered. The wording might be a bit verbose, and the example might not be the best, so feel free to treat this as a work in progress.

I do feel strongly that this is pretty confusing gotcha, and is arguably more related to jQuery’s implementation of `html()`, but `this.$el.html(…)` has long been the canonical render method body, and so this is something a lot of folks might conceivably run into, as I have multiple times.

shuvalov-anton:
+1. I come to the same problem with `.render()` and `.delageteEvents()` and I had wasted all day to debug it.

jashkenas:
Thanks, but no thanks — this isn't a very nice way to write it. If it's always 1:1 like this, you can just get rid of your subview, and put all of the HTML in the parent's template. Or you can put the events on the parent view. Or you can avoid blowing away the child in the first place. — *All* of which are nicer than this pattern.

thenickcox:
@jashkenas Noted. Thanks.

PR #3553, opened 2015-04-02
Use _.create in extend

megawac:
As mentioned here https://github.com/jashkenas/underscore/pull/2074


jashkenas:
Is it worth breaking backwards compatibility so soon for this? Not worth waiting until `_.create` has a chance to penetrate for a while?

megawac:
Just raising an issue, can sit on it. It shouldn't break backwards compat unless people are relying on old versions of underscore

jridgewell:
The extend vs assign (in create) is a midly breaking change. I think the common case, though, is simple objects.

backbone.js
```diff
@@ -1793,14 +1793,8 @@
...
-    // `parent`'s constructor function.
-    var Surrogate = function(){ this.constructor = child; };
-    Surrogate.prototype = parent.prototype;
-    child.prototype = new Surrogate;
-
-    // Add prototype properties (instance properties) to the subclass,
-    // if supplied.
-    if (protoProps) _.extend(child.prototype, protoProps);
+    // `parent`'s constructor function and add the prototype properties.
+    child.prototype = _.create(parent.prototype, protoProps);
```
jridgewell: You'll also need to set `child.prototype.constructor = child`.

jridgewell:
Ping @megawac, want to rebase and add in the `child.prototype.constructor = child` change?

megawac:
Right, done

jridgewell:
LGTM.

akre54:
April's long enough. Let's do this.

PR #3568, opened 2015-04-16
Added Events.onTry method for handling asynchronous errors 

dsheiko:
When working with asynchronous flows handling exceptions may become a headache. Promises API provides catch method where exceptions can be caught after the call is done. While working with Backbone one is on his own. This simple solution provides a unified way for handling async exceptions.

`onTry` extends `Events.on` by wrappping the callback in try/catch what 
in the case of exception triggers `error` event in the context and calls a given error method if it's available

Examples of use:
```javascript
// -----------
// Example 1
// -----------
// First specify what we do with exceptions
view.on("error", function(context, err){
 console.error(err);
});
// Now subscribe with exception handling
view.onTry("test-event", function(){
  // oops
	throw new Error("Test exception");
});
// Trigger async
setTimeout(function(){
 view.trigger("test-event");
}, 50);
```
```javascript
// -----------
// Example 2
// -----------
// First specify what we do with exceptions
var globalErrorHandler = function(err) {
  console.error(err);
	// we still keep the context
	// console.dir( this );
};

// Now subscribe with exception handling
view.onTry("test-event", function(){
  // oops
	throw new Error("Test exception");
}, globalErrorHandler, view);

// Trigger async
setTimeout(function(){
 view.trigger("test-event");
}, 50);
```


jridgewell:
Thanks for the PR. I'm not convinced that this is right for Backbone, though.

In promises, you have to emulate normal `return` / `throw` behaviors since an `onFullfilled` callback is only passed `value` and not a `resove` / `reject`. Additionally, promises need to chain into the next promise fluidly.

Neither of those is true for Backbone since chaining of events isn't natively supported (you can `this.trigger` from inside your listener, but that's not the same as chaining multiple return values together). The one improvement I see over using a try-catch in any potentially-throwing listener is putting it in a higher scope than in the hot-path listener. But, Underscore's [already rejected](https://github.com/jashkenas/underscore/issues/1727) a generalized version of that.

As a final point, Backbone already has a [defined](http://backbonejs.org/#Events-catalog) `error` event for when a model's or a collection's request to remote server fails.

I think this is best left as a mixin for now.

dsheiko:
Hello jridgewell,
I agree that the solution isn't that elegant, just one working for me. Yet, the problem is not about chaining. Look at that:
```javascript
try {
  model.on("event", function(){
  throw new TypeError("Test");
  });
  setTimeout(function(){
    model.trigger("event");
  }, 50);
} catch( err ) {
// Nothing comes here
}
```
Top scope try/catch won't catch exception that happens during an async call. When browsers at least report such exceptions, in NW.JS you won't even get notified about the let's say TypeError in the called code. If you want gracefully report the error to the UI, you have no option other than wrapping any callback code in try/catch manually. But that the problem relevant only  for asynchronous flows.

akre54:
Try [trigger-then](https://github.com/bookshelf/trigger-then) by one of the Backbone contributors. Might be what you're looking for.

jridgewell:
You're writing the try-catch in the wrong function:

```javascript
model.on("event", function(){
  try {
    throw new TypeError("Test");
  } catch( err ) {
    this.trigger('error', err);
  }
});
setTimeout(function(){
    model.trigger("event");
}, 50);
```

This is the same reason you can't wrap a new Promise in a try-catch. Async blocks don't respect Sync catching. But, Promises were designed to mimic Sync control flows, so they have you `return` and `throw`. Promises follow this pattern so that you don't have to keep using `resolve` and `reject` in `onFullfilled` callbacks. Imagine if you had to do this:

```javascript
Promise.resolve(1).then(function(data, resolve, reject) {
    reject(new Error('test'));
}).catch(function(err, resolve, reject) {
    // do something
});
```

You might argue that `#onTry` would take Backbone from being the above's horrible version of Promises to being what real Promises are:

```
// Analogous to Promise example above
Backbone.on('event', function() {
    this.trigger('error', new Error('test'));
});
Backbone.on('error', function(err) {
    // do something
});
Backbone.trigger('event');
```

 But, Promises create tree-like chains of callbacks wired together. Backbone events aren't tree-like, they're completely flat. Since events don't chain together, we're not in a better situation. This just defines a sugar syntax that doesn't add much, since we can't define which error event to chain into next, or how to resolve an error into a new non-error event.

PR #3571, opened 2015-04-16
Backbone.Event.all for handling parallel async calls

dsheiko:
This method resolves when all of the events of the supplied array have resolved similar to Promise.all()

## Basic Usage

The `all` method can be used where you need the moment when a row of concrete asynchronous events resolved:
```html
<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/jquery/2.1.3/jquery.min.js"></script>
<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/underscore.js/1.8.3/underscore-min.js"></script>
<script type="text/javascript" src="https://cdnjs.cloudflare.com/ajax/libs/backbone.js/1.1.2/backbone.js"></script>
<script type="text/javascript" src="https://raw.githubusercontent.com/dsheiko/backbone-event-all/master/build/backbone-event-all.min.js"></script>
<script type="text/javascript">

var View = Backbone.View.extend({}),
    // Create a concrete view (or any other Backbone EventEmitter)
    view = new View();
// Subscribe to the event when both `foo` and `bar` resolved regardless the order they fired
view.all( [ "foo", "bar" ], function( tasks ){
 console.log( "`foo` and `bar` both resolved" );
 console.log( "`foo` emitted with parameters", tasks.foo );
 console.log( "`bar` emitted with parameters", tasks.bar );
});
// Fire the events
setTimeout(function(){
 console.log( "`bar` fired" );
 view.trigger( "bar", "bar" );
}, 100 );

console.log( "`foo` fired" );
view.trigger( "foo", "foo" );

</script>
```


akre54:
Backbone events are meant to be synchronous. For an async-friendly implementation, try [trigger-then](https://github.com/bookshelf/trigger-then).

PR #3576, opened 2015-04-23
Add badges to README

SimenB:
I'm aware of https://github.com/jashkenas/backbone/pull/1876#issuecomment-11148492, but reading the README on [npmjs.com](https://www.npmjs.com/package/backbone) is pretty normal now, I'd think.

I added an npm badge as well. It's an easy way to identify latest release.

Also made the links into actual clickable links. It works fine here on GitHub, but npmjs.com doesn't automatically create links.

If it's accepted, I could do the same for underscore.

EDIT: According to [this](http://daringfireball.net/projects/markdown/syntax#autolink) adding `<>` to the links should be enough, I'm not sure though. Anybody knows?

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,6 +10,7 @@
     (_'_______________________________________________________________________________'_)
     (_.———————————————————————————————————————————————————————————————————————————————._)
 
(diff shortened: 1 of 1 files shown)


jashkenas:
Not a big fan of badges.

PR #3577, opened 2015-04-23
Always call success async

jridgewell:
Since the `success` callback is called async when `model` isn't new, we should always call `success` async.

Let's not [unleash Zalgo](http://blog.izs.me/post/59142742143/designing-apis-for-asynchrony).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -639,18 +639,18 @@
       };
 
       options.success = function(resp) {
-        if (wait || model.isNew()) destroy();
+        if (wait) destroy();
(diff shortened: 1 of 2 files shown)


braddunbar:
Ok by me, though we should probably write this up in the notes. :+1:

jridgewell:
Notes?

braddunbar:
Release notes. Sorry, should've been more specific! :+1:

jashkenas:
Seems wise.

PR #3579, opened 2015-04-28
The 1.1.2 distribution on npm is different than the one on github

coltonTB:
There are discrepancies between the code here and the distribution on NPM, although they both say 1.1.2. For example, the History implementation hosted on github uses `getPath` within `getFragment` ([here](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1562)), while `getFragment` from the current 1.1.2 distribution on npm does not:
```
    getFragment: function(fragment, forcePushState) {
      if (fragment == null) {
        if (this._hasPushState || !this._wantsHashChange || forcePushState) {
          fragment = decodeURI(this.location.pathname + this.location.search);
          var root = this.root.replace(trailingSlash, '');
          if (!fragment.indexOf(root)) fragment = fragment.slice(root.length);
        } else {
          fragment = this.getHash();
        }
      }
      return fragment.replace(routeStripper, '');
    },
```


jridgewell:
You're looking at master. Here's [v1.1.2's code](https://github.com/jashkenas/backbone/blob/1.1.2/backbone.js).

PR #3580, opened 2015-04-28
If the root doesn't match, no routes can match.

braddunbar:
It's hard to believe this has gone unreported for so long, but here it is! Currently, a route like 'test' will match the url `/test`, regardless of the root, and the url `/xxxx/test`, as long as the root is four characters long. This is obviously incorrect as routes should only be executed when the root matches.

There may be further implications from matching the root, but this is certainly more correct than what exists currently.

Thanks to @pfiller for catching it! :heart:

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1529,7 +1529,9 @@
 
     // Does the pathname match the root?
     matchRoot: function() {
-      return this.rootMatcher.test(this.location.pathname);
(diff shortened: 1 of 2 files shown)


akre54:
Nice. Always better late than never

pfiller:
Thanks for the amazingly fast fix, @braddunbar! Thanks for merging @akre54! Wowzers!

PR #3582, opened 2015-04-28
Align few places to match with possible overriding of Backbone.sync

Artazor:
Overriding Backbone.sync is a great way to switch a persistence layer. It just lacks two moments that should be overridable too:

* Result of `model.save` if model was not validated
* Result of `model.destory` if model was new

For example if my implementation of Backbone.sync returns promises, then current implementation forces me to check the result before calling '.then()'. With the proposed changes, we will be able to unify `sync` behavior with the rest of the Backbone.

@jashkenas ?

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -586,9 +586,9 @@
       // `set(attr).save(null, opts)` with validation. Otherwise, check if
       // the model will be valid when the attributes, if any, are set.
       if (attrs && !wait) {
-        if (!this.set(attrs, options)) return false;
+        if (!this.set(attrs, options)) return Backbone.notValidated(this);
       } else {
-        if (!this._validate(attrs, options)) return false;
+        if (!this._validate(attrs, options)) return Backbone.notValidated(this);
       }
 
       // Set temporary attributes if `{wait: true}`.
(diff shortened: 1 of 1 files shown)


jridgewell:
I think this would be more useful if we always returned a thenable?

Artazor:
@jridgewell It would be awesome, but it will be backward incompatible (I suppose).

Anyway there will be a leak of thenable in `Collection::create()` as long as it uses thenable-returning operation (`Model::save()`) but does not return created thenable (it returns a model)

Artazor:
Anyway this PR will help me not to maintain my own version of Backbone -)
Would appreciate any ideologically compatible change/merge

akre54:
> I think this would be more useful if we always returned a thenable?

Time to revisit #2489?

Artazor:
@akre54 Yes its just about the same issue. However I think that the current PR is way more tiny, yet sufficient to use watever promise library you want. 

However Backbone can implement it's own thenable stubs since in both mentioned cases returned promises are already settled (i.e. have very limited range of reactions for `.then()`)

jridgewell:
> Time to revisit #2489?

:+1:

jashkenas:
Let's go revisit over there then.

PR #3584, opened 2015-04-29
Fix grammar in Helpers documentation.

kuychaco:
Remove comma from comment. No changes to source code made.


akre54:
These docs are auto-generated from the source. Editing the generated version does nothing.

PR #3585, opened 2015-04-29
Improve consistency for object notation.

kuychaco:
Use dot notation for property assignment rather than bracket notation.


jridgewell:
`class` is a reserved word, and this will cause older browsers to throw an error.

PR #3586, opened 2015-04-29
Improve consistency for object notation.

kuychaco:
Use dot notation for property assignment rather than bracket notation.


jridgewell:
Closing with https://github.com/jashkenas/backbone/pull/3585. Also, there's no need to modify the docs, they're updated automatically _only_ when we release a new version.

PR #3589, opened 2015-04-30
Capitalize constructor functions.

kuychaco:
Indicate that 'parent' and 'child' in the 'extend' function are constructor functions by capitalizing them per convention, clarifying their purpose.


jashkenas:
No. They aren't particular classes – they're classes that vary every time the function is called. Lowercase variable names are better.

PR #3590, opened 2015-04-30
Clarify reference to the parent constructor.

kuychaco:
The parent's constructor is 'Function', which is the constructor that created parent. 
The comments refer to the parent constructor function, not it's constructor function (that is, the constructor that created it).


akre54:
Sure, looks good. Thanks.

PR #3598, opened 2015-05-05
Fix #3597 revert annotated src to v1.1.2

### LINKED ISSUE #3597: Annotated source version number is misleading
The published annotated source gives the impression that it is the annotated source of version 1.1.2

screenshot 2015-05-05 16 12 27
http://backbonejs.org/docs/backbone.html

which is misleading, as it's really whatever the state of master was at last `redoc`. I lost a bunch of time thinking I'd gone mad as a consequence. That hasn't been ruled out.

Would you consider:
1. bumping the version number to `1.1.3-rc` in `backbone.js` on master, so that the generated docs are less surprising.
2. ...or generating the docs from the `1.1.2` tag, so the contents match the the release.
3. ...or publishing both stable & master docs, labelled as such.

I'm happy to make a PR, but wanted to see which route is preferred.

olizilla:
Updated `docco` to 0.7.0 and ran against `backbone.js` from gh-pages branch, to revert `backbone.html` to the v1.1.2 tag.

### DIFF
--- a/docs/backbone.html
+++ b/docs/backbone.html
@@ -10,33 +10,33 @@
 <body>
   <div id="container">
     <div id="background"></div>
-    
+
     <ul class="sections">
-        
+
           <li id="title">
               <div class="annotation">
                   <h1>backbone.js</h1>
               </div>
           </li>
-        
-        
-        
+
+
+
         <li id="section-1">
             <div class="annotation">
-              
+
               <div class="pilwrap ">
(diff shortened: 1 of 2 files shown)


akre54:
Sorry, didn't see this. Already fixed in 5f21e0089fcc671a950cd902f8a2ad70e81f5e04

I'll go ahead and bump the docco version too.

Issue #1088, opened 2012-03-11
Backbone.View Element Namespace Support


ryanhorn:
The `svg` element (and possibly others) needs to be created with `createElementNS` and the appropriate namespace to render properly. Views that use `svg` need to override the `make` method to create elements using `document.createElementNS` rather than `document.createElement`. See http://stackoverflow.com/questions/9651167/svg-not-rendering-properly-as-a-backbone-view.

PR #3600, opened 2015-05-11
Allow specifying XML namespace in Backbone views

SylvainCorlay:
Another namespace than HTML may be specified with the `namespace` view attribute. The `namespace` attribute may be a string such as `'http://www.w3.org/2000/svg'` or a function returning such a string.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1261,8 +1261,10 @@
 
     // Produces a DOM element to be assigned to your view. Exposed for
     // subclasses using an alternative DOM manipulation API.
-    _createElement: function(tagName) {
-      return document.createElement(tagName);
+    _createElement: function(tagName, namespace) {
+        return namespace ?
+            document.createElementNS(namespace, tagName) :
+            document.createElement(tagName);
     },
 
     // Ensure that the View has a DOM element to render into.
@@ -1274,7 +1276,8 @@
         var attrs = _.extend({}, _.result(this, 'attributes'));
         if (this.id) attrs.id = _.result(this, 'id');
         if (this.className) attrs['class'] = _.result(this, 'className');
-        this.setElement(this._createElement(_.result(this, 'tagName')));
+        this.setElement(this._createElement(_.result(this, 'tagName'),
+                                            _.result(this, 'namespace')));
         this._setAttributes(attrs);
       } else {
         this.setElement(_.result(this, 'el'));


jridgewell:
What is the benefit of this?

SylvainCorlay:
Enabling backbone views linked to some SVG `<g>` element for example. Currently, backbone only handles HTML elements.

jridgewell:
I'm inclined to say :-1: to this, but I'll defer to one of the other maintainers. This has also been brought up in #1357 and #1088 -- both dormant for 3 years -- but those were shot down due to bad browser support.

Giving Views another "special" HTML property (in addition to `tagName`, `id`, `className`, and `attributes`) doesn't seem like a huge win, since the core use of Backbone is HTML. This is also much easier to implement by Devs since #3003 added View Hooks.

```javascript
  _createElement: function(tagName) {
    var namespace = _.result(this, 'namespace');
    return namespace ? document.createElementNS(namespace, tagName) : document.createElement(tagName);
  }
});
```

SylvainCorlay:
@akre54 suggested I submit a patch for this in https://github.com/akre54/Backbone.D3View/issues/2.
(Backbone.D3View has a `namespace` attribute). 

Bu the way, it seems that these issues remained dormant because of poor browser support for svg. I don't think is still the case and when things like d3.js have become mainstream.

backbone.js
```diff
@@ -1261,8 +1261,12 @@
 
     // Produces a DOM element to be assigned to your view. Exposed for
     // subclasses using an alternative DOM manipulation API.
-    _createElement: function(tagName) {
-      return document.createElement(tagName);
+    _createElement: function(tagName, namespace) {
+        if (typeof namespace === 'undefined') {
```
braddunbar: Can this just be `if (namespace == null) {`.

braddunbar:
I *have* wanted to do this in the past in order to use Backbone with SVG elements. That said, writing your own `_createElement` is really easy to do and I might just recommend that. This change implies that Backbone Views support SVG elements and I'm not sure what other land mines surround that implication because I don't have tons of experience with it. I dunno!

backbone.js
SylvainCorlay: sure, then even
```JavaScript
return namespace ?  document.createElementNS(namespace, tagName) : document.createElement(tagName);
```
Although if you are strict about comparisons with `undefined`, you would probably not want to do this.

SylvainCorlay:
@braddunbar `_ensureElement` also calls `_createElement` and would have to be overloaded as well. It makes it less simple.

backbone.js
jridgewell: The `== null` or ternary would be better than a `typeof` check. If you check the source, Backbone uses `== null` several times.

jridgewell:
> _ensureElement also calls _createElement and would have to be overloaded as well.

Not necessarily. I agree that if we accept this, it should be grabbed in `#_ensureElement` and passed to `#_createElement`, but even `D3View` [just overrides](https://github.com/akre54/Backbone.D3View/blob/master/backbone.d3view.js#L49-L54) `#_createElement` to add basic support.

backbone.js
SylvainCorlay: Ok. Will make the change (I used typeof because I saw it elsewhere in backbone and I wanted to follow your standards.) 
By the way, the shortest way to get the real `undefined` is `void 0`, even when undefined is defined. This is what is done in underscore.

braddunbar:
Right! What @jridgewell said. I was just typing out a similar response.

SylvainCorlay:
ok, rebased including your remarks. it is really trivial. Just waiting for your final sentence.

backbone.js
```diff
@@ -1262,7 +1262,10 @@
     // Produces a DOM element to be assigned to your view. Exposed for
     // subclasses using an alternative DOM manipulation API.
     _createElement: function(tagName) {
-      return document.createElement(tagName);
+        namespace =  _.result(this, 'namespace');
```
jridgewell: Sorry to confuse you, but `_.result(this, 'namespace')` should be in `#_ensureElement` to keep all our access in the same method. My [basic implementation](https://github.com/jashkenas/backbone/pull/3600#issuecomment-100942084) was just a stopgap in case we didn't decide to support namespaces.

SylvainCorlay: gotcha

akre54:
> @akre54 suggested I submit a patch for this in akre54/Backbone.D3View#2.
(Backbone.D3View has a namespace attribute).

I meant on the Backbone.D3View repo. For the record I'd rather see this in userland or plugin code, I think what we have now works fine.

jridgewell:
>  For the record I'd rather see this in userland or plugin code, I think what we have now works fine.

I think that's three "mehs", closing for now. Thanks for bringing it up, though, @SylvainCorlay.

SylvainCorlay:
Too bad. In the context of IPython widgets, we use svg namespace all the time. For now I am working around backbone 1.1.2 but this would have made things cleaner in my opinion.

mborman:
For anyone coming across this issue see http://stackoverflow.com/a/9652241/4479273 for a super simple solution from the author of D3 (mbostock).

In a nutshell, use D3 to append the svg element to your view. D3 will add the proper namespace.

Here is what I did in `initialize`. 

    d3.select(this.el).append('svg');

It could also do as part of `render` depending on your needs.

Ricyteach:
BB newbie here: I'm sorry to comment on a 3 year old thread, but would someone be willing to walk me through-- on a bit of a more basic level-- the correct way to attach a view to an SVG element for today's version of BB?

I'm trying to build what I believe is a relatively simple widget using SVG and ipywidgets and [was bitten](https://stackoverflow.com/questions/56659218/fix-broken-display-of-svg-using-ipywidgets-domwidget-in-jupyter-notebook) by this namespace problem. As a python developer coming to this with almost no knowledge of javascript to start with, much less Backbone, it took me DAYS to even figure out what to search for to be able to understand why `createElement()` was not rendering an SVG to the page.

As a reply to the previous comments above that say, basically "all you have to do is foo the bar", try putting yourself in the shoes of someone like me. I suggest that it might be worth making this a little easier to do, or at least make finding out how to do it a little easier to fine, given the huge popularity of SVG.

akre54:
I would suggest creating the element yourself ([see MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/createElementNS)) and then passing to your view as the `el` prop.

Something like this should work:

```js
const svg = document.createElementNS('http://www.w3.org/2000/svg', 'svg');
const view = new MyView({el: svg});
```

You could also override `_createElement` in your View (as this pull request offers to do). A good pattern might be to create a base `NSView` that would allow you to use any svg tag seamlessly:

```
const NSView = Backbone.View.extend({
  _createElement(tagName) {
    return document.createElementNS('http://www.w3.org/2000/svg', tagName);
  }
});

const MySVG = NSView.extend({
  tagName: 'svg'
});
```

You might also want to think about using d3 for manipulating svgs, as that is a bit easier to work with.

SylvainCorlay:
@Ricyteach I recommend you check out the bqplot code base.

Marks (plot items) are generally attached to an svg-namespaced element, and we have to deal with the namespace thingy.

Ricyteach:
Excellent, thanks to you both that is exactly what I needed to know. 

I don't relish the idea of adding a dependency like d3 to something so simple. The main reason being: I do not write JavaScript but more than once every six months or so at the most. This will almost certainly be the most complicated thing I've ever put together in js.

Every time I come back to js, I literally have had to start with a tutorial to even remember what anything that doesn't look kind of like python (eg, "what the hell is a prototype again?" "Wait, when do I use commas and semicolons again?" or even "ugh how do I debug in a browser again?") does. It would be even worse for a dependency. So I will probably get this working, use it for a couple years, and then have to relearn everything it is doing when I need to make a change later. 

Anything that isn't python makes my head spin. I'm sure I'll get over that eventually but I'm 100% self taught, and I'm not even a professional developer (I'm a civil engineer; the app I'm writing is to help me edit existing finite element meshes in a gui for a FEA program that doesn't have one) and learning something I am extremely likely to forget a year from now seems like a bad use of time. 

Sorry I'm sure that was way more info than needed.

HeywoodKing:
Very good

发送自 Windows 10 版邮件<https://go.microsoft.com/fwlink/?LinkId=550986>应用

________________________________
发件人: Rick Teachey <notifications@github.com>
发送时间: Thursday, June 20, 2019 6:47:32 PM
收件人: jashkenas/backbone
抄送: Subscribed
主题: Re: [jashkenas/backbone] Allow specifying XML namespace in Backbone views (#3600)


Excellent, thanks to you both that is exactly what I needed to know.

I don't relish the idea of adding a dependency like d3 to something so simple. The main reason being: I do not write JavaScript but more than once every six months or so at the most. This will almost certainly be the most complicated thing I've ever put together in js.

Every time I come back to js, I literally have had to start with a tutorial to even remember what anything that doesn't look kind of like python (eg, "what the hell is a prototype again?" "Wait, when do I use commas and semicolons again?" or even "ugh how do I debug in a browser again?") does. It would be even worse for a dependency. So I will probably get this working, use it for a couple years, and then have to relearn everything it is doing when I need to make a change later.

Anything that isn't python makes my head spin. I'm sure I'll get over that eventually but I'm 100% self taught, and I'm not even a professional developer (I'm a civil engineer; the app I'm writing is to help me edit existing finite element meshes in a gui for a FEA program that doesn't have one) and learning something I am extremely likely to forget a year from now seems like a bad use of time.

Sorry I'm sure that was way more info than needed.

―

Issue #3566, opened 2015-04-14
Use Function("return this") to get global?


creatorrr:
Backbone currently breaks in `'strict'` mode while trying to access properties on the global object. The global object is being set using the `(function(root) { ... })(this)` pattern. However, in this case, `root` remains `undefined` and hence the issue.

Using the `Function("return this")()` pattern instead alleviates the issue, though arguably ugly. Thoughts?

**Steps to reproduce**: Import backbone in a `babel-jest` test harness. (uses strict mode)

creatorrr:
This issue affects [Underscore](/jashkenas/underscore) as well. Relevant issue: jashkenas/underscore#2152

PR #3602, opened 2015-05-13
Amend #3566 fix; global detection

megawac:
Up for discussion, see https://github.com/jashkenas/underscore/pull/2153 comments

Amendment to 5a6ffcb873ec555bda9d623b411a243da4efc9a7


jashkenas:
Thanks. (And thanks for putting in the comment.) Makes sense.

PR #3603, opened 2015-05-13
Juggle context param with event maps

jridgewell:
When using event maps, it's a little nicer to specify the `context` parameter as the second argument. This keeps backwards compatibility with 1.1.2, though the feature wasn't documented.

```javascript
obj.on({
    a: method,
    b: other
}, context);
```

When you explicitly pass a `context`, we don't override.

```javascript
obj.on({
    a: function() { ok(this === context); },
}, otherContext, context);
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -96,6 +96,7 @@
     var i = 0, names;
     if (name && typeof name === 'object') {
       // Handle event maps.
+      if (callback !== void 0 && 'context' in opts && opts.context === void 0) opts.context = callback;
       for (names = _.keys(name); i < names.length ; i++) {
         memo = iteratee(memo, names[i], name[names[i]], opts);
       }
--- a/test/events.js
+++ b/test/events.js
@@ -66,6 +66,24 @@
     equal(obj.counter, 5);
   });
 
+  test("binding and trigger with event maps context", 2, function() {
+    var obj = { counter: 0 };
+    var context = {};
+    _.extend(obj, Backbone.Events);
+
+    obj.on({
+        a: function() {
+            strictEqual(this, context, 'defaults `context` to `callback` param');
+        }
+    }, context).trigger('a');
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -97,6 +97,7 @@
     if (name && typeof name === 'object') {
       // Handle event maps.
       for (names = _.keys(name); i < names.length ; i++) {
+        if (callback && 'context' in opts && opts.context === void 0) opts.context = callback;
```
megawac: Why are you doing this in the loop?

jridgewell: Oversight, whoops.

paulfalgout:
:+1:

paulfalgout:
As it turns out this also explained the call stack error I was getting earlier.  I was using a backbone.radio channel and setting the context of a hash of events as in this issue, but one of the events listened to was `all`.  It was then attempting to trigger an event on the object using the radio channel, but since the context was wrong it was triggering on itself.

Here's a simple example: http://codepen.io/paulfalgout/pen/pJypOb

```js
var channel = new Backbone.Model();

$(window).on('mousedown', function(){
  channel.trigger('something');
});

var Service = Backbone.Model.extend({
  initialize: function(){
    // This works
    //channel.on('all', this.triggerActivity, this);
    
    // This doesn't
    channel.on({ 'all': this.triggerActivity }, this);
  },
  triggerActivity: function(){
    console.log('about to trigger');
    // in bb 1.2.0 this == channel causing the callstack exceeded bug
    // in bb 1.1.2 this == Service
    this.trigger('activity');
  }
});

new Service();
```

paulfalgout:
Is this not something that'd go out with `1.2.1`?  @jashkenas

jridgewell:
I'll merge this tomorrow unless there are objections.

PR #3605, opened 2015-05-14
Model.set optimization for speed

gaperton:
- Optimized comparison function for primitive types
- Optimized previous attributes clone operation

As measured in Chrome, Model.set now:
- 10x faster with primitive types
- 3x faster with other types

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -42,9 +42,7 @@
   var _isEqual = _.isEqual;
 
   function isEqual( a, b ){
-      var typea = typeof a;
-      if( typea !== typeof b ) return false;
-      return typea !== 'object' || !a || !b ? a === b : _isEqual( a, b );
+      return a && b && typeof a === 'object' && typeof b === 'object' ? _isEqual( a, b ) : a === b;
   }
 
   // Current version of the library. Keep in sync with `package.json`.
--- a/backbone.js
+++ b/backbone.js
@@ -38,6 +38,15 @@
   var array = [];
   var slice = array.slice;
 
+  // Optimized deep comparison function.
+  var _isEqual = _.isEqual;
+
+  function isEqual( a, b ){
+      var typea = typeof a;
+      if( typea !== typeof b ) return false;
+      return typea !== 'object' || !a || !b ? a === b : _isEqual( a, b );
+  }
+
   // Current version of the library. Keep in sync with `package.json`.
   Backbone.VERSION = '1.1.2';
 
@@ -346,11 +355,13 @@
       changing        = this._changing;
       this._changing  = true;
 
+      current = this.attributes, prev = this._previousAttributes;
+
       if (!changing) {
-        this._previousAttributes = _.clone(this.attributes);
+        prev = this._previousAttributes = {};
+        for( attr in current ){ prev[ attr ] = current[ attr ]; }
         this.changed = {};
       }
-      current = this.attributes, prev = this._previousAttributes;
 
       // Check for changes of `id`.
       if (this.idAttribute in attrs) this.id = attrs[this.idAttribute];
@@ -358,8 +369,8 @@
       // For each `set` attribute, update or delete the current value.
       for (attr in attrs) {
         val = attrs[attr];
-        if (!_.isEqual(current[attr], val)) changes.push(attr);
-        if (!_.isEqual(prev[attr], val)) {
+        if (!isEqual(current[attr], val)) changes.push(attr);
+        if (!isEqual(prev[attr], val)) {
           this.changed[attr] = val;
         } else {
           delete this.changed[attr];
@@ -421,7 +432,7 @@
       var val, changed = false;
       var old = this._changing ? this._previousAttributes : this.attributes;
       for (var attr in diff) {
-        if (_.isEqual(old[attr], (val = diff[attr]))) continue;
+        if (isEqual(old[attr], (val = diff[attr]))) continue;
         (changed || (changed = {}))[attr] = val;
       }
       return changed;
--- a/backbone.js
+++ b/backbone.js
@@ -38,13 +38,6 @@
   var array = [];
   var slice = array.slice;
 
-  // Optimized deep comparison function.
-  var _isEqual = _.isEqual;
-
-  function isEqual( a, b ){
-      return a && b && typeof a === 'object' && typeof b === 'object' ? _isEqual( a, b ) : a === b;
-  }
-
   // Current version of the library. Keep in sync with `package.json`.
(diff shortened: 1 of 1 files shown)


akre54:
This is a pretty smart idea, but isn't Underscore the better place to make this change?

We typically like to see benchmarks run through [jsperf](https://jsperf.com). Mind creating a simple test case?

braddunbar:
Yea, that was my thought as well. Looks like a nice addition to `_.isEqual`.

backbone.js
```diff
@@ -38,6 +38,13 @@
   var array = [];
   var slice = array.slice;
 
+  // Optimized deep comparison function.
+  var _isEqual = _.isEqual;
+
+  function isEqual( a, b ){
+      return a && b && typeof a === 'object' && typeof b === 'object' ? _isEqual( a, b ) : a === b;
```
jridgewell: This'll hit issues with `NaN`, since it isn't reflexive.

megawac: This won't work for `0 === -0` or `NaN`. I expect this to be faster for cases where `_isEqual(a, b)` is false but roughly the same when `a === b` (minus JIT optimizations)

backbone.js
```diff
@@ -346,11 +353,13 @@
       changing        = this._changing;
       this._changing  = true;
 
+      current = this.attributes, prev = this._previousAttributes;
+
       if (!changing) {
-        this._previousAttributes = _.clone(this.attributes);
+        prev = this._previousAttributes = {};
+        for( attr in current ){ prev[ attr ] = current[ attr ]; }
```
megawac: Not a big fan of this without evidence. This is essentially what `clone` does anyway (but it uses an extra loop). This will also not support some of the iteration bugs for things like `hasOwnProperty`, etc. that we've fixed in underscore

gaperton: I'm not a big fan of this either, but I've got a test showing 10x performance improvement in Chrome. I will share it.

In the case of Model.attributes it's just a simple hash, so there are no any need for "hasOwnProperties". So, in this special case this is an opportunity for optimisation. Cause Model.set is slow like hell, because of _.clone and _.isEqual.

backbone.js
gaperton: > 0 === Nan
false

> 0 === -0
Are you kidding?

gaperton: _.isEqual is terribly slow on typical Model.set use cases, like assignments of primitive values like integers. This is not just "faster", but in fact it's 10x faster (open profiler and see), no matter what someone might expect. No one can really predict how JIT works on real program.

gaperton: Also, NaN is not serializable to JSON, so it's not an issue for Models. Although, it won't hit issues with NaN either. NaN !== NaN, so what. Bound case your application isn't supposed to work in either, what's the deal if you'll receive extra change event.

gaperton:
Ok, I will create the test for jsper, when I will have some spare time. Not sure how can I do it with a framework patch, though. But may be, later, if I won't be lazy...

Or you can make local test as I did, and see the difference I reported. Just try to make model.set( 'attr', 5 ) 1 million times.

Or you can add performance test to the backbone test suite.

megawac:
Just use raw git  and a noconflict on jsperf to add your branch and
backbone 1.2. Im on my phone or I'd give an example

gaperton:
Ok, I will do it. Thanks.

megawac:
@gaperton here's an example of how I've done it in the past http://jsperf.com/trigger-with-for-loop-vs-slice/6

Also you may want to fork underscore instead and change `isEqual` for simple values of `a` and `b`

gaperton:
I'm not sure is it ok for underscore or not. For backbone it's ok since we need deep comparison in case if we have objects and arrays only, and we shouldn't care for other types in model's attributes. Because, everything you have in model attributes must be serializable somehow, so requirements are less restrictive here.

gaperton:
Here.
http://jsperf.com/backbone-model-set-performance2

For some reason appears to be 10x faster in my local test. But still, jsperf shows 4x improvement in Chrome, and about 2x improvement in Safari.

In Firefox it's about 25% faster. Firefox is terrifically slow with backbone, I'm going to do something about it too.

megawac:
An underscore patch to `_.isEquals` could be done like

```js
_.isEqual = function(a, b) {
   var typeA = typeof a, typeB= typeof b;
   return typeA != typeB || typeA == 'object' || a === 0 ? eq(a, b) : a === b;
}
```

jridgewell:
With the other contributors, I'd love to see this change made in underscore so everyone can benefit.

gaperton:
I wouldn't merge this isEqual to underscore. undescore's isEqual approach the comparison in much more paranoid way, and it's done intentionally. In contrast, this is optimized application-specific comparison for the special case of model.set changed attributes detection. So in this particular case it's not a problem if it will treat 0 and -0 as equal values (no difference between them in JSON and most type systems anyway), or it wil always fire change on assignment of NaN (it's not serializable either).

The same story about _.clone - there's a special case of cloning in model.set which may and worth to be optimized. Both optimizations has significant impact on Model.set performance, so... Please, decide. For me it doesn't make significant difference; we're not using backbonejs directly, so it's even easier for me to incorporate these optimizations to Backbone.NestedTypes plugin.

megawac:
Did you see the amendment I suggested? Something like that could make it
into underscore

gaperton:
Yes, I did.

( typeA != typeof b ) is obviously enough to decide that values are not equal. Also, we might optimize for nulls. So...

```
var typeA = typeof a;
if( typeA != typeof b ) return false;
if( typeA == 'object' ) return a && b ? eq( a, b ) : a === b;
return a === 0 || a != a ? eq(a, b) : a === b;
```

Looks good?

gaperton:
Now, look at this :)
http://jsperf.com/underscore-isequal

gaperton:
Ok, created pull request with heavily optimized _.isEqual(). It's freaking fast, 1,5-2 times faster than lodash implementation.
 
https://github.com/jashkenas/underscore/pull/2179

megawac:
> `typeA != typeof b` thats actually not sufficient for cases like `new String('a') == 'a'` ;)

gaperton:
Yes, I have noticed it doesn't pass unit test. That's what I was afraid of.

But if you do something, it's better to do it right. So, take look at refined implementation. Performance is terrific, much better than lodash. And it still pass unit test.

https://github.com/jashkenas/underscore/pull/2180

gaperton:
Ok, I removed isEqual optimization from this pull request. There's only inlined optimized clone function.

Are you ready to merge?

PR #3609, opened 2015-05-15
Document Collection#modelId

akre54:


jridgewell:
LGTM.

jashkenas:
"pollymorphic". Spelling.

Also, that terminology might be unclear to our readers. Let's not assume a CS degree.

akre54:
We use `polymorphic` in two other places, including one right above this from 2013 (df7a6f20). Any suggestions for an alternative?

PR #3613, opened 2015-05-16
Fix opening and closing iframe document in jsdom

ahumphreys87:
Some of our tests were failing in Marionette due to jsDom not returning the `document` on `document.open()`. Also it seems that jsDom does not attach the `frameElement` property to `iframe.contentWindow`. I imagine other using jsDom may see similar issues.

Some small changes here, all tests still pass.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1647,14 +1647,15 @@
       // support the `hashchange` event, HTML5 history, or the user wants
       // `hashChange` but not `pushState`.
       if (!this._hasHashChange && this._wantsHashChange && !this._usePushState) {
-        var iframe = document.createElement('iframe');
-        iframe.src = 'javascript:0';
-        iframe.style.display = 'none';
-        iframe.tabIndex = -1;
+        this.iframeElement = document.createElement('iframe');
+        this.iframeElement.src = 'javascript:0';
+        this.iframeElement.style.display = 'none';
+        this.iframeElement.tabIndex = -1;
         var body = document.body;
         // Using `appendChild` will throw on IE < 9 if the document is not ready.
-        this.iframe = body.insertBefore(iframe, body.firstChild).contentWindow;
-        this.iframe.document.open().close();
+        this.iframe = body.insertBefore(this.iframeElement, body.firstChild).contentWindow;
+        this.iframe.document.open();
+        this.iframe.document.close();
         this.iframe.location.hash = '#' + this.fragment;
       }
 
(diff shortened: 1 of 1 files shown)


backbone.js
```diff
@@ -1647,14 +1647,15 @@
...
       // `hashChange` but not `pushState`.
       if (!this._hasHashChange && this._wantsHashChange && !this._usePushState) {
-        var iframe = document.createElement('iframe');
-        iframe.src = 'javascript:0';
-        iframe.style.display = 'none';
-        iframe.tabIndex = -1;
+        this.iframeElement = document.createElement('iframe');
+        this.iframeElement.src = 'javascript:0';
+        this.iframeElement.style.display = 'none';
+        this.iframeElement.tabIndex = -1;
```
jridgewell: Are these changes still needed after https://github.com/marionettejs/backbone.marionette/pull/2561/files#diff-f173a967c2031784640ae886e7d379e8R35?

ahumphreys87: Yeah unfortunately - jsDOM dies here: https://github.com/jashkenas/backbone/pull/3613/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR1697 as `iframe.contentFrame` is undefined :disappointed: 

The change in marionette is still required as jsDOM was throwing an error when removing an element that ain't in the DOM - so so strange. It could be because its an older version of jsDOM but we can't upgrade as they stopped supporting node.

backbone.js
```diff
@@ -1647,14 +1647,15 @@
...
         var body = document.body;
         // Using `appendChild` will throw on IE < 9 if the document is not ready.
-        this.iframe = body.insertBefore(iframe, body.firstChild).contentWindow;
-        this.iframe.document.open().close();
+        this.iframe = body.insertBefore(this.iframeElement, body.firstChild).contentWindow;
```
jridgewell: Instead of storing both `iframeElement` and `iframe`, how about just using just `iframe` (which would be the element, not the `contentWindow`). We can use a local variable to reference the `contentWindow` when we're `open`/`closing` it.


```js
this.iframe = document.createElement('iframe');
//...
var iWindow = this.iframe.contentWindow;
iWindow.document.open();
iWindow.document.close();
```

ahumphreys87:
@jridgewell updated :smile:

jridgewell:
@braddunbar Mind giving your opinion?

jridgewell:
Merging tomorrow, unless there are objections.

jashkenas:
All the tests still pass in IE7 and IE8? 

If someone can confirm that's the case, then feel free to merge.

jridgewell:
@ahumphreys87: Getting one error in IE6 Test suite IE6

jdalton:
The failing IE6 tests aren't due to this PR. They were failing on the 1.2.0 release. If you all want to support IE6 you might add it to your sauce test run.

jdalton:
I assumed Backbone just dropped support for it which is reasonable.

ahumphreys87:
does this mean this can be merged?

PR #3614, opened 2015-05-16
add storehouse for intentionally stored models

craigmichaelmartin:
It is common for multiple components to need access to the same data. In this case, either a view is conflated so that the same model can be accessed in its different facets, the model is passed around to each view, multiple models are used with logic to keep them in sync, or a global model is used. All of these have deficiencies: non-atomic/componentized views, a fragile hierarchy, state issues, or encapsulation breakdowns. 

This pull request proposes a storehouse into which explicitly decided models are saved, and from which can be explicitly drawn. If a model is created with a custom `cid` passed in, the reference returned will be from the storehouse if one exists, or else to the newly create model (whose reference is saved to the storehouse). In this way, using the `cid` from options, there will be no memory hit for how backbone is being used currently.

As proposed then, a view can be sanely small, atomic, componentized, and encapsulated with a model whether it has the only reference to it, or whether there are dozens more, the view does not know and the creation and experience is as desired. When any view changes the model's state, all views update accordingly. If all but the original model remain, or if dozens of new references emerge, each view has the same desired behavior for its model without knowing or caring if other views have the same model.

There are way more lines here than the pull request, as this is more of adding a new philosophic option than it is code (it weighs in with 5 new lines of code, and 4 changed lines). If there is an optimal way to achieve this, or more tests are desired, please let me know!

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -367,23 +367,30 @@
   // Backbone.Model
   // --------------
 
+  // A storehouse to retrieve models explicitly stored via a custom cid.
+  var modelStore = {};
+
   // Backbone **Models** are the basic data object in the framework --
   // frequently representing a row in a table in a database on your server.
   // A discrete chunk of data and a bunch of useful, related methods for
   // performing computations and transformations on that data.
 
-  // Create a new model with the specified attributes. A client id (`cid`)
-  // is automatically generated and assigned for you.
+  // Create a new model with the specified attributes, or retrieve an existing
+  // model by passing its client id (`cid`). If a new model is created, and no
+  // cid is passed in, one is automatically generated and assigned for you.
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
-    this.cid = _.uniqueId(this.cidPrefix);
-    this.attributes = {};
-    if (options.collection) this.collection = options.collection;
-    if (options.parse) attrs = this.parse(attrs, options) || {};
+    var obj = modelStore.hasOwnProperty(options.cid) ? modelStore[options.cid] : this;
+    obj.cid = options.cid || _.uniqueId(obj.cidPrefix);
+    obj.attributes = obj.attributes || {};
+    if (options.collection) obj.collection = options.collection;
+    if (options.parse) attrs = obj.parse(attrs, options) || {};
+    if (modelStore.hasOwnProperty(options.cid)) return obj.set(attrs, options);
     attrs = _.defaults({}, attrs, _.result(this, 'defaults'));
     this.set(attrs, options);
     this.changed = {};
+    options.cid && (modelStore[options.cid] = this);
     this.initialize.apply(this, arguments);
   };
 
--- a/test/model.js
+++ b/test/model.js
@@ -66,6 +66,66 @@
     equal(model.get('last_name'), 'Unknown');
   });
 
+  test("initialize and retrieve same model", 6, function() {
+    var Model = Backbone.Model.extend({
+      initialize: function() {
+        this.one = 1;
+        equal(this.collection, collection);
+      }
+    });
+    var model = new Model({}, {cid: 'test', collection: collection});
+    equal(model.one, 1);
+    equal(model.collection, collection);
+    var modelSame = new Model({}, {cid: 'test'});
+    equal(modelSame.one, 1);
+    equal(modelSame.collection, collection);
+    equal(model, modelSame);
+  });
+
+  test("initialize and retrieve same model with options", 3, function() {
+    var Model = Backbone.Model.extend({
+      initialize: function(attributes, options) {
+        this.one = options.one;
+      }
+    });
+    var model = new Model({}, {cid: 'test', one: 1});
+    equal(model.one, 1);
+    var modelSame = new Model({}, {cid: 'test'});
+    equal(modelSame.one, 1);
+    equal(model, modelSame);
+  });
+
+  test("initialize and retrieve same model with parsed attributes", 3, function() {
+    var Model = Backbone.Model.extend({
+      parse: function(attrs) {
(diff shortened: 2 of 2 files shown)


megawac:
I'm not a big fan of this on first glance as it seems like it will prevent models from ever being garbage collected. This can also be implemented pretty easily in user lab (though they should use a `WeakMap` instead of an object) by either extending or patching model

jridgewell:
I was just having a conversation with coworkers about PHP version of this, and we decided it was a **big** :-1:. Additionally, I've actually written a plugin for this on an old client project, and it turned out to be a huge source of bugs in our code.

In essence, what does `new Model({})` do? It returns a brand new instance of a model with specified attributes.

```javascript
var model = new Model({ cid: 1 });
var model2 = new Model({ cid: 1 });
model.set('test', 'test');
```

In this example, developers immediately understand that only `model`'s attributes have changed. The fact that `model2` was also updated is completely unintuitive. That isn't normal javascript, and it's going to trip people up.

```javascript
var model = new Model({ cid: 1 });
var model2 = model;
model.set('test', 'test');
```

In this, it's very explicit that `model` and `model2` are the same instance, so it's very clear that modifying one is going to modify the other.

I **strongly** recommend the "the model is passed around to each view" option, because it makes dependencies very clear -- it's dependency injection!

craigmichaelmartin:
Thank you for your feedback. Here are some initial responses.

> I'm not a big fan of this on first glance as it seems like it will prevent models from ever being garbage collected. 

Not all models, but yes, those intentionally stored. Let me think on this.

> This can also be implemented pretty easily in user lab

Agreed, but this is a powerful option that could be included 'out of the box' for a very small footprint.

> In essence, what does new Model({}) do?

It does what you expect. The proposal does not affect all model constructions, just when specifically and intentionally desired by passing in an option to do so. 

> The fact that model2 was also updated is completely unintuitive. 

I don't believe so. It is expected because an option was specifically passed in for this behavior. `cid`s are expected to be unique per model, so it is a perfect fit. If desired though, this could be explicit by calling it something like primaryKey. I think `cid` is a perfect fit, though.

> I strongly recommend the "the model is passed around to each view" option ...

I find this approach works fine in some cases, but in many others creates fragile and non-componentized views which are harder to reuse, and expensive to maintain. When passing around a model, it often begins journeying an hierarchy where many of the intermediate views have no relation to the model.

> ... because it makes dependencies very clear -- it's dependency injection!

I don't think its fair to think necessarily of a model in a view in this pattern. 

In general, this is not a use case for every created model. But it is a powerful option with a small footprint which enables atomic and encapsulated views for when the same model should underlie or surface data for many different scattered views. It has proven fruitful and greatly simplified management of data, flexibility of reuses, and maintainability .

evanstern:
First of all: Hey Craig :-)

This idea is something I've tinkered with before. After a while I came to some of the same conclusions that others here are expressing. Passing a model around explicitly does make for more easily understandable code. I also discovered that in most of the places where I wanted to have this kind of functionality what I was lacking was a well defined hierarchy of views/controllers.

I'd suggest considering why you want to this in the first place. When multiple views require access to the same model you probably have either a nested hierarchy of views or multiple sibling views (which could be thought to be contained by a parent view/controller object). In either case it seems that managing model to view relationships is the responsibility of the parent view. 

If you want to keep a store that tracks and manages model creation then you could create a factory object and make it a dependency of the parent and child views. This works well if you want to allow sibling views/controllers to have the ability to interact with the shared store of models directly while still fully encapsulating the logic in one module. It also has the benefit of being very explicit. 

Garbage collection would be a concern unless you are careful with how you handle the model store. When I've created shared collections like this I've tended to attach them to a parent controller (view) and tie destruction of the collection to the `close` method of that view.

That's just my two cents on this. Feel free to contact me if you'd like to discuss this off-line; maybe we could come up with some sort of plugin for this kind of functionality if we can distill a good specific set of use cases.

xerona:
I'll preface by saying I'm biased on this concept. It would work well anywhere passing models would, but the association in the middle views will no longer be required. Admittedly, I'm not particularly experienced, and so I appreciate any comments on my thoughts.

> I'm not a big fan of this on first glance as it seems like it will prevent models from ever being garbage collected

I can't imagine where garbage collection would ever happen in this process. That should be handled.

> The fact that model2 was also updated is completely unintuitive. That isn't normal javascript, and it's going to trip people up.

It might feel unintuitive if developers are using numerical cids, however sharing models should be deliberate so a meaningful cid would be a better practice.

> Passing a model around explicitly does make for more easily understandable code. I also discovered that in most of the places where I wanted to have this kind of functionality what I was lacking was a well defined hierarchy of views/controllers.

In my experience, a component could be used in two different places, but it may then need to pass two different models through hurting readability. Using a storehouse may also hurt readability if you're making certain assumptions, but it permits highly decoupled components.

jridgewell:
> Using a storehouse ... permits highly decoupled components.

Those aren't decoupled components, they're now dependent on global state.

xerona:
> Those aren't decoupled components, they're now dependent on global state.

That's interesting. Where is the dependency? It shouldn't create any more dependency than passing a model.

craigmichaelmartin:
> Those aren't decoupled components, they're now dependent on global state.

I think that gets to the heart of it (for better or worse for my proposal): the components are not *dependent* of this global state. This application state is not a dependency of the view, but is simply application state. If state is gained from a retrieval vs a creation, that is fine and well. If it is not, that is fine and well. There may be multiple references to the model, and the new model not actually new but retrieved, and this does not matter. In this manner, same as in the 'passing around a model' manner, the view does not know/care whether or no it has a unique reference or not.

craigmichaelmartin:
I was looking at `Symbol.for` and just remembered this PR. What a terrifying idea to overload the constructor! Thanks for the kind words of warning.

Issue #3611, opened 2015-05-15
listenTo change in behavior, 1.2.0?


bazineta:
Using the 1.1.2 release, listenTo() was able to listen for events from objects that provide the typical on() interface; in our case, an Ace editor session, which emits 'change' events, e.g., EditSession.on('change', function (e)).  Under 1.1.2, we've used listenTo() to listen for these events from the editor session.

Using the 1.2.0 release, listenTo() does not appear to be able to listen for these events; they're emitted, and the editor session is present in the View's _listeningTo, but the callback is no longer fired.

We were using listenTo() for automatic cleanup of the listener function.  Curious if this was never supposed to work in this manner, or if it's a bug.  The changes to listening in 1.2.0 are quite extensive, so it's a bit difficult to see what might be the key difference for one unfamiliar with them.

akre54:
Mind writing up a small test case so we can take a look? I'd be interested to see what your code looks like.

jridgewell:
```javascript
// Test Case
var events = {};
var obj = {
  // Any non Backbone.Events event implementation will work...
  on(name, callback) {
    events[name] = callback;
  },
  off(name) {
    events[name] = null;
  }
};

var view = new Backbone.View();
view.listenTo(obj, 'test', console.log);
view.stopListening();
```

This is a side-effect of fixing https://github.com/jashkenas/backbone/pull/3455#issuecomment-71677970. Since now both the listener and the listenee should be able to clean up memory in the other object, I had to share private state between the two objects. [`internalOn`](https://github.com/jashkenas/backbone/blob/dba0faad09d40818fb85dcf84c1e7b97543d3108/backbone.js#L121) is used so that this private data is never leaked outside of Backbone. But it also means that `#listenTo` uses only internal Backbone code instead of the public `#on`.

I'm able to [fix](https://github.com/jashkenas/backbone/compare/master...jridgewell:private-listening-public-on) this by sharing the state privately inside Backbone instead of through a private function. I'll have to look into the performance impact.

akre54:
Ah. I'd say it's not in our contract to support other events systems, but if we can make it work we probably should. Is there any way other than a singleton event @jridgewell? This seems pretty hacky and brittle.

jridgewell:
Reverting https://github.com/jashkenas/backbone/pull/3594 should also do the trick, since the listener will keep track of the events.

akre54:
What did #3455 and #3594 get us? Kind of hard to follow those discussions.

[...]

PR #3594, opened 2015-05-02
Refine eventsApi

jridgewell:
Based off code from https://github.com/jashkenas/backbone/pull/3476

These are the last changes (hopefully!) from the [eventsApi refactor](https://github.com/jashkenas/backbone/pull/3455).

PR #3615, opened 2015-05-18
Use the public `on` method when listening

jridgewell:
Fixes https://github.com/jashkenas/backbone/issues/3611.

This uses a private `listening` var to share state between a Backbone
"listener" and "listenee", instead of using a private `internalOn()` to
share state. This allows `#listenTo` to use the public `#on` method and
keeps interop between Backbone and any other event library.

http://jsperf.com/internal-listening-public-on

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -127,7 +127,7 @@
       listeners[_listening.id] = _listening;
       // Allow the listening to use a counter, instead of tracking
       // callbacks for library interop
-      _listening.backbone = true;
+      _listening.interop = false;
     }
 
     return this;
@@ -154,7 +154,7 @@
 
     if (error) throw error;
     // If the target obj is not Backbone.Events, track events manually.
-    if (!listening.backbone) listening.on(name, callback);
+    if (listening.interop) listening.on(name, callback);
 
     return this;
   };
@@ -210,7 +210,7 @@
       if (!listening) break;
 
       listening.obj.off(name, callback, this);
-      if (!listening.backbone) listening.off(name, callback);
+      if (listening.interop) listening.off(name, callback);
     }
     return this;
   };
@@ -343,7 +343,7 @@
     this.id = listener._listenId;
     this.listener = listener;
     this.obj = obj;
-    this.backbone = false;
+    this.interop = true;
     this.count = 0;
     this._events = void 0;
   };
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -149,10 +146,20 @@
     }
 
     // Bind callbacks on obj, and keep track of them on listening.
-    internalOn(obj, name, callback, this, listening);
+    var error = tryCatchOn(obj, name, callback, this);
+    listening = void 0;
```
megawac: Why is this necessary - won't it be gced anyway

jridgewell: It's [now](https://github.com/jashkenas/backbone/pull/3615/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR91) a private global.

jridgewell:
Hold off on this for today, it has an issue because the other library won't update `listening.count`, so we don't know when we can cleanup references to `listenee` from the `listener`.

jashkenas:
"The other library" ?

Also, `tryCatchOn` isn't a good thing.

jridgewell:
I've just revised the implementation to use a hybrid approach between #3455 and #3594.

If the listenee is using Backbone's `#on`, the `listening` will keep track of callbacks using a fast counter (#3594's approach). If not, then callbacks will be stored in an `_events` object on the `listening` object (#3455). This allows full interoperability while also properly cleaning up memory bindings when all callbacks have been removed.

[JsPerf](http://jsperf.com/internal-listening-public-on/4) tends to like the implementation, except Chrome doesn't like `#listenTo` (running the same benchmark in node v0.12 shows a less than 0.2% slowdown).

- - -

> Also, tryCatchOn isn't a good thing.

Can you explain?

DomBlack:
Is there any chance this PR will be merged?

alexbaumgertner:
@DomBlack 👍  
I'm interested in this code too. 
This PR can fix at least one painful bug in my project. 
Do the author need any help?

alexbaumgertner:
@jridgewell I tried to rebase and fix merge errors in https://github.com/jashkenas/backbone/pull/4043 but a bunch of tests fails. Could you help me?

jbboehr:
Any updates on this? I've merged master into it in #4131, if that helps at all.

PR #3627, opened 2015-05-23
Model#set style

jridgewell:
Declares variables when they're needed, instead of all at the top.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -444,10 +444,10 @@
     // the core primitive operation of a model, updating the data and notifying
     // anyone who needs to know about the change in state. The heart of the beast.
     set: function(key, val, options) {
-      var attr, attrs, unset, changes, silent, changing, prev, current;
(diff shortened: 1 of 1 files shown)


akre54:
I think this is a style thing. In general it's considered a good practice to hoist your var declarations to the top of the current scope, to prevent re-initialization and to give you a sense of the variables the function uses.

jridgewell:
Definitely just a style change, but one that's used almost everywhere else in Backbone.

akre54:
As a counterpoint, CoffeeScript does this same hoisting. I imagine this is one of those Backbone things that has been around since time eternal.

braddunbar:
I'm a big :+1: on this one. Declaring a variable along with its first assignment gives you a clue that this is the first time this variable is needed. In a method that's nearly 100 lines long, this is good information to have and helps to organize and possibly split apart.

PR #3629, opened 2015-05-23
Model#save style

jridgewell:
- Makes it apparent what code needs temporarily `wait`ed attributes
- Uses extend to default `parse` option
- Removes an unneeded object check (just checking for false-y does it)
- Prevents mutating passed in `attrs`
- Declares variables when they are needed, instead of all at the top

(That's the last of the `Model` style updates, I promise. :wink:)

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -590,7 +590,7 @@
         (attrs = {})[key] = val;
       }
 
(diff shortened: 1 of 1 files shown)


jashkenas:
> (That's the last of the Model style updates, I promise. )

Aw, man.

PR #3630, opened 2015-05-24
Test View

jridgewell:
- Adds some missing `View` tests.
- Returns view from `#delegate` and `#undelegate`


jashkenas:
Groovy.

jashkenas:
(Just as a fun side-note, for all the SemVer weenies out there, this is a **BREAKING CHANGE**.)

jridgewell:
>  this is a BREAKING CHANGE.

The return value? I'd consider this just a minor release (now returning something), not a major (changing what we were returning).

jashkenas:
That was tongue in cheek. It's a change in behavior. No exceptions. No mercy.

PR #3631, opened 2015-05-24
Grab id using #get in Model#url

jridgewell:
Also, use the same slash normalizing as Router and History


braddunbar:
Very nice. :+1:

PR #3632, opened 2015-05-24
View style

jridgewell:


braddunbar:
Looks good to me. :+1:

backbone.js
```diff
@@ -1170,7 +1170,6 @@
   // if an existing element is not provided...
   var View = Backbone.View = function(options) {
     this.cid = _.uniqueId('view');
-    options || (options = {});
```
akre54: this is cool because we're `>= 1.7.0` which made a change to `_.pick` for undefined arguments right?

jridgewell: Exactly. Sorry, I put it in the [commit message](https://github.com/jridgewell/backbone/commit/8266bd4609cf3faddf70dde671ced2a3dd434d75) and not the PR. :flushed:

PR #3633, opened 2015-05-24
Do not overwrite this.routes in #_bindRoutes

jridgewell:


backbone.js
```diff
@@ -1476,11 +1476,9 @@
     // order of the routes here to support behavior where the most general
     // routes can be defined at the bottom of the route map.
     _bindRoutes: function() {
-      if (!this.routes) return;
-      this.routes = _.result(this, 'routes');
-      var route, routes = _.keys(this.routes);
-      while ((route = routes.pop()) != null) {
-        this.route(route, this.routes[route]);
+      var routes = _.result(this, 'routes');
```
megawac: what happens if `routes` is undefined. I would highly suggest using `_.each` here TBH as it will support undefined

PR #3635, opened 2015-05-26
Fix for history on Opera 12 #3634

### LINKED ISSUE #3634: `router.navigate` doesn't update history after an async request (Opera 12)
This is a singular question, but please bear with me. I faced a strange issue when executing a Backbone app on a rather ancient browser (Opera 9.8) some of the routes were not being registered in the browser history. With a little research saw that this would only happen in the following scenario:
- user interaction
- ajax request
- ajax callback handler that executes a `router.navigate`

In this cases **the url did change**, the app **could navigate to that route** but it **doesn't update the browser history**, hence the user can't navigate back.

This had the smell of a browser bug, created an small demo and bingo, this happens with just updating `document.location.hash` on a ajax request handler, here is [the SO question](http://stackoverflow.com/questions/30457126/change-history-after-async-ajax-request-on-opera-9-8) with that situation.

Looking for a fix for this I found that if I use `document.href` instead of `document.location.hash` on the [Backbone.history._updateHash](https://github.com/jashkenas/backbone/blob/61d18408addb54f4af4d3e14b0d47111ef94de99/backbone.js#L1796-L1804) method it works:
```javascript
Backbone.history._updateHash = function(location, fragment, replace) {
  var href = location.href.replace(/(javascript:|#).*$/, '');
  var url = href + "#" + fragment;
  if (replace) {
    location.replace(url); 
  } else {
    location.href = url;
  }
};
```
App testing are green and looks like my problem it's solved, but I've some question:

1. Is this a new issue? Although I couldn't find anything regarding this it's strange that no one experience this before
2. Is it safe to override a Backbone method the way I did? I see that preceding `_` on the method name so I assume that this wasn't designed to override.
3. Should this be fixed in Backbone instead? If that is the case I could do a PR.
4. Is there any alternative workaround that you can think of?

Thank you

am:
This PR fixes #3634

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1794,12 +1794,12 @@
     // Update the hash location, either replacing the current entry, or adding
     // a new one to the browser history.
     _updateHash: function(location, fragment, replace) {
+      var href = location.href.replace(/(javascript:|#).*$/, '');
+      var url = href + "#" + fragment;
       if (replace) {
-        var href = location.href.replace(/(javascript:|#).*$/, '');
-        location.replace(href + '#' + fragment);
+        location.replace(url);
       } else {
-        // Some browsers require that `hash` contains a leading #.
-        location.hash = '#' + fragment;
+        location.href = url;
       }
     }
 


am:
Closing this since [recently realise](https://github.com/am/backbone/commit/7e744954812d9c26e6d6daffc84363c32b496c32) that this problem only happens when using jQuery 2. Unit test where passing using Opera even without the fix, so I'll leave this weirdness be solved by downgrading to jQuery 1.x.x. In any case if someone needs to support Opera 12 and depends on jQuery 2 then this fix would be the way to go.

Issue #2824, opened 2013-10-22
Collection.create() with wait: true triggers Collection.parse()


chris-h-sg:
Since version 1.1, calling

``` javascript
myCollection.create( {} , { wait: true } );
```

triggers a call to `myCollection.parse()` once the server's response is returned. According to the documentation, it looks like parse should only be called in reaction to `fetch()`.

Additionally, the parameters to this `parse()` call seem to be different: When calling `parse()` via `fetch()`, the first parameter contains the raw data of the response, i.e. an array of JS objects. When `parse()` is triggered from `create()`, the first parameter contains the newly created model.

As a workaround, it's possible to set `parse: false` when calling `create()`, this will prevent the call to `parse()`.

PR #3636, opened 2015-05-26
Create with wait:true should not call collection.parse

w0rm:
This seems to be a regression of #2824 

`collection.create` calls `model.save` that sets `options.parse = true` and passes these options downwards `success` callback -> `collection.add` -> `collection.set` which then tries to call `collection.parse` with model as the first argument.

I've added a test case and an ugly fix (not sure if it can be done in a prettier way).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -614,10 +614,7 @@
 
       // After a successful server-side save, the client is (optionally)
       // updated with the server-side state.
-      if (options.parse === void 0) {
-        options.parse = true;
-        options.enabledParseForModel = true;
-      }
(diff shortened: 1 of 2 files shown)


w0rm:
As a temporary fix of this bug, I added this in `Collection.parse`:

```javascript
parse: function (data, options) {
   if (data instanceof Model) { 
       return data
   }
   // return parsed data
}
```

jridgewell:
Seems this broke with https://github.com/jashkenas/backbone/pull/3488.

PR #3637, opened 2015-05-27
Return removed models from #_removeModels

jridgewell:
Instead of returning a boolean, just return the removed models (or false if nothing is removed).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1098,6 +1098,9 @@
 
     // Internal method to sever a model's ties to a collection.
     _removeReference: function(model, options) {
+      delete this._byId[model.cid];
+      var id = this.modelId(model.attributes);
+      if (id != null) delete this._byId[id] = model;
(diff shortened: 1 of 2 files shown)


akre54:
PhantomJS tests are failing but otherwise nice work!

jridgewell:
Gah, I forgot to add local changes.

jridgewell:
I just can't seem to grok git today... https://github.com/jashkenas/backbone/pull/3639 fixes a syntax error.

jridgewell:
Also, this _is_ a bug fix, highlighted by the [changed tests](https://github.com/jridgewell/backbone/commit/61e765087d955266225dc2eefad26122f1af9975). #3512 incorrectly used `Array#slice` inside `#_removeModels`, which doesn't mutate the `models` variable in `#remove`.

jashkenas:
Ah.

Issue #1407, opened 2012-06-13
Collection.parse should be called in .add if options.parse is true


davidmarble:
In the constructor of Collection (currently line 554 - https://github.com/documentcloud/backbone/blob/master/backbone.js#L554), the below line indicates that the `reset` method, or `add` method called by `reset`, should check `options.parse`, and if if true, call `parse` on the models passed in.

```
if (models) this.reset(models, {silent: true, parse: options.parse});
```

However, there is no such logic in the `add` or `reset` methods of Collection. Seems to be an oversight.

PR #3644, opened 2015-05-30
Collection parse

jridgewell:
Fixes #3636, #2824.

Reverts #1407, so **breaking change**.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -739,6 +739,7 @@
     if (options.comparator !== void 0) this.comparator = options.comparator;
     this._reset();
     this.initialize.apply(this, arguments);
+    if (models && options.parse) models = this.parse(models, options);
     if (models) this.reset(models, _.extend({silent: true}, options));
   };
 
@@ -789,7 +790,6 @@
     // the core operation for updating the data contained by the collection.
     set: function(models, options) {
       options = _.defaults({}, options, setOptions);
-      if (options.parse) models = this.parse(models, options);
       var singular = !_.isArray(models);
       models = singular ? (models ? [models] : []) : models.slice();
       var id, model, attrs, existing, sort;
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -983,13 +983,13 @@
     // collection when they arrive. If `reset: true` is passed, the response
     // data will be passed through the `reset` method instead of `set`.
     fetch: function(options) {
-      options = options ? _.clone(options) : {};
-      if (options.parse === void 0) options.parse = true;
-      var success = options.success;
+      options = _.extend({parse: true}, options);
```
akre54: `_.defaults`?

jridgewell: No, because they can specify `{ parse: false }` if they want to. And, we don't want to mutate, their `options` object (that, and `_.defaults(null, {parse: true})` will return `null`).

akre54: cool, thanks

jashkenas:
Mind explaining the behavior of this — I don't want to guess incorrectly.

Collections should be parsing whenever data comes down from the server ... but if you set data directly on the client side, it shouldn't parse by default. Right?

jridgewell:
> Collections should be parsing whenever data comes down from the server ... but if you set data directly on the client side, it shouldn't parse by default. Right?

Yup.

jridgewell:
Wait, are we going to releasing this with v1.2.1? I thought the [consensus](https://github.com/jashkenas/backbone/pull/3651#issuecomment-107257566) was this would be released in v2?

jashkenas:
This looks like a bugfix, doesn't it? We shouldn't be parsing your models with data set on the client-side.

jridgewell:
> This looks like a bugfix, doesn't it?

Yah, but it's fixing it by changing core functionality. I think it'd be better to label it as a known bug (with an [easy patch](https://github.com/jashkenas/backbone/pull/3636#issuecomment-106603978)) that's fixed in the next major.

jridgewell:
We could actually pull the `data instanceof Model` check into v1.2.1, and leave this for v2:

```patch
# L795
      options = _.defaults({}, options, setOptions);
-     if (options.parse) models = this.parse(models, options);
+     if (options.parse && !this._isModel(models)) models = this.parse(models, options);

jashkenas:
Who do you imagine (What use case do you imagine —) that we'd be breaking by simply including this fix now?

jridgewell:
Well, my job's codebase uses it extensively (not that we'll be updating anytime soon...).

I understand the rationale that people want to parse their collection when setting (I just think it's wrong), so I do think this will be a pretty major change for users. _Espescially_ in a patch release.

jashkenas:
Okay, fine ... but it makes me sad that parsing-while-set'ing landed in first place.

jridgewell:
https://github.com/jashkenas/backbone/commit/daf6892d22d26148aacf00b103d6c308abef8e4d#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR804 :stuck_out_tongue:

Don't worry, I'll open another PR so we can remove it in v2.

PR #3649, opened 2015-05-31
v2: Collection#remove immutable models

### LINKED ISSUE #3534: model.destroy() breaks collection.each
```JS

  test("destroy models in each loop", 4, function() {
    var collection = new Backbone.Collection([{id: 1}, {id: 2}, {id: 3}, {id: 4}]);
    collection.each(function(model) {
      notStrictEqual(model, undefined);
      if (model !== undefined) {
        model.destroy({url: function() {}});
      }
    });
  });
```

That test breaks because `model.destroy` fires `destroy` which causes the collection to call `this.remove(model)` which then splices out the array and causes the `each` to break. This technically could be considered a bug in lodash/underscore (@jdalton ?) but released versions of those cause this so we might want to fix it in backbone.

Deferring the `coll.remove` seems like a bad idea so maybe we should just slice the models into a new array when calling `_.each` and `_.forEach`. We could also make `model.destroy` just do all the destroying in a defer?

Let me know what you guys think.

jridgewell:
- Make `#models` immutable (follows #3648)
- Uses a O(k+n) algorithm instead of a O(k*lg(n)) algorithm

The trigger loop could be combined with the one above, but I wanted to
mirror `Collection#add` where `#length` and `#models` already reflect
all of the added models. This way, the 'remove' event already reflects the
removed models.

Fixes #3534.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -797,7 +797,7 @@
     remove: function(models, options) {
       options = _.extend({}, options);
       var singular = !_.isArray(models);
-      models = singular ? [models] : _.clone(models);
+      models = singular ? [models] : models;
       var removed = this._removeModels(models, options);
       if (!options.silent && removed) this.trigger('update', this, options);
       return singular ? removed[0] : removed;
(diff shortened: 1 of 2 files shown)


paulfalgout:
I ran into this today.

We have a collection where each model has a status.  Today we added the ability to batch change the status.  Under some situations the status change will trigger some models to be removed.  Using `myCollection.invoke('set', 'status', 'closed');` was errorring out as the status change was mutating the original collection.

jgonggrijp:
Closing for the same reason as #3534.

PR #3651, opened 2015-05-31
Backbone.Promise

jridgewell:
Competes with #2489.

Specifically, it implements `Backbone.Promise` instead of `Backbone.Deferred`, so it can be easily swapped with any ES6 compatible Promise library.

This is phase 1. Phase 2 would be:

- returning a `Backbone.Promise` from `Backbone.sync`.
- removing all the `options.success` and `options.error` callbacks in favor of chained promises.
  - That includes transitioning from `success` callbacks to rejected promises, ie. `Model#fetch` where validation fails with the server's attributes.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1415,21 +1415,22 @@
     throw new Error('Backbone does not provide a spec compliant Promise by default.');
   };
 
-  // A helper method that forces jQuery's first `then` callback to be
-  // executed asynchronously.
-  // This is used so we can guarantee our async return values execute
-  // callbacks async, not async sometimes and sync other times.
-  var asyncDeferred = function(method) {
-    return function(value) {
-      var deferred = Backbone.$.Deferred();
-      _.defer(deferred[method], value);
-      return deferred.promise();
-    };
-  };
-
   _.extend(Backbone.Promise, {
-    resolve: asyncDeferred('resolve'),
-    reject: asyncDeferred('reject')
+    // A wrapper around jQuery's normal resolve to force it to adopt a
+    // thenable's state, and execute then callbacks asynchronously.
+    resolve: function(value) {
+      var deferred = Backbone.$.Deferred();
+      _.defer(deferred.resolve);
+      return deferred.promise().then(_.constant(value));
+    },
+
+    // A wrapper around jQuery's normal reject to force it to execute
+    // then callbacks asynchronously.
+    reject: function(reason) {
+      var deferred = Backbone.$.Deferred();
+      _.defer(deferred.reject, reason);
+      return deferred.promise();
+    }
   });
 
   // Backbone.Router
--- a/backbone.js
+++ b/backbone.js
@@ -601,9 +601,9 @@
       // `set(attr).save(null, opts)` with validation. Otherwise, check if
       // the model will be valid when the attributes, if any, are set.
       if (attrs && !wait) {
-        if (!this.set(attrs, options)) return false;
+        if (!this.set(attrs, options)) return Backbone.Promise.reject(this.validationError);
       } else {
-        if (!this._validate(attrs, options)) return false;
+        if (!this._validate(attrs, options)) return Backbone.Promise.reject(this.validationError);
       }
 
       // After a successful server-side save, the client is (optionally)
@@ -639,7 +639,7 @@
     // Optimistically removes the model from its collection, if it has one.
     // If `wait: true` is passed, waits for the server to respond before removal.
     destroy: function(options) {
-      options = options ? _.clone(options) : {};
+      options = _.extend({}, options);
       var model = this;
       var success = options.success;
       var wait = options.wait;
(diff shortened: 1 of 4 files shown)


tgriesser:
For any "major breaking change" PR's such as this #3649, #3644 or others where existing tests are being significantly altered/added it might be wise to create / PR against a 2.0 branch so it doesn't put master in a state where patch releases can't be made against the current master 1.x master.

Thoughts? /cc @braddunbar @jashkenas @akre54

megawac:
Agreed, and :+1: to adding `BB.Promise`

braddunbar:
That sounds good to me. :+1:

backbone.js
```diff
@@ -1408,6 +1408,31 @@
     return Backbone.$.ajax.apply(Backbone.$, arguments);
   };
 
+  // A psuedo Promise implementation used to ensure asynchronous methods
+  // return thenables.
+  // Override this if you'd like to use a different ES6 library.
+  Backbone.Promise = function() {
```
megawac: I'd say a promise implementation should be included by default (I would suggest [`npo`](https://github.com/getify/native-promise-only))

braddunbar: I think I disagree (feel free to convince me). This seems very similar to polyfilling `JSON`, which we do not provide a default implementation of.

megawac: The difference is BB doesn't use JSON directly. Though I'd be fine with just assuming `Promise` is defined globally

jridgewell: I'm also a little wishy-washy about depending on jQuery's Deferreds. They're badly non-standard, and we're now demanding a dependency on jQuery (or some other promise library) for Models and Collections to work.

We could go our "jQuery require" route, attempting to require some library `'promise'`, then feed that into the factory function. Or just require overwriting `BB.Promise` on node if you want to use async methods. Hmmm.

tgriesser: > We could go our "jQuery require" route, attempting to require some library 'promise', then feed that into the factory function. Or just require overwriting BB.Promise on node if you want to use async methods. Hmmm.

As of ES6, we can just expect `Promise` it to be a builtin global (and it is already in [many environments](http://caniuse.com/#search=promise)) so it'd be exactly as Brad describes, the same as polyfilling JSON. The only requirement here would be that BB wouldn't depend on any specific non-standard library specific methods/helpers.

akre54: How about a `if (typeof Promise != "undefined") Backbone.Promise = Promise`
line right after our pollyfill using deferreds? You should probably be
pollyfilling Promise anyways if you need it.

jridgewell: We could do a fallback system:

```js
if (root.Promise) {
  BB.Promise = root.Promise;
} else if (Backbone.$) {
  BB.Promise = jQueryDeferredWrapper;
} else {
  BB.Promise = GiveHelpfulErrorMessageAKAYouAintGotNoPromises; // :wink:
}
```

megawac: -1 for jquery fallback, the implementation differs to greatly from the
promise spec

akre54: @megawac but does it fail for our use case here? Backbone and Underscore tend to opt for the ponyfill approach over pollyfills, encapsulating the desired behavior in functions instead of shimming support globally.

If we force a Promise library (or a Promise global), it'll be confusing to developers why this fails in older browsers but works in newer ones. On the other hand if we set the precedent here that we're ok with making users pollyfill older behaviors, we can start to [follow the lead of React](https://facebook.github.io/react/docs/working-with-the-browser.html#browser-support-and-polyfills) in requiring developers to pollyfill native methods (map, forEach, etc).

Cleod9:
Referencing this ticket: https://github.com/jashkenas/backbone/issues/3739

I just wanted to bring the above to everyone's attention since if Promises are implemented some extra care should probably be taken to make sure that certain `error()` arguments I mentioned are not lost somewhere in the promise. Otherwise I could anticipate some scary situations where devs are confused on the reason of an ajax failure when no relevant detail is provided by the Promise's error handler.

jeromecovington:
Allowing developers to handle their own polyfills is desirable in general and more forward-thinking, provided the specific polyfills that meet BB's requirements are clearly documented as such.

jeromecovington:
+1 @Cleod9 re: error handling. Inadvertently handling errors with a `.catch()` without re-throwing up the Promise chain can be a huge source of debugging confusion.

godlark:
I think if we want to return a promise it would be good to pass parsed response to then's callback. Read more at https://github.com/jashkenas/backbone/pull/3779

marlun:
Have there been any further development on this?

mjeanroy:
Hi, it seems that Backbone plans to use promise as the return value of the `sync` method in a futur major release and it's a good news :+1: 

However, what about XHR abortion ? Currently, according to the official spec, it's not possible to cancel a promise, but aborting an ajax request is an important feature I guess. How will you handle this case ?

akre54:
We can either wrap the promise for you, and attach an `abort` method, or you can do that yourself in `Backbone.ajax`. I'm inclined towards the latter, unless there's enough of a reason to add it to core.

PR #3657, opened 2015-06-02
Fix failing IE6 test

jridgewell:
It was a false positive. Our “location” parsing uses an A tag, which doesn’t always work the way the real location parsing does.


jridgewell:
@jashkenas: I think after this, we're good to ship a v1.2.1.

braddunbar:
Fantastic. Thanks @jridgewell!

PR #3658, opened 2015-06-02
Add jquery dependency to package.json

gmaclennan:
The new `require('jquery')` is great for using browserify, but it doesn't work because jquery is not listed as a dependency in `package.json`.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,6 +5,7 @@
   "keywords"      : ["model", "view", "controller", "router", "server", "client", "browser"],
   "author"        : "Jeremy Ashkenas <jeremy@documentcloud.org>",
   "dependencies"  : {
+    "jquery"      : "^2.1.4",
     "underscore"  : ">=1.7.0"
   },
   "devDependencies": {


jashkenas:
Comments from those who know about this?

package.json
```diff
@@ -5,6 +5,7 @@
   "keywords"      : ["model", "view", "controller", "router", "server", "client", "browser"],
   "author"        : "Jeremy Ashkenas <jeremy@documentcloud.org>",
   "dependencies"  : {
+    "jquery"      : "^2.1.4",
```
megawac: The two big issues are we don't know if the user wants `^1.x` or `^2.x` and we do not want to force the user to download jquery if they are using BB on the server

See #3421

megawac:
Correct me if I'm wrong but if you add `jquery: ^2.0, backbone: ^1.2` in your projects `package.json` the import should work

akre54:
Eh? If your main app depends on jQuery, browserify should [pick it up with no issues](https://gist.github.com/akre54/d7ebd8893128e4bf9235/5cf29e4fea40c24213714a1ff82191b0a1c6cd89).

gmaclennan:
In my case my app does not depend on jquery, but it does depend on Backbone. I have no `require('jquery')` statements in my code, but if I pull in `require('backbone')` it fails in browserify because it can't find jquery.

In my opinion if the backbone code includes `require('jquery')` then it should be included as a dependency. I mean, it is a dependency, even if it is an optional one.

But I guess @megawac's solution does work. It just seems to go against how dependencies work - should I be declaring dependencies of sub-modules in my app?

gmaclennan:
Ok, seeing the discussion in #3421 now. I guess that makes the most sense for now.

akre54:
If you're not using jQuery in your app, you have to [tell Browserify to exclude it](https://github.com/substack/node-browserify#bexcludefile). [See the wiki](https://github.com/jashkenas/backbone/wiki/Using-Backbone-without-jQuery#without-jquery).

Sorry if any confusion. Is there a way we can make this clearer to folks?

PR #3659, opened 2015-06-02
Collection#add shouldn't parse a model instance

jridgewell:
Fixes #3636, but in a backwards compatible way.

Reverts #3644, so that it can be added to v2 release.


jridgewell:
The diff is a bit wonky because of the revert, here's what it looks like if #3644 were never merged: https://github.com/jashkenas/backbone/compare/5ea158506ff65def13b132b30f0b0c9ecb0d31a2...jridgewell:collection-parse-v1-2-1

Issue #3653, opened 2015-06-01
Router/History v2


jamesplease:
Backbone's router is inspired by Ye Olde Routers from several years ago, before single-page applications were a 'thing,' and when one usually reached for a handful of jQuery plugins to build a JavaScript app.

These routers are distinguished by three characteristics:
1. Each route is independent from all other routes
2. Each route is a callback that is triggered when the associated URL is matched
3. The router is used to bootstrap the page, and is never used again

It's pretty clear that Backbone's router exhibits the first two characteristics. Over time, changes to the Router have made number 3 a li'l less obvious. The existence of the `trigger` option and the fact that it is `false` by default are evidence of that third characteristic. Numerous blog posts and books always warn _against_ setting `trigger: true`. Additionally, @jashkenas expressed to me that this is how he believed the router should be used when we chatted about it at BBConf a few months ago.

Every other popular JavaScript library/framework has moved away from this pattern. In its place, a style of routing pioneered by Ember has become the de facto way to structure single page apps. Ember's router went on to heavily influence the React router, Angular's UI Router, and Angular v2's router.

While reaching feature parity with Ember's router might be a bit much for Backbone (it would prob. increase the size of the library by 2x lol), I think that looking there for inspiration and coming up with a minimal version of that sort of routing system would tremendously improve Backbone. And I think it could be done in a pretty small amount of code!

Further, by providing the right hooks, it would make it easier for third party developers to create an Ember-style router for Backbone as a third party library, which is quite the challenge right now (there's more on that at the end of this post).

These are what I've gathered to be a few of the important characteristics of those routers:
1. Routes are **states**, and the router is a **state machine**.
2. Transitions between states are **asynchronous**.
3. **State** can optionally be encoded into a **URL** when they are transitioned into.
4. States can have **children states** (this is where a large part of the complexity of Ember-style routing comes from)
5. Routes, at the minimum, have **two callbacks** associated with them: an **asynchronous** callback and a **synchronous** callback.
6. A route is activated **each and every time** the URL changes.
7. **History** is a read/write interface to the browser history, and has no other roles
# 

FAQ for some of those points:

**Why two hooks?**

There are generally two things you want to do when the URL changes:
1. Fetch data (maybe)
2. Render a view (also maybe – but most of the time)

Generally, fetching data is asynchronous, and showing some HTML is synchronous. So, two hooks.

[...]

PR #3660, opened 2015-06-03
Router/History update

jamesplease:
This is a proof of concept of some of the changes discussed in https://github.com/jashkenas/backbone/issues/3653.

There are two types of changes here. The first is providing more hooks for users who may wish to change the behavior of the Router. @jridgewell described them well, I think, when he said that they are similar to the hooks provided in Views. The other type of change is more clearly separating the role of the Router and History. History now does one thing: interfaces with whatever browser APIs are necessary to update the URL. The Router handles everything that relates to the 'route handlers.'

There is more work to do, o' course. I haven't changed the unit tests, for instance.

The API exposed to the user _should_ be mostly backwards compatible. An repo with an example [is here](https://github.com/jmeas/backbone.routing-v2).
# 

The biggest backwards incompatible change so far is the removal of the `trigger` option. It has been replaced with a `silent` option, which effectively does the same thing. It's also `true` by default, which means that route callbacks won't be executed when history updates the URL.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1492,6 +1492,8 @@
 
     // Match a fragment with a registered handler
     matchFragment: function(fragment) {
+      // If the root doesn't match, no routes can match either.
+      if (!Backbone.history.matchRoot()) return false;
       return _.find(Router.handlers, function(handler) {
         return handler.route.test(fragment);
       });
--- a/backbone.js
+++ b/backbone.js
@@ -1473,8 +1473,6 @@
 
     // Simple proxy to `Backbone.history` to save a fragment into the history.
     navigate: function(fragment, options) {
-      if (options === void 0) options = {silent: true};
-      if (options === false) options = {silent: false};
       Backbone.history.navigate(fragment, options);
       return this;
     },
@@ -1759,7 +1757,8 @@
     // you wish to modify the current URL without adding an entry to the history.
     navigate: function(fragment, options) {
       if (!History.started) return false;
-      if (!options || options === true) options = {trigger: !!options};
+      if (options === void 0) options = {silent: true};
+      if (options === false) options = {silent: false};
 
       // Normalize the fragment.
       fragment = this.getFragment(fragment || '');
--- a/backbone.js
+++ b/backbone.js
@@ -1417,8 +1417,8 @@
     options || (options = {});
     if (options.routes) this.routes = options.routes;
     this.handlers = [];
-    this._bindRoutes();
     this.bindNavigationListener();
+    this._bindRoutes();
     this.initialize.apply(this, arguments);
   };
 
@@ -1459,7 +1459,7 @@
       this.handlers.push(handler);
     },
 
-     // Return a handler object from a route, name, and callback
+    // Return a handler object from a route, name, and callback
     getHandler: function(route, name, callback) {
       if (!_.isRegExp(route)) route = this._routeToRegExp(route);
       if (_.isFunction(name)) {
@@ -1467,12 +1467,17 @@
         name = '';
       }
       if (!callback) callback = this[name];
-
       return {
         route: route,
         callback: callback,
         name: name
       };
+    },
+
+    // Execute a route handler with the provided parameters.  This is an
+    // excellent place to do pre-route setup or post-route cleanup.
+    execute: function(callback, args, name) {
+      if (callback) callback.apply(this, args);
     },
 
     onNavigate: function(fragment) {
@@ -1484,12 +1489,6 @@
         this.trigger('route', matchedRoute.name, args);
       }
       return this;
-    },
-
-    // Execute a route handler with the provided parameters.  This is an
-    // excellent place to do pre-route setup or post-route cleanup.
-    execute: function(callback, args, name) {
-      if (callback) callback.apply(this, args);
     },
 
     // Match a fragment with a registered handler
@@ -1673,15 +1672,16 @@
       // support the `hashchange` event, HTML5 history, or the user wants
       // `hashChange` but not `pushState`.
       if (!this._hasHashChange && this._wantsHashChange && !this._usePushState) {
-        var iframe = document.createElement('iframe');
-        iframe.src = 'javascript:0';
-        iframe.style.display = 'none';
-        iframe.tabIndex = -1;
+        this.iframe = document.createElement('iframe');
+        this.iframe.src = 'javascript:0';
+        this.iframe.style.display = 'none';
+        this.iframe.tabIndex = -1;
         var body = document.body;
         // Using `appendChild` will throw on IE < 9 if the document is not ready.
-        this.iframe = body.insertBefore(iframe, body.firstChild).contentWindow;
-        this.iframe.document.open().close();
-        this.iframe.location.hash = '#' + this.fragment;
+        var iWindow = body.insertBefore(this.iframe, body.firstChild).contentWindow;
+        iWindow.document.open();
+        iWindow.document.close();
+        iWindow.location.hash = '#' + this.fragment;
       }
 
       // Add a cross-platform `addEventListener` shim for older browsers.
@@ -1700,8 +1700,8 @@
       }
 
       if (!this.options.silent) {
-      return this.trigger('navigate', this.getFragment());
-    }
+        return this.trigger('navigate', this.getFragment());
+      }
     },
 
     // Disable Backbone.history, perhaps temporarily. Not useful in a real app,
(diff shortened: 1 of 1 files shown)


backbone.js
```diff
@@ -1427,6 +1428,9 @@
   var splatParam    = /\*\w+/g;
   var escapeRegExp  = /[\-{}\[\]+?.,\\\^$|#\s]/g;
 
+  // Route handlers registered through Router instances are stored here.
+  Router.handlers = [];
```
akre54: I'm not a fan of this. This just moves the singleton responsibility from `history` to here.

backbone.js
```diff
@@ -1447,16 +1457,12 @@
...
-          Backbone.history.trigger('route', router, name, args);
-        }
-      });
-      return this;
+      var handler = {
+        route: route,
+        callback: callback,
+        name: name
+      };
+      Router.handlers.push(handler);
```
jridgewell: Why use a global `handlers` array and not an instance variable?

backbone.js
```diff
@@ -1471,6 +1477,28 @@
...
+        this.trigger.apply(this, ['route:' + matchedRoute.name].concat(args));
+        this.trigger('route', matchedRoute.name, args);
+      }
+      return this;
+    },
+
+    // Match a fragment with a registered handler
+    matchFragment: function(fragment) {
+      // If the root doesn't match, no routes can match either.
+      if (!Backbone.history.matchRoot()) return false;
```
jridgewell: I think this belongs in `History`, guarding the `trigger('navigate')`.

jmeas: :+1:

backbone.js
jmeas: I had that in my first revision, but I changed it to mimic the current behavior to be less breaking. The current behavior is that if routerA and routerB are both using the same instance of history, and both register the same route, then only one of those handlers will be called, and never both. This is likely why the handlers array is currently stored on the history instance.

backbone.js
jmeas: Oh, actually, you know...I think this might be worth discussing further. If history changes the URL, I can see one arguing that it should emit the navigation event **always** – even if you navigate away from the app's root. This seems simpler to document, and more predictable.

The alternative is defining the `navigate` event as behaving something like:

> The navigate event will be triggered whenever `history` navigates to a fragment, unless that fragment does not contain the application's URL root.

I suppose there could potentially be a *separate* event fired in that case...some sort of navigated-outside-the-app-root event. But this seems unnecessarily complex than it just always navigating.

After typing all of this up, it's making me wonder if the urlRoot even belongs in history, now. I should double-check where it's being used in history...

jridgewell: > After typing all of this up, it's making me wonder if the root even belongs in history, now. I should double-check where it's being used in history...

Mounting routers at different roots? :+1:

backbone.js
jridgewell: I'm not sure that's necessary

backbone.js
```diff
@@ -1748,7 +1757,8 @@
     // you wish to modify the current URL without adding an entry to the history.
     navigate: function(fragment, options) {
       if (!History.started) return false;
-      if (!options || options === true) options = {trigger: !!options};
+      if (options === void 0) options = {silent: true};
+      if (options === false) options = {silent: false};
```
megawac: Nit

```js
if (!options) options = {silent: options === false};
```

megawac:
Very enthusiastic about these changes, esp. the silent gotcha.

backbone.js
jmeas: *If two routers register the same route, why not have both fire there callbacks?*

Shrug. Dunno. I would expect it to behave that way, since the router doesn't really *do* all that much. I think the argument might be something along the lines of..."each route should only have 1 thing happen. so you only need one callback, ever, even if you have multiple routers." But I don't really think this limitation is going to be particularly useful to anyone making a webapp. I just kept the behavior to keep it as BC as possible.

backbone.js
jmeas: Do you mean 

```js
if (!options) options = {silent: options !== false};
```

This seems to work the same as I have :stuck_out_tongue:

jmeas:
There are a few additional changes I would make, but they're less crucial than the ones I've already made. I also think they're less likely to be merged.

1. I would expose the creation of a 'handler' into a separate method. [All of this stuff](https://github.com/jmeas/backbone/blob/61dbf4780422120a4589b2a1b1dab7d28552e112/backbone.js#L1454-L1464), I mean. This clearly separates 'handler' creation from route registration.

   ```js
   route(route, name, callback) {
     var handler = this.createHandler(route, name, callback);
     Router.handlers.push(handler);
   }
   ```

2. I would re-think `onNavigate`. We're finally giving users a single point of entry to respond to when the route changes, which makes it approximately four point two million times more useful than the handful of things that `execute` allows them to do. I'd get rid of execute altogether, and just tell people to overwrite `onNavigate` if they want to do anything special. Based on the Backbone apps I have seen, I expect nearly 100% of developers to refactor their code and override `onNavigate`.

3. I would redefine the terminology used by Backbone. Use of the word 'handler' is unnecessary at best, and confusing at worst. By mapping the terms like so:

  route => path
  handler => route

  one ends up with a much cleaner nomenclature around routing. Rather than associating a route (say, `pasta/:is/*good`) with a callback, and tying these things together in a 'handler' (whatever that is), we can say that a route object has a path and a callback. When the route's path is matched, the route's callback is executed.

3. I would promote using these callbacks in the documentation in ways that some maintainers of Backbone disagree with. Most, if not all, Backbone routers that I see in apps I've worked on are terrible hacks trying to add additional functionality aside from just executing a callback. It would be a huge step forward if Backbone embraced this customizability. I know of very few developers outside of Backbone core who think that a router that simply executes a callback is useful...and I've talked to far too many people about routing in the past year.

jmeas:
> However, I would also like to see the silent gotcha be talked here as well. I want {silent: false} to be the default instead of {silent: true}

Yes, agreed.

jridgewell:
> 1. I would expose the creation of a 'handler' into a separate method.

That's fine. It's a bit like how we separate out `#_createElement`.

> 2. I would re-think `onNavigate`... more useful than the handful of things that `execute` allows them to do.

I don't think there's any benefit in keeping `#execute`, then. Pulling the `trigger('route')` out of `#onNavigate` will make it even easier to override, effectively replacing `#execute`.

```js
class Router {
  constructor() {
    // etc...
    this.listenTo(Backbone.history, 'navigate', this._onNavigate);
  }

  _onNavigate(fragment) {
    if (this.onNavigate(fragment) !== false) {
        this.trigger.apply(this, ['route:' + matchedRoute.name].concat(args));
        this.trigger('route', matchedRoute.name, args);
    }
  }

  onNavigate(fragment) {
    // Override this for your custom code
  }
}
```

> I know that y'all may never be convinced that any router other than the one Backbone provides is a good idea, and that's fine.

I don't think we're fighting you here. As long as routing remains as simple as it is currently, we're all good with exposing hooks to allow customization.

When I say "as simple as it is currently", I mean I want to extend `Router` to include a `routes` hash like it does now. When I instantiate, it "binds" callbacks to URLs. As long as that works, go ahead and make the changes you think are right. That includes the `silent` navigating, etc. We can iron out smaller details about the implementation once all the major ideas have formed.

jmeas:
Todo list:

- [ ] Check where `matchRoot` is being used in Bb.history. Determine if it should be moved to the Router
- [ ] Change `silent` to be `false` by default
- [ ] Update tests
- [ ] Maybe simplify `onNavigate`? Get rid of `execute`? I want to make this as lean as possible, then update the docs to encourage developers to override this method when they want to add additional functionality to the router.

jcrben:
Where's this at? I'm assuming this will make it easy to support this use case http://stackoverflow.com/questions/13970699/backbone-js-force-to-login-route but I'm not exactly sure how that would work it as currently proposed.

I ran into a similar issue with a Marionette app that has a few routers and I wanted to filter access to any (all) routes within a given router (and prevent execution of any associated callback). I ended up monkeypatching Backbone and Marionette code to get Marionette's onRoute to do what I want which is not ideal.

akre54:
@jcarbo we added the `execute` hook for precisely this scenario. Give that a try and see if it solves your issue.

jcrben:
Thanks, not sure how I missed that. But it seems like this is blocking Marionette from providing a more sophisticated router, so I'm hopeful this can be moved forward.

jamesplease:
Closing. It’s probably not worth pursuing this solution; my opinion of routing best practices have changed since I opened this.

The popular routers in the React community are pretty close to a perfect solution, imo

PR #3663, opened 2015-06-04
Collection#set style (Backwards Compatible)

jridgewell:
This is a backwards compatible version of #3648.

- Declares variables where they are needed (and removes duplicate `var i` declarations)
- Exits early when `models` is nullish
- Makes the "merge" and "add" loop easier to follow
- Makes inserting models with `at` an order of magnitude faster
  - [Base case (`#unshift`)](http://jsperf.com/insert-array-into-another-array/14)
  - [Base case (`#push`)](http://jsperf.com/insert-array-into-another-array/15)
  - ["Normal" collection size](http://jsperf.com/insert-array-into-another-array/16)
  - [Large collection](http://jsperf.com/insert-array-into-another-array/17)
  - [Big fetch](http://jsperf.com/insert-array-into-another-array/18)

Depending on how much performance is an issue, we can use an even faster splice algorithm (also included in the jsperfs), it's just a bit complicated.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -857,7 +857,7 @@
 
       // See if sorting is needed, update `length` and splice in new models.
       var orderChanged = false;
-      if (!sortable && add && remove) {
+      if (set.length && !sortable && add && remove) {
         orderChanged = set.length != this.length || _.any(this.models, function(model, index) {
           return model !== set[index];
         });
--- a/backbone.js
+++ b/backbone.js
@@ -865,16 +865,13 @@
         this.length = set.length;
       } else if (toAdd.length) {
         if (sortable) sort = true;
-        this.length += toAdd.length;
+        models = this.models;
         if (at == null) {
-          for (i = 0; i < toAdd.length; i++) {
-            this.models.push(toAdd[i]);
-          }
+          this.models = models.concat(toAdd);
         } else {
-          for (i = 0; i < toAdd.length; i++) {
-            this.models.splice(at + i, 0, toAdd[i]);
-          }
+          this.models = models.slice(0, at).concat(toAdd).concat(models.slice(at, this.length));
         }
+        this.length += toAdd.length;
       }
 
       // Silently sort the collection if appropriate.
--- a/backbone.js
+++ b/backbone.js
@@ -887,7 +887,7 @@
 
       // Unless silenced, it's time to fire all appropriate add/sort events.
       if (!options.silent) {
-        var addOpts = at != null ? _.clone(options) : options;
+        var addOpts = at == null ? options : _.clone(options);
         for (i = 0; i < toAdd.length; i++) {
           model = toAdd[i];
           if (at != null) addOpts.index = at + i;
--- a/backbone.js
+++ b/backbone.js
@@ -823,8 +823,8 @@
 
         // If a duplicate is found, prevent it from being added and
         // optionally merge it into the existing model.
-        var existing;
-        if (existing = this.get(model)) {
+        var existing = this.get(model);
+        if (existing) {
           if (remove) modelMap[existing.cid] = true;
           if (merge && model !== existing) {
             var attrs = this._isModel(model) ? model.attributes : model;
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -788,83 +797,88 @@
...
     set: function(models, options) {
+      if (models == null) return;
+
       options = _.defaults({}, options, setOptions);
       if (options.parse && !this._isModel(models)) models = this.parse(models, options);
+
       var singular = !_.isArray(models);
-      models = singular ? (models ? [models] : []) : models.slice();
-      var id, model, attrs, existing, sort;
+      models = singular ? [models] : _.clone(models);
```
akre54: is `clone` faster here? Shouldn't it be quicker to avoid the logic branch in clone?

akre54:
:+1:

backbone.js
jridgewell: Ha, didn't even notice I changed that.

test/collection.js
```diff
@@ -1565,7 +1565,7 @@
     collection.on('sort', function() {
       ok(true);
     });
-    collection.set([{id: 3}, {id: 2}, {id: 1}, {id: 0}]);
+    collection.set([{id: 1}, {id: 2}, {id: 3}, {id: 0}]);
```
akre54: What's the reason for this change?

test/collection.js
```diff
@@ -506,7 +506,7 @@
     };
     collection.url = '/test';
     collection.fetch();
-    this.syncArgs.options.success();
+    this.syncArgs.options.success([]);
```
akre54: What's the reason for this change?

test/collection.js
jridgewell: "Adding a model should trigger a sort". Seemed better to _only_ add a model, as the mixed order case is handled in the [test above](https://github.com/jashkenas/backbone/pull/3663/files#diff-e95d8d577a70e10552b2cdfe9a7b1178R1549).

test/collection.js
jridgewell: I [preemptively return](https://github.com/jashkenas/backbone/pull/3663/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR800) on `null` now, instead of letting it go through parse.

akre54: Ok but that's a breaking change. Any way to still allow `success()` with no args?

jridgewell: We can remove the check, but I don't think `collection.set()` should clear out the collection...

test/collection.js
akre54: fair nuf. maybe split this out?

test/collection.js
akre54: Agreed. Lets add a test for that.

backbone.js
```diff
@@ -746,6 +746,15 @@
...
   var addOptions = {add: true, remove: false};
 
+  // Splices `insert` into `array` at index `at`.
+  var splice = function(array, insert, at) {
+    var tail = Array(array.length - at);
+    var length = insert.length;
+    for (var i = 0; i < tail.length; i++) tail[i] = array[i + at];
+    for (i = 0; i < length; i++) array[i + at] = insert[i];
+    for (i = 0; i < tail.length; i++) array[i + length + at] = tail[i];
+  };
```
megawac: Whats the advantage of this over the Array.prototype.splice

jridgewell: Speed. :racehorse: (How is there not a racecar emoji?) See the OP.

megawac: Proof of speed?

jridgewell: It's in the [OP](https://github.com/jashkenas/backbone/pull/3663#issue-84837714). :stuck_out_tongue:

test/collection.js
jridgewell: Added a test.

test/collection.js
jridgewell: Split it into what?

jashkenas:
Changing `set` internals is always an adventure — I hope this doesn't break some unexpected things ... but — `splice` function unfortunate black magic aside, this looks pretty nice.

PR #3664, opened 2015-06-04
v2: Collection#parse

jridgewell:
This is the revert of #3659 (which reverted #3644). So, really, it's just #3644 (which reverted #1407... reverts! :stuck_out_tongue:).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -739,6 +739,7 @@
     if (options.comparator !== void 0) this.comparator = options.comparator;
     this._reset();
     this.initialize.apply(this, arguments);
+    if (models && options.parse) models = this.parse(models, options);
     if (models) this.reset(models, _.extend({silent: true}, options));
   };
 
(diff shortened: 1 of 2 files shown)


akre54:
So are we good? We done reverting?

jridgewell:
Our patch release hasn't shipped yet.... :cry:.

jridgewell:
Instead of reverting again, mind if I just rebase it off of master? I promise not to submit another breaking PR until v1.2.1 ships.

akre54:
Not sure what you mean...

jridgewell:
The point of #3659 is I didn't want to ship a (major) breaking change in a patch release. https://github.com/jashkenas/backbone/pull/3644#issuecomment-108026345

akre54:
Ok right but keep in mind that you should never rebase something that others are potentially using. I doubt anyone has pulled from master in the last 2 minutes, but we really should never modify the history of master or bad things will happen. 

(if in the future something shouldn't be merged, please put a big **DO NOT MERGE** tag on it!)

PR #3667, opened 2015-06-07
Collection methods with model-attributes-style predicates

sophietk:
As underscore provides it, a Backbone.Collection functional helper should be able to be called with a model-attributes-style predicate (in addition to function-style).

Example:
```js
this.collection.reject({user: 'guybrush'});
```

Methods concerned: find, detect, filter, select, reject, every, all, any, some, partition.
Compatible with underscore#1.7.0.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -72,6 +72,23 @@
   // `attributes` object or collection's `models` array behind the scenes.
   // `Function#apply` can be slow so we pass the method's arg count, if we know
   // it.
+  var modelMatcher = function(attrs) {
+    var matcher = _.matches(attrs);
+    return function(model) {
+      return matcher(model.attributes);
+    };
+  };
+  var modelProperty = function(key) {
+    return function(model) {
+      return model == null ? void 0 : model.get(key);
+    };
+  };
+  var cb = function(iteratee, instance) {
+    if (_.isFunction(iteratee)) return iteratee;
+    if (_.isObject(iteratee) && !instance._isModel(iteratee)) return modelMatcher(iteratee);
+    if (_.isString(iteratee)) return modelProperty(iteratee);
+    return iteratee;
+  };
   var addMethod = function(length, method, attribute) {
     switch (length) {
       case 1: return function() {
@@ -81,10 +98,10 @@
         return _[method](this[attribute], value);
       };
       case 3: return function(iteratee, context) {
-        return _[method](this[attribute], iteratee, context);
+        return _[method](this[attribute], cb(iteratee, this), context);
       };
       case 4: return function(iteratee, defaultVal, context) {
-        return _[method](this[attribute], iteratee, defaultVal, context);
+        return _[method](this[attribute], cb(iteratee, this), defaultVal, context);
       };
       default: return function() {
         var args = slice.call(arguments);
@@ -947,10 +964,7 @@
     // Return models with matching attributes. Useful for simple cases of
     // `filter`.
     where: function(attrs, first) {
-      var matches = _.matches(attrs);
-      return this[first ? 'find' : 'filter'](function(model) {
-        return matches(model.attributes);
-      });
+      return this[first ? 'find' : 'filter'](attrs);
     },
 
     // Return the first model with matching attributes. Useful for simple cases
@@ -1134,24 +1148,11 @@
       contains: 2, invoke: 0, max: 3, min: 3, toArray: 1, size: 1, first: 3,
       head: 3, take: 3, initial: 3, rest: 3, tail: 3, drop: 3, last: 3,
       without: 0, difference: 0, indexOf: 3, shuffle: 1, lastIndexOf: 3,
-      isEmpty: 1, chain: 1, sample: 3, partition: 3 };
+      isEmpty: 1, chain: 1, sample: 3, partition: 3, groupBy: 3, countBy: 3,
+      sortBy: 3, indexBy: 3};
 
   // Mix in each Underscore method as a proxy to `Collection#models`.
   addUnderscoreMethods(Collection, collectionMethods, 'models');
-
-  // Underscore methods that take a property name as an argument. Proxy to
-  // the model's attributes.
-  var attributeMethods = ['groupBy', 'countBy', 'sortBy', 'indexBy'];
-
-  _.each(attributeMethods, function(method) {
-    if (!_[method]) return;
-    Collection.prototype[method] = function(value, context) {
-      var iterator = _.isFunction(value) ? value : function(model) {
-        return model.get(value);
-      };
-      return _[method](this.models, iterator, context);
-    };
-  });
 
   // Backbone.View
   // -------------
--- a/test/collection.js
+++ b/test/collection.js
@@ -623,7 +623,7 @@
     equal(coll.findWhere({a: 4}), void 0);
   });
 
-  test("Underscore methods", 16, function() {
+  test("Underscore methods", 19, function() {
     equal(col.map(function(model){ return model.get('label'); }).join(' '), 'a b c d');
     equal(col.any(function(model){ return model.id === 100; }), false);
     equal(col.any(function(model){ return model.id === 0; }), true);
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -1126,16 +1126,21 @@
...
-      head: 3, take: 3, initial: 3, rest: 3, tail: 3, drop: 3, last: 3,
-      without: 0, difference: 0, indexOf: 3, shuffle: 1, lastIndexOf: 3,
-      isEmpty: 1, chain: 1, sample: 3, partition: 3 };
+  var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 4, foldl: 4,
+        inject: 4, reduceRight: 4, foldr: 4, include: 2, contains: 2, invoke: 0,
+        max: 3, min: 3, toArray: 1, size: 1, first: 3, head: 3, take: 3, initial: 3,
+        rest: 3, tail: 3, drop: 3, last: 3, without: 0, difference: 0, indexOf: 3,
+        shuffle: 1, lastIndexOf: 3, isEmpty: 1, chain: 1, sample: 3 },
+      collectionMethodsWithModelMatching = { find: 3, detect: 3, filter: 3, select: 3,
+        reject: 3, every: 3, all: 3, any: 3, some: 3, partition: 3 };
```
jridgewell: I think this applies to more than just these methods.

- `map` (`collect`)
- `max`
- `min`

backbone.js
```diff
@@ -79,6 +79,9 @@
         return _[method](this[attribute], value);
       };
       case 3: return function(iteratee, context) {
+        if (predicateTransform && !_.isFunction(iteratee) && _.isObject(iteratee)) {
```
jridgewell: I'm not sure we should limit to just `{ attr: 'value' }` iterating syntax. Think, `col.find('active')`, etc.

backbone.js
```diff
@@ -1126,16 +1126,21 @@
...
+        rest: 3, tail: 3, drop: 3, last: 3, without: 0, difference: 0, indexOf: 3,
+        shuffle: 1, lastIndexOf: 3, isEmpty: 1, chain: 1, sample: 3 },
+      collectionMethodsWithModelMatching = { find: 3, detect: 3, filter: 3, select: 3,
+        reject: 3, every: 3, all: 3, any: 3, some: 3, partition: 3 };
 
   // Mix in each Underscore method as a proxy to `Collection#models`.
   addUnderscoreMethods(Collection, collectionMethods, 'models');
+  addUnderscoreMethods(Collection, collectionMethodsWithModelMatching, 'models', function(attrs) {
+    return function(model) {
+      return _.matches(attrs)(model.attributes);
```
jridgewell: `_.matches` crates a new function on every call, meaning we're creating `n` functions here. If we pull it out to outer method, we can reuse the same method multiple times (just like the [old `#where`](https://github.com/jashkenas/backbone/pull/3667/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcL947))

jridgewell:
I really like the idea, I'm surprised this hasn't been requested before. Mind addressing the comments?

backbone.js
sophietk: I think it can only apply to filtering methods.

What should return ```collection.map({user: 'lechuck'})```? (not filter, but transformation)
And ```collection.max({user: 'lechuck'})```? (max must take a compared attribute, not a set of attributes)

backbone.js
sophietk: You're totally right. Improvement is coming.

backbone.js
sophietk: Indeed it would be helpful to have (like underscore):
```js
collection.filter('user');
collection.map('user');
```
for
```js
collection.filter(function isUserDefined(model) {
  return !_.isUndefined(model.get('user')); // or isFalsy ?
});
collection.map(function getUser(model) {
  return model.get('user');
});
```
I'm working on it.

backbone.js
sophietk: Have played with underscore, and I'm surprised this use is allowed! I'm trying to make the same for collection.

backbone.js
```diff
@@ -70,6 +70,22 @@
 
   // Proxy Underscore methods to a Backbone class' prototype using a
   // particular attribute as the data argument
+  var modelMatcher = function(attrs) {
+    return function(model) {
+      return _.matches(attrs)(model.attributes);
```
akre54: You can pull `_.matches` into the outer scope to save recreating the function all the time:

```js
var modelMatcher = function(attrs) {
  var matcher = _.matches(attrs);
  return function(model) {
    return matcher(model.attributes);
  };
};

```

backbone.js
```diff
@@ -70,6 +70,23 @@
...
+    };
+  };
+  var modelProperty = function(key) {
+    return function(model) {
+      return model == null ? void 0 : model.get(key);
+    };
+  };
+  var cb = function(iteratee) {
+    if (_.isFunction(iteratee)) return iteratee;
+    if (_.isObject(iteratee) && !(iteratee instanceof Backbone.Model)) return modelMatcher(iteratee);
```
jridgewell: Is there a test case for the `!instanceof` condition? It's not obvious to me why it's needed.

sophietk: Actually the `collection.indexOf` test failed when trying to locate a Backbone.Model in the collection. Since it's an Object, it was catched by the second `if` and was converted to a function. In order to keep the indexOf behaviour unchanged, I added the `!instanceof` test.

jridgewell: Ah, gotcha. In that case, Collections have `#_isModel` to check for `instanceof` that can be overridden to support polymorphic models.

jridgewell:
Looks like the [`attributeMethods`](https://github.com/sophietk/backbone/blob/1a3e38a97a029c109d86d3f9a794c174715736c3/backbone.js#L1155-L1166) can be folded into `collectionMethods` with this as well.

backbone.js
```diff
@@ -70,6 +70,23 @@
...
+    };
+  };
+  var modelProperty = function(key) {
+    return function(model) {
+      return model == null ? void 0 : model.get(key);
+    };
+  };
+  var cb = function(iteratee) {
+    if (_.isFunction(iteratee)) return iteratee;
+    if (_.isObject(iteratee) && !Backbone.Collection.prototype._isModel(iteratee)) return modelMatcher(iteratee);
```
jridgewell: If you pass in `this`, we can check directly on the instance:

```js
var cb = function(iteratee, instance) {
  if (_.isFunction(iteratee)) return iteratee;
  if (_.isObject(iteratee) && !instance._isModel(iteratee)) return modelMatcher(iteratee);
  if (_.isString(iteratee)) return modelProperty(iteratee);
  return iteratee;
}

// ...
  return _[method](this[attribute], cb(iteratee, this), context);
```

sophietk: Done, feel free to collaborate on this branch.

backbone.js
```diff
@@ -70,6 +70,23 @@
...
+    };
+  };
+  var modelProperty = function(key) {
+    return function(model) {
+      return model == null ? void 0 : model.get(key);
+    };
+  };
+  var cb = function(iteratee, instance) {
+    if (_.isFunction(iteratee)) return iteratee;
+    if (_.isObject(iteratee) && !instance._isModel(iteratee)) return modelMatcher(iteratee);
```
akre54: it's cool to use `instance._isModel` here because instance will only ever be a Collection, right? (because `cb` is only used on 3- and 4-length methods, and modelMethods only uses 0- or 1-length).

akre54:
This looks great, thanks. Can you squash down into a single commit and rebase from master and we'll merge?

backbone.js
jridgewell: That's what I was thinking.

akre54: Ok cool. We're probably fine but this would be a bug if we ever introduced 3- or 4-length methods in the future.

jridgewell:
Thanks for the PR!

sophietk:
And thanks for the code reviews :)

akre54:
@jridgewell @sophietk this caused the build to break in IE7 and IE8: https://travis-ci.org/jashkenas/backbone/builds/66548299. Mind taking a look?

sophietk:
Oops, didn't see that. Your help is appreciated.

PR #3670, opened 2015-06-09
Remove moot `version` property from bower.json

kkirsche:
Per bower/bower.json-spec@a325da3


akre54:
Cool, thanks. It makes a lot more sense to just use tags (not sure why npm can't do this too).

PR #3673, opened 2015-06-09
Fix initialization of views with undefined arguments

iaman:
Fix-It Felix strolling in
[A recent commit](https://github.com/jashkenas/backbone/commit/51eed189bf4d25877be4acdf51e0a4c6039583c5#comments) caused view initialization to fail if the view is initialized without arguments. This may be an edge case, but in my opinion `new Backbone.View()` shouldn't be something that breaks like this (and a few others on the comment thread for that commit share my opinion).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1166,6 +1166,7 @@
   // if an existing element is not provided...
   var View = Backbone.View = function(options) {
     this.cid = _.uniqueId('view');
+    options = typeof options !== 'undefined' ? options : {};
     _.extend(this, _.pick(options, viewOptions));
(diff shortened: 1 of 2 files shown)


jridgewell:
You're using an older version of underscore. For reference, Backbone depends on underscore v1.7 (where `_.pick(undefined, 'attr')` support was added) [since v1.2](https://github.com/jashkenas/backbone/blob/1.2.0/package.json#L8).

iaman:
@jridgewell Huh, I thought I was using the vendored version but apparently that is not the case. Thanks!

PR #3674, opened 2015-06-09
Remove uses of _.bindAll

### LINKED ISSUE #3671: Ditch bindAll use.
`bindAll` is used once in Backbone and it could be swapped out with `bind`.

jridgewell:
Fixes #3671.


jridgewell:
Deferring to #3675.

PR #3675, opened 2015-06-09
Remove the only `_.bindAll` use.

jdalton:
Remove the only `_.bindAll` use.


jridgewell:
Lol, I'll defer since you opened the initial issue. Mind removing the `_.bindAll` in `test/router.js`?

PR #3676, opened 2015-06-09
Add Collection#includes for underscore v1.8

### LINKED ISSUE #3672: Drop `include` method use for `includes`
The Underscore method `include` should be dropped in favor of `includes` in BB.

jridgewell:
Closes #3672 (kind of, we still can't deprecate `Collection#include` until we officially bump our underscore dependency to 1.8).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1128,8 +1128,8 @@
   // right here:
   var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 4,
       foldl: 4, inject: 4, reduceRight: 4, foldr: 4, find: 3, detect: 3, filter: 3,
-      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 2,
-      contains: 2, invoke: 0, max: 3, min: 3, toArray: 1, size: 1, first: 3,
+      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 3, includes: 3,
+      contains: 3, invoke: 0, max: 3, min: 3, toArray: 1, size: 1, first: 3,
       head: 3, take: 3, initial: 3, rest: 3, tail: 3, drop: 3, last: 3,
       without: 0, difference: 0, indexOf: 3, shuffle: 1, lastIndexOf: 3,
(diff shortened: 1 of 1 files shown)


akre54:
Might we think about taking the same tact as React to `_.once`-wrap deprecated functions?

backbone.js
```diff
@@ -1128,7 +1128,7 @@
   // right here:
   var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 4,
       foldl: 4, inject: 4, reduceRight: 4, foldr: 4, find: 3, detect: 3, filter: 3,
-      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 2,
+      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 2, includes: 2,
```
jridgewell: Ha, includes takes 3 arguments, anyways. Should we add `fromIndex` to `Collection#contains` as well?

jridgewell:
> Might we think about taking the same tact as React to _.once-wrap deprecated functions?

I'd be cool with that. Same PR?

backbone.js
jdalton: Nooice!

jridgewell:
I'm going to defer on once wrapping to another PR.

backbone.js
```diff
@@ -1128,7 +1128,7 @@
   // right here:
   var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 4,
       foldl: 4, inject: 4, reduceRight: 4, foldr: 4, find: 3, detect: 3, filter: 3,
-      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 2,
+      select: 3, reject: 3, every: 3, all: 3, some: 3, any: 3, include: 2, includes: 3,
```
megawac: Should `include` and `contains` change to 3 arity?

jridgewell: I only did it for `includes` since it's the meant to emulate the ES6 method, but I could see putting the others to `3` as well.

jridgewell: Amended.

jdalton:
Ping?

PR #3678, opened 2015-06-10
Replace `_.any` with `_.some`.

jdalton:
Replace `_.any` with `_.some`.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1730,7 +1730,7 @@
       // If the root doesn't match, no routes can match either.
       if (!this.matchRoot()) return false;
       fragment = this.fragment = this.getFragment(fragment);
-      return _.any(this.handlers, function(handler) {
+      return _.some(this.handlers, function(handler) {
         if (handler.route.test(fragment)) {
           handler.callback(fragment);
           return true;


docs/backbone.html
```diff
@@ -4556,7 +4556,7 @@ <h2 id="backbone-history">Backbone.History</h2>
             
             <div class="content"><div class='highlight'><pre>      <span class="hljs-keyword">if</span> (!<span class="hljs-keyword">this</span>.matchRoot()) <span class="hljs-keyword">return</span> <span class="hljs-literal">false</span>;
       fragment = <span class="hljs-keyword">this</span>.fragment = <span class="hljs-keyword">this</span>.getFragment(fragment);
-      <span class="hljs-keyword">return</span> _.any(<span class="hljs-keyword">this</span>.handlers, <span class="hljs-function"><span class="hljs-keyword">function</span><span class="hljs-params">(handler)</span> </span>{
+      <span class="hljs-keyword">return</span> _.some(<span class="hljs-keyword">this</span>.handlers, <span class="hljs-function"><span class="hljs-keyword">function</span><span class="hljs-params">(handler)</span> </span>{
```
jridgewell: Mind reverting this change? [Trying](https://github.com/jashkenas/backbone/issues/3597) to keep docs in sync with the release source.

akre54:
Eh. I think `any` actually reads cleaner here. Just because it's the canonical name doesn't mean we have to sacrifice semantics.

PR #3679, opened 2015-06-10
Update comments throughout the code

akre54:
This is an attempt to unify and clarify Backbone's comments to make it easier to understand some of the tricky parts. Particularly where we have extra optimizations and higher-order functions (underscore methods and Events mainly), it'd be great to make sure it's clear *why* we're doing what we're doing.

I'd also like to detangle `Events` even further. There's a lot going on there that can be hard to follow at first and is very implementation-specific. Some info about the way we're passing around listen ids and callbacks would go a long way, I think.

For instance: @jridgewell should we rename `memo` in `eventsApi` to `events`? I get that it's a reference to `reduce`, but I think the `memo` terminology is confusing, especially since, with the exception of `once`, it always refers to the object's `_events` property.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -68,8 +68,10 @@
   // form param named `model`.
   Backbone.emulateJSON = false;
 
-  // Proxy Underscore methods to a Backbone class' prototype using a
-  // particular attribute as the data argument
+  // Proxy Backbone class methods to Underscore functions, wrapping the model's
+  // `attributes` object or collection's `models` array behind the scenes.
+  // `Function#apply` can be slow so we pass the method's arg count, if we know
+  // it.
   var addMethod = function(length, method, attribute) {
     switch (length) {
       case 1: return function() {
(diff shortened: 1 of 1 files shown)


backbone.js
```diff
@@ -68,8 +68,10 @@
   // form param named `model`.
   Backbone.emulateJSON = false;
 
-  // Proxy Underscore methods to a Backbone class' prototype using a
-  // particular attribute as the data argument
+  // Proxy Backbone class methods to Underscore functions, wrapping the model's
+  // attributes object or collection's models array behind the scenes.
```
jridgewell: Should we wrap `attributes` and `models` in backticks?

jridgewell:
> should we rename memo in eventsApi to events?

Sure, that sounds smart to me.

backbone.js
akre54: Yeah good call, I wasn't sure since it's kind of inconsistent at the moment. Same with Model vs model and Collection vs collection.

PR #3680, opened 2015-06-10
Event maps should support multiple event syntax

akre54:
Edge case, but `obj.on({'a b c': increment})` should work the same as `obj.on('a b c', increment)`.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -126,10 +126,10 @@
       // Handle event maps.
       if (callback !== void 0 && 'context' in opts && opts.context === void 0) opts.context = callback;
       for (names = _.keys(name); i < names.length ; i++) {
-        events = iteratee(events, names[i], name[names[i]], opts);
+        events = eventsApi(iteratee, events, names[i], name[names[i]], opts);
       }
     } else if (name && eventSplitter.test(name)) {
-      // Handle space separated event names.
+      // Handle space separated event names by delegating them individually.
       for (names = name.split(eventSplitter); i < names.length; i++) {
         events = iteratee(events, names[i], callback, opts);
       }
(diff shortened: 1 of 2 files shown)


jridgewell:
I didn't realize that was supported with the old `eventsApi`. But oh man, that's going to be expensive. Since objects can only have string objects, do we want to duplicated the split for-loop, or combine the events hash and splits into a single loop?

akre54:
> I didn't realize that was supported with the old eventsApi. 

I'm not sure it ever was (or necessarily should be now). But it seems a tad inconsistent to treat them differently.

> But oh man, that's going to be expensive. 

Is it? Not much more than it is currently I would imagine... jQuery [seems](https://github.com/jquery/jquery/blob/2.1.4/src/event.js#L777-L779) [to be managing](https://github.com/jquery/jquery/blob/2.1.4/src/event.js#L815-L817) alright. 

> combine the events hash and splits into a single loop

How would we do that?

jridgewell:
> Is it? Not much more than it is currently I would imagine... jQuery seems to be managing alright.

I guess it's not too bad of a hit. http://jsperf.com/eventsapi-multi-event-maps

> combine the events hash and splits into a single loop

My ways's actually slower, so never mind.

akre54:
For reference, this was supported with the old `eventsApi` because we [called `object.on` instead of running the `eventsApi` again](https://github.com/jashkenas/backbone/blob/1.1.2/backbone.js#L183). The events refactor missed this because we didn't have any tests for this behavior.

jashkenas:
> The events refactor missed this because we didn't have any tests for this behavior.

The culprit. Bummer.

stevewillard:
Anyone know when this change will be part of a release?

PR #3682, opened 2015-06-12
modelProperty doesn't need to check for null

jridgewell:
Since `#models` is filled with only model instances, there's no need to check if model isn't `null`.

re: #3667.


akre54:
:+1: we could probably also inline this into `cb`.

Issue #3560, opened 2015-04-07
Backbone and ES6 Classes


benmccormick:
With the final changes to the ES6 class spec (details [here](http://www.2ality.com/2015/02/es6-classes-final.html)), it's no longer possible to use ES6 classes with Backbone without making significant compromises in terms of syntax.  I've written a full description of the situation [here](http://benmccormick.org/2015/04/07/es6-classes-and-backbone-js/) (make sure to click through to the comments at the bottom for an additional mitigating option), but essentially there is no way to add properties to an instance of a subclass prior to the subclasses parents constructor being run.

So this:

```javascript
class DocumentRow extends Backbone.View {

    constructor() {
        this.tagName =  "li";
        this.className = "document-row";
        this.events = {
            "click .icon":          "open",
            "click .button.edit":   "openEditDialog",
            "click .button.delete": "destroy"
        };
        super();
    }

    initialize() {
        this.listenTo(this.model, "change", this.render);
    }

    render() {
        //...
    }
}
```

is no longer valid in the final ES6 spec.  Instead you effectively have 3 (not very appealing) options if you want to try to make this work:

- **Attach all properties as functions** - Backbone allows this, but it feels dumb to write something like this: 

```javascript
class DocumentRow extends Backbone.View {

    tagName() { return "li"; }

    className() { return "document-row";}

    events() {
        return {
            "click .icon":          "open",
            "click .button.edit":   "openEditDialog",
            "click .button.delete": "destroy"
        };
    }

    initialize() {
        this.listenTo(this.model, "change", this.render);
    }

    render() {
        //...
    }
}
```

compared to the current extends syntax

- **Run the constructor twice** - I don't view this as a real option due to the issues it would cause running initialize a second time with different cids, etc.

- **Pass all properties as default options to the superclass constructor** -  This was suggested by a commenter on my blog and is probably the most practical current option.  It looks something like this:

```javascript
class MyView extends Backbone.View {
  constructor(options) {
    _.defaults(options, {
      // These options are assigned to the instance by Backbone
      tagName: 'li',
      className: 'document-row',
      events: {
        "click .icon": "open",
        "click .button.edit": "openEditDialog",
        "click .button.delete": "destroy"
      },
      // This option I'll have to assign to the instance myself
      foo: 'bar'
    });


    super(options);


    this.foo = options.foo;
  }
}
```

[...]

PR #3684, opened 2015-06-15
Allow #idAttribute as a constructor option

jridgewell:
Also allows `#idAttribute` to be a method like `View`'s constructor options.

This will ease ES6 class transition by allowing either:
- Constructor default overloading (https://github.com/jashkenas/backbone/issues/3560#user-content-default-options)
- Option method definitions (https://github.com/jashkenas/backbone/issues/3560#user-content-methods)

**Mildly Breaking**: Instantiating a Collection with an explicit `{ comparator: undefined }` will now disable sorting on the instance. Previously, only the other false-y values would disable.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -394,15 +394,18 @@
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
-    this.cid = _.uniqueId(this.cidPrefix);
     this.attributes = {};
-    if (options.collection) this.collection = options.collection;
+    _.extend(this, _.pick(options, modelOptions));
+    this.cid = _.uniqueId(this.cidPrefix);
     if (options.parse) attrs = this.parse(attrs, options) || {};
     attrs = _.defaults({}, attrs, _.result(this, 'defaults'));
     this.set(attrs, options);
     this.changed = {};
     this.initialize.apply(this, arguments);
   };
+
+  // List of models options to be set as properties
+  var modelOptions = ['cidPrefix', 'idAttribute', 'collection'];
 
   // Attach all inheritable methods to the Model prototype.
   _.extend(Model.prototype, Events, {
@@ -704,7 +707,10 @@
 
     // Create a new model with identical attributes to this one.
     clone: function() {
-      return new this.constructor(this.attributes);
+      return new this.constructor(this.attributes, {
+        idAttribute: this.idAttribute,
+        cidPrefix: this.cidPrefix
+      });
     },
 
     // A model is new if it has never been saved to the server, and lacks an id.
--- a/backbone.js
+++ b/backbone.js
@@ -396,7 +396,7 @@
     options || (options = {});
     this.attributes = {};
     _.extend(this, _.pick(options, modelOptions));
-    this.cid = _.uniqueId(this.cidPrefix);
+    this.cid = _.uniqueId(_.result(this, 'cidPrefix'));
     if (options.parse) attrs = this.parse(attrs, options) || {};
     attrs = _.defaults({}, attrs, _.result(this, 'defaults'));
     this.set(attrs, options);
@@ -509,7 +509,7 @@
       }
 
       // Update the `id`.
-      this.id = this.get(this.idAttribute);
+      this.id = this.get(_.result(this, 'idAttribute'));
 
       // Trigger all relevant attribute changes.
       if (!silent) {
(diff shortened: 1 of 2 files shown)


akre54:
I'm not sure how I feel about this pull.. I don't think this is the way we should be supporting ES6 classes.

The existing options are all conceivably ones you'd want to pass in during instantiation, you really wouldn't want to be passing `idAttribute` or `cidPrefix` every time and every place you instantiate your model. It puts the responsibility on the wrong part of your code and ends up being a pretty nasty hack.

jridgewell:
> The existing options are all conceivably ones you'd want to pass in during instantiation, you really wouldn't want to be passing idAttribute or cidPrefix every time

Definitely agree to this. I think this will be a rarely used feature, but I was aiming to mirror what we are doing with other constructor options.

> Forcing this all to go in the constructor, meanwhile, is also a hack. You're now bloating the memory usage of your app (you no longer share these properties via prototypes, they're now specific to the instance)

I don't agree with this, though. Only individual instances will have an `#idAttribute` directly set, everything that doesn't pass the option will defer to the prototype.

akre54:
> Only individual instances will have an #idAttribute directly set, everything that doesn't pass the option will defer to the prototype.

Not sure what you mean by this. Two models created by `new Model({idAttribute: '_id'})` (or the constructor way) will have two different strings for their `idAttribute` instead of sharing one as they would've with a shared prototype property (as we do now).

Not a huge deal -- we're talking about small strings -- but it feels unclean and is something we'll have to keep around as cruft. What's the issue with using the `get` way if you feel compelled to use ES6 `class`?

jridgewell:
Oh, I mean if you create two instances without passing `{ idAttribute: '_id' }`, they'll share the same `idAttribute` from the prototype.

```js
new Model({idAttribute: '_id'}); // Has it's own idAttribute
new Model(); // Shares idAttribute
new Model(); // Shares idAttribute
```

> What's the issue with using the `get` way if you feel compelled to use ES6 class?

Old IE support and drastically slower access times, mainly.

akre54:
But how does that help us here? If you need a custom idAttribute you'll want it used on all instances of your model. In your example the idAttribute can only ever be `"id"`.

jridgewell:
The [other part](https://github.com/jridgewell/backbone/commit/ab394c28b793c6b7e5442d8b0c4d3a9f988d863f) of the PR is allowing `idAttribute` to be a function.

Allowing the constructor option allows a subclass to pass [default options](https://github.com/jashkenas/backbone/issues/3560#user-content-default-options) from the subclass constructor, and allowing functions allows it to be set [directly as a method](https://github.com/jashkenas/backbone/issues/3560#user-content-methods). Those are the two patterns we identified as the way forward using ES6 classes.

akre54:
> The other part of the PR is allowing idAttribute to be a function.

I don't have a huge objection to this, but aside from ES6 `class` interop I'm not sure when you'd want this to be a function. It seems close to `View#tagName` (which I'm also not sure why you would ever be a function, except that `tagName` aligns with `className` and `id`, which you might).

> > What's the issue with using the get way if you feel compelled to use ES6 class?

> Old IE support

But with ES6 isn't that a moot point anyway? Even with a babel transpile it's [still using Object.defineProperty](http://babeljs.io/repl/#?experimental=false&evaluate=false&loose=false&spec=false&playground=false&code=let%20Backbone%20%3D%20%7B%7D%0ABackbone.View%20%3D%20class%20View%20%7B%7D%0A%0Aclass%20Row%20extends%20Backbone.View%20%7B%0A%20%20tagName()%20%7B%20return%20'li'%3B%20%7D%0A%7D%0A) [which is IE8+](http://kangax.github.io/compat-table/es5/#Object.defineProperties).

> and drastically slower access times, mainly.

Fair enough. I personally dislike getters but if you absolutely must use Backbone with ES6 classes this is your best option, in my opinion.

My point is that we're working at the bleeding edge, trying to fix semantics that likely may change in our favor in the near future. The same week TC39 approves static properties on classes, we're likely to see it land on Babel core and devs already using Babel will get it for free instantly. Given that almost no one today is using Backbone with ES6 classes on a native environment, we're probably safe when this happens. We no longer have to wait for the long tail of oldIE to flush out before we can take advantage of brand new language features (especially when using other brand new features with it).

This PR is well-intentioned but seems like a stopgap, a half-solution to the issues plaguing ES6 classes. My suggestion is to use `Model.extend` until real static properties land in ES classes, or getters if you can't wait. It's not worth adding something that uglies our API and may be changed in short order.

PR #3685, opened 2015-06-17
Allow tagName to use parameters passed to the constructor

mattparlane:
I recently wanted to define tagName as a function and I wanted the return value to vary based on parameters passed to the constructor, but I found that _ensureElement is called before initialize and so the constructor parameters were not available in tagName.

This PR simply reverses the order of initialize and _ensureElement. All tests pass, plus a new one added.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1170,8 +1170,8 @@
   var View = Backbone.View = function(options) {
     this.cid = _.uniqueId('view');
     _.extend(this, _.pick(options, viewOptions));
-    this._ensureElement();
     this.initialize.apply(this, arguments);
+    this._ensureElement();
   };
 
   // Cached regex to split keys for `delegate`.
--- a/test/view.js
+++ b/test/view.js
@@ -205,6 +205,22 @@
     ok(new View().$el.is('p'));
   });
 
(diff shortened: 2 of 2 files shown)


akre54:
I'm not sure about the wisdom of changing the `initialize` ordering at this point. You usually want `this.el` to be set up for you in `initialize` and making this change will break many peoples' code.

If you need it in your app, you can always try putting your logic in the constructor:

```js
  constructor: function(options) {
    this.myVar = options.myVar;
    return Backbone.View.apply(this, arguments);
  }
});
```

jridgewell:
> I'm not sure about the wisdom of changing the initialize ordering at this point. You usually want this.el to be set up for you in initialize and making this change will break many peoples' code.

Agreed. I think a better option would be ferrying `options` into `_ensureElement`, and into our view hook methods.

akre54:
Yeah we could do that. But since `_.result` doesn't pass along any options, you'd have to set `this.myVar` in `_ensureElement` or something. Got a better idea?

akre54:
Also @mattparlane what's your use case for switching up `tagName` dynamically?

jashkenas:
Just call `setElement` in initialize.

PR #3689, opened 2015-06-24
Fix minor toJSON sample usage code (it used JSON.stringify)

dxuehu:
prbackbone


dxuehu:
Nevermind, I should've looked more closely at the JSON api as it said.

PR #3690, opened 2015-06-24
Fix test for IE7/IE8 build

sophietk:
Wrapping test object keys as string to avoid IE7/IE8 error (reserved identifiers).
Tests were added with #3667 : https://travis-ci.org/jashkenas/backbone/builds/66548299.

### DIFF
--- a/test/collection.js
+++ b/test/collection.js
@@ -678,8 +678,8 @@
     deepEqual(coll.map('a'), [1, 2, 3, 4]);
     deepEqual(coll.max('a'), model);
     deepEqual(coll.min('e'), model);
-    deepEqual(coll.countBy({a: 4}), {false: 3, true: 1});
-    deepEqual(coll.countBy('d'), {undefined: 4});
+    deepEqual(coll.countBy({a: 4}), {'false': 3, 'true': 1});
+    deepEqual(coll.countBy('d'), {'undefined': 4});
   });
 
   test("reset", 16, function() {


jridgewell:
Thanks for looking into it!

akre54:
Grr oldIE.

Thanks for the quick fix.

megawac:
Hmm, we should set up (actually run) linting for backbone
/cc @michaelficarra

michaelficarra:
Yes, probably. A linting step is basically free, and takes 5 minutes to set up. I suggest someone with more familiarity with the project than myself does this, though.

akre54:
It's [already](https://github.com/jashkenas/backbone/blob/1.2.1/package.json#L22) [set up](https://github.com/jashkenas/backbone/blob/1.2.1/docs/jsl.conf). We could use eslint, like Underscore, too though.

megawac:
> It's already set up. We could use eslint, like Underscore, too though.
Yeah but we don't run it through CI and haven't updated it in a long time

akre54:
For sure, but it would be a [one line change](https://github.com/jashkenas/underscore/blob/1.8.3/package.json#L28-L29) to set it up with CI. I'd be fine just cribbing Underscore's .eslintrc here.

michaelficarra:
If the projects follow similar styles and have similar browser support, I'd say that's a good idea.

Issue #3688, opened 2015-06-23
Collection: get by id/_addReference is broken


jpaugh:
It looks like the `Collection._addReference` method is broken, around line [1108](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1108):

      var id = this.modelId(model.attributes);
      if (id != null) this._byId[id] = model;

Shouldn't this look more like:

      var idAttr = this.modelId(model.attributes);
      if (idAttr != null) var id = model[idAttr];
      if (id != null) this._byId[idAttr] = model;

That is, it's trying to index the model by the value `"id"` rather than the value `model["id"]`.

This, of course, breaks `get` as well, and probably other things.

akre54:
The code is working as intended but the docs are confusingly worded. `modelId` is meant to return the value of the `id` given the model attributes. It's the equivalent of the result of `model.attributes[idAttribute]`, not `idAttribute` itself.

I'll update the docs.

jpaugh:
Thank you for your quick response to my issue. However, I don't think this change makes the documentation for `modelId` any more (or less) clear, at least for my specific issue. I've added clarification to #3692, and reopened it.

jpaugh:
@akre54 I see in in #3692, you've updated the documentation for `modelId`; However, what I'm really misunderstanding at the moment the documentation for `Collection.get`:

> get            collection.get(id) 
> Get a model from a collection, specified by an id, a cid, or by passing in a model.

From that, it seems that I should be able to use an `id` to retrieve a model from the collection, but when reading the code, it does not fetch my a model id, but by the literal string `"id"` instead.

I have this issue 
What I misunderstand at the moment the documentation for `Collection.get`:

> get            collection.get(id) 
> Get a model from a collection, specified by an id, a cid, or by passing in a model.

However, when reading the code, it does not fetch my a model id, but by the literal string `"id"` instead. 

I'm running into this issue on a live site, but in case I'm misreading the (bootstrap) code, I'm currently in the process of constructing a smaller test case.

jridgewell:
Grabbing a model by id is handled by the [first conditional](https://github.com/jashkenas/backbone/blob/f55b4a22458ad01a8246bc79081858c8a2e872dc/backbone.js#L955) in `#get` (ignoring the `#modelId` call):

```js
get: function(obj) {
  return this._byId[obj]
}
```

jpaugh:
[...]

PR #3692, opened 2015-06-24
update modelId docs

akre54:
Make documentation a little clearer. Addresses #3688

### DIFF
--- a/index.html
+++ b/index.html
@@ -1739,7 +1739,7 @@ bill.set({name : "Bill Jones"});
     </p>
 
     <p id="Collection-model">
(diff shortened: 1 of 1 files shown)


jpaugh:
Thank you for your quick response to #3688. However, I don't think this change makes the documentation for `modelId` any more (or less) clear, at least to me.

akre54:
Yeah definitely. This method is pretty specific to the internal implementation details of Backbone, so its docs are bound to be jargon-heavy. I'll see what I can do about teasing out a simpler description.

PR #3695, opened 2015-06-25
Optimize Events.trigger method

wovchok:


Florian-R:
IIRC it was done in a similar fashion before, but was changed for perfs considerations. See 2c8398483ef3589a1b01b52fd11923c26e68a0df

Issue #3662, opened 2015-06-03
.trigger('change')


acjohn:
v1.2.0 line 1117 doesn't contain 'model &&'
When setting value (using "silent:true") and then chaining & triggering a "change"  event, it throws an error

model.set({'some':'data'}, {'silent':true}).trigger('change');

I know it says "not" to use silent...

jridgewell:
I'm not sure this needs to be fixed. The `change` event always passes `model` as the first argument, and an `options` object as the second. Anything else is breaking the contract.

`model.trigger('change', model, options)`.

acjohn:
Hi Justin,

It's odd that it only happens when the model is in a collection.
I guess it's in my code somewhere.



Regards,

John Schroeder
johns@artisancolour.com
480-948-0009

PR #3699, opened 2015-06-30
Fix manually triggering "change" on model belonging to a collection

rafBM:
Since 4e2d209 there would always be an error when doing `model.trigger('change')` when `model` was part of a collection.


jridgewell:
Dupe of https://github.com/jashkenas/backbone/issues/3662.

PR #3702, opened 2015-07-07
Code clean-up (History.getFragment).

nowamasa:
A minor code clean-up.


akre54:
No thanks. We like to keep lines under 80 characters if we can.

Issue #3708, opened 2015-07-13
Collection set method going into merge with same models/objects already present


elgubenis:
```js
var collection = new Backbone.Collection([{id: 1}, {id: 2}]);
collection.set([{id: 1}, {id: 2}]);
```

This code will enter the merging process of the 2 already existing models, im not sure if thats right, because no change of attributes will happen.

Instantiating a new collection with an array of objects creates a model for every object, but the 'merge-check' in backbone is comparing against bare attributes if no models (but an array of objects is) added on the set-method.
Though it says "// Turn bare objects into model references, ...", but i can't spot where the bare object is turned into a model reference (for comparing if there is a need to merge)

```js
if (merge && attrs !== existing) {
```
from
```js
      // Turn bare objects into model references, and prevent invalid models
      // from being added.
      for (var i = 0; i < models.length; i++) {
        attrs = models[i];

        // If a duplicate is found, prevent it from being added and
        // optionally merge it into the existing model.
        if (existing = this.get(attrs)) {
          if (remove) modelMap[existing.cid] = true;
          if (merge && attrs !== existing) {
            attrs = this._isModel(attrs) ? attrs.attributes : attrs;
            if (options.parse) attrs = existing.parse(attrs, options);
            existing.set(attrs, options);
            if (sortable && !sort && existing.hasChanged(sortAttr)) sort = true;
          }
          models[i] = existing;

        // If this is a new, valid model, push it to the `toAdd` list.
        } else if (add) {
          model = models[i] = this._prepareModel(attrs, options);
          if (!model) continue;
          toAdd.push(model);
          this._addReference(model, options);
        }
```

If the existing piece of code is correct as is, could someone point out the reason why?

elgubenis:
as far as I understand, we would get false here:
```javascript
attrs !== existing
```
because:
```javascript
{ a: 1 } != { a: 1 }
// and
{ a: 1, b: 1} != { b: 1, a: 1 }
```

elgubenis:
I am already preparing some PR, but maybe thats not necessary because there is just some explanation missing for me to understand why thats done like that.

elgubenis:
As im thinking, that might be the reason why the merge switch exists for the options, but i think that even when merging is desired that there are some models which dont need any merging because they are exactly alike.

jridgewell:
I'm unsure what the problem is?

elgubenis:
[...]

PR #3711, opened 2015-07-13
adding feedback to update event of collection

### LINKED ISSUE #3700: Backbone update event on collection should return the newly added or removed model(s)

elgubenis:
collection's update event receives 2 additional attributes, in case of removal of existing models, previousModels is an array containing the removed models which where referenced to the collection before .set was called. The second attribute, addedModels contains an array of models which were added to the collection calling the .set method.

Fixes #3700

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -808,6 +808,7 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       var removed = this._removeModels(models, options);
+      options.removedModels = removed;
       if (!options.silent && removed) this.trigger('update', this, options);
       return singular ? removed[0] : removed;
     },
@@ -832,6 +833,7 @@
       var set = [];
       var toAdd = [];
       var toRemove = [];
+      var toMerge = [];
       var modelMap = {};
 
       var add = options.add;
@@ -856,6 +858,7 @@
             var attrs = this._isModel(model) ? model.attributes : model;
             if (options.parse) attrs = existing.parse(attrs, options);
             existing.set(attrs, options);
+            toMerge.push(existing);
             if (sortable && !sort) sort = existing.hasChanged(sortAttr);
           }
           if (!modelMap[existing.cid]) {
@@ -904,7 +907,7 @@
       // Silently sort the collection if appropriate.
       if (sort) this.sort({silent: true});
 
-      // Unless silenced, it's time to fire all appropriate add/sort events.
+      // Unless silenced, it's time to fire all appropriate sort/update events.
       if (!options.silent) {
         for (i = 0; i < toAdd.length; i++) {
           if (at != null) options.index = at + i;
@@ -912,7 +915,13 @@
           model.trigger('add', model, this, options);
         }
         if (sort || orderChanged) this.trigger('sort', this, options);
-        if (toAdd.length || toRemove.length) this.trigger('update', this, options);
+        if (toAdd.length || toRemove.length || toMerge.length) {
+          options.changes = {};
+          options.changes.added = toAdd;
+          options.changes.removed = toRemove;
+          options.changes.merged = toMerge;
+          this.trigger('update', this, options);
+        }
       }
 
       // Return the added (or merged) model (or models).
--- a/backbone.js
+++ b/backbone.js
@@ -819,7 +819,7 @@
       if (at < 0) at += this.length + 1;
       var sortable = this.comparator && (at == null) && options.sort !== false;
       var sortAttr = _.isString(this.comparator) ? this.comparator : null;
-      var toAdd = [], toRemove = [], modelMap = {};
+      var toAdd = [], toRemove = [], toMerge = [], modelMap = {};
       var add = options.add, merge = options.merge, remove = options.remove;
       var order = !sortable && add && remove ? [] : false;
       var orderChanged = false;
@@ -837,6 +837,7 @@
             attrs = this._isModel(attrs) ? attrs.attributes : attrs;
             if (options.parse) attrs = existing.parse(attrs, options);
             existing.set(attrs, options);
+            toMerge.push(existing);
             if (sortable && !sort && existing.hasChanged(sortAttr)) sort = true;
           }
           models[i] = existing;
@@ -900,8 +901,10 @@
         }
 
         // Create arrays of the models which were added and removed.
-        options.addedModels = toAdd;
-        options.removedModels = toRemove;
+        options.changes = {};
+        options.changes.added = toAdd;
+        options.changes.removed = toRemove;
+        options.changes.merged = toMerge;
 
         // Only trigger if there were models added or removed.
         if (toAdd.length || toRemove.length) {
--- a/backbone.js
+++ b/backbone.js
@@ -808,8 +808,10 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       var removed = this._removeModels(models, options);
-      options.removedModels = removed;
-      if (!options.silent && removed) this.trigger('update', this, options);
+      if (!options.silent && removed) {
+        options.removedModels = removed;
+        this.trigger('update', this, options);
+      }
       return singular ? removed[0] : removed;
     },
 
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -890,15 +890,19 @@
...
+      // Unless silenced, it's time to fire all appropriate add/update/sort events.
       if (!options.silent) {
         var addOpts = at != null ? _.clone(options) : options;
         for (var i = 0; i < toAdd.length; i++) {
           if (at != null) addOpts.index = at + i;
           (model = toAdd[i]).trigger('add', model, this, addOpts);
         }
+        if (toAdd.length || previousModels.length) {
+          if (toAdd.length) options.addedModels = toAdd;
+          if (previousModels.length) options.previousModels = previousModels;
```
megawac: is `previousModels` the right term or should it be removed models?

backbone.js
```diff
@@ -867,7 +867,7 @@
         for (var i = 0; i < this.length; i++) {
           if (!modelMap[(model = this.models[i]).cid]) toRemove.push(model);
         }
-        if (toRemove.length) this._removeModels(toRemove, options);
+        if (toRemove.length) previousModels = this._removeModels(toRemove, options);
```
jridgewell: Everything in `toRemove` has already been [verified](https://github.com/jashkenas/backbone/pull/3711/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR868) from the current models, shouldn't need to grab `#_removeModels` return.

megawac:
:+1:

elgubenis:
i would also like to include mergedModels but i think that wont be possible (see #3708)

backbone.js
elgubenis: i thought it should be removedModels, but i saw that .reset returns previousModels aswel, so i think keeping both the same would not be wrong.

jridgewell: Mind calling these `options.add` and `options.remove`? It mirrors the `add` and `remove` event names.

backbone.js
elgubenis: so should it previosModels equal to toRemove then?

backbone.js
megawac: I think past tense works better here, `added` & `removed`?

jridgewell:
Mind adding the removed models to the `update` event in `#remove`?

elgubenis:
sure

backbone.js
jridgewell: Yup.

backbone.js
jridgewell: I'm pro past tense, but nothing else in Backbone's public API uses it. @jashkenas, @braddunbar, or @akre54 want to weigh in?

jridgewell:
Looks good to me, the only thing I want feedback on is the `addedModels`/`removedModels` naming.

backbone.js
elgubenis: options.add and options.remove are already set, because they are the options passed to the .set

elgubenis:
yeah, i didnt like it either, but since .reset returns previousModels, i thought they should be named alike, well lets wait for the feedback.

elgubenis:
should i write some tests to see that the events in .remove return the removed models which it should return? as well as the removed/previous and added ones for .set? or is that not necessary

jridgewell:
> should i write some tests to see that the events in .remove return the removed models which it should return? as well as the removed/previous and added ones for .set?

:+1:

test/collection.js
```diff
@@ -1671,4 +1671,98 @@
     collection.invoke('method', 1, 2, 3);
   });
 
+  test("#3711 - remove's `update` event returns one removed model", function() {
+    var model = new Backbone.Model({id: 1, title: 'First Post'});
+    var collection = new Backbone.Collection([model]);
+    collection.on('update', function(context, options) {
+      if (options.removedModels[0] !== model) {
+        ok(false);
```
akre54: Try just `ok(options.removedModels[0] === model)` or even better, `strictEqual(options.removedModels[0], model)`

elgubenis: like that? ```js !strictEqual(options.removedModels[0], model) ``` i guess?

megawac: notStrictEqual()

backbone.js
```diff
@@ -890,15 +891,19 @@
...
+      // Unless silenced, it's time to fire all appropriate add/update/sort events.
       if (!options.silent) {
         var addOpts = at != null ? _.clone(options) : options;
         for (var i = 0; i < toAdd.length; i++) {
           if (at != null) addOpts.index = at + i;
           (model = toAdd[i]).trigger('add', model, this, addOpts);
         }
+        if (toAdd.length || toRemove.length) {
+          if (toAdd.length) options.addedModels = toAdd;
+          if (toRemove.length) options.previousModels = toRemove;
```
megawac: why not just always set `addedModels` and `previousModels` on options and trigger updates only if theres a `length` change. Seems mostly harmless (actually better IMO as the user won't have to check if `toAdd` is an array or not)

elgubenis: i saw the same behavior in the set method, previousModels is only set when there are previousModels, so i just copied that, normally i would also be in favor of having it all the time, but i didnt want to change how the other methods do work.

jridgewell: The only other `previousModels` always attaches. I think attaching an empty array here is fine as well.

Also, can we rename this `removedModels`? `previousModels` makes sense in `#reset` where every model is "previous", but here only a subset are being removed.

elgubenis: ok

elgubenis:
sorry, wrote the test a little too late

jridgewell:
LGTM.

jashkenas:
Naming isn't quite there yet — but more importantly without some sort of `mergedModels`, this change feels incomplete. Having the full bookkeeping of changes to models isn't possible?

jridgewell:
We can add a `toMerge` [here](https://github.com/jashkenas/backbone/blob/master/backbone.js#L833-L834), and push to it [here](https://github.com/jashkenas/backbone/blob/master/backbone.js#L855-L860).

jashkenas:
Alright. And there isn't a big perf cost from the extra bookkeeping?

Naming wise `added`, `removed`, `merged` sounds good. Perhaps even under a sub-object if necessary.

jridgewell:
> Alright. And there isn't a big perf cost from the extra bookkeeping?

With everything else this method is doing, I don't image one.

elgubenis:
hi, im working on the PR, how could the sub-object be named? changes, status, updates, move(s)/ments, details?

elgubenis:
i will dub it 'changes' for now if thats ok

jridgewell:
@elgubenis: You'll have to rebase off of the current master for this to merge cleanly.

elgubenis:
will write some tests for the merging

jridgewell:
LGTM.

jridgewell:
@elgubenis: Mind rebasing? I'll merge this afterwards.

Mordentware:
Just wanted to add that I was after this exact functionality, cheers guys for implementing. Any chance of getting the conflicts sorted out?

elgubenis:
hi, will do.

backbone.js
```diff
@@ -912,7 +917,13 @@
           model.trigger('add', model, this, options);
         }
         if (sort || orderChanged) this.trigger('sort', this, options);
-        if (toAdd.length || toRemove.length) this.trigger('update', this, options);
+        if (toAdd.length || toRemove.length || toMerge.length) {
+          options.changes = {};
```
akre54: minor style nit:

```js
options.changes = {
  added: toAdd,
  removed: toRemove,
  merged: toMerge
}
```

backbone.js
```diff
@@ -808,7 +808,10 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       var removed = this._removeModels(models, options);
-      if (!options.silent && removed) this.trigger('update', this, options);
+      if (!options.silent && removed) {
+        options.removedModels = removed;
```
akre54: Why is this one `options.removedModels` but below it's `options.changes.removed`?

megawac:
@elgubenis mind addressing @akre54 feedback and we'll merge? Thanks

PR #3713, opened 2015-07-14
Collection set: when one model's cid == another model's id

raymond-lam:
Given the following code:

```javascript
var model1 = new Backbone.Model({
  id: 'foo'
});
var model2 = new Backbone.Model({
  id: model1.cid
});

var collection = new Backbone.Collection([model1]);
collection.set([model1, model2]);
```

The result is that `collection` contains only `model1`. Also, oddly, `model1.id` is changed to its `cid`.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -818,7 +818,7 @@
       if (at < 0) at += this.length + 1;
       var sortable = this.comparator && (at == null) && options.sort !== false;
       var sortAttr = _.isString(this.comparator) ? this.comparator : null;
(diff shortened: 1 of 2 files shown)


akre54:
Thanks but no thanks. If your model ids are colliding with your cids (i.e. they match `/^c\d+$/`) then you should change the [`cidPrefix`](https://github.com/jashkenas/backbone/blob/1.2.1/backbone.js#L403-L405) in your Model.

PR #3719, opened 2015-07-17
Add docs for View#events

DannyDelott:
Predefining the **events** hash in a Backbone View is pretty common, and it's not super obvious that **delegateEvents** is where one should look for information about the **events** hash itself.

For instance, Backbone calls delegateEvents before it runs the View's initialize method. If the user explicitly defines the **events** hash in the View itself, then there is a chance that the user never has to explicitly call delegateEvents at all. 

I think it stands to reason that the **events** hash property get its own section in the docs.

**1. Edit:** Cleaned up whitespace, wrapped long lines, corrected wording.
**2. Edit:** Used better wording suggested by @akre54.
**3. Edit:** Refactor example View code as suggested by @akre54 and @megawac.

screen shot 2015-07-19 at 1 58 34 pm

### DIFF
--- a/index.html
+++ b/index.html
@@ -450,6 +450,7 @@
       <li>– <a href="#View-template">template</a></li>
       <li>– <a href="#View-render">render</a></li>
       <li>– <a href="#View-remove">remove</a></li>
+      <li>– <a href="#View-events">events</a></li>
       <li>– <a href="#View-delegateEvents">delegateEvents</a></li>
       <li>– <a href="#View-undelegateEvents">undelegateEvents</a></li>
     </ul>
@@ -2930,6 +2931,36 @@ var Bookmark = Backbone.View.extend({
       events that the view has <a href="#Events-listenTo">listenTo</a>'d.
     </p>
 
+    <p id="View-events">
+      <b class="header">events</b><code>view.events or view.events()</code>
+      <br />
+      The <b>events</b> hash (or method) can be used to specify a set of DOM
+      events that will be bound to methods on your View
+      through <a href="#View-delegateEvents">delegateEvents</a>.
+    </p>
+
+    <p>Backbone will automatically attach the event listeners at instantiation
+    time, right before invoking <a href="#View-constructor">initialize</a>.
+    </p>
+
+<pre>
+var InputView = Backbone.View.extend({
+
+  tagName: 'input',
+
+  events: {
+    "keydown" : "keyAction",
+  },
+
+  render: function(){ ... },
+
+  keyAction: function(e) {
+    var isEnterKey = e.which === 13;
+    if(isEnterKey){ this.collection.addEntry(this.$el.val()); }
+  }
+});
+</pre>
+
     <p id="View-delegateEvents">
       <b class="header">delegateEvents</b><code>delegateEvents([events])</code>
       <br />
@@ -4264,7 +4295,7 @@ ActiveRecord::Base.include_root_in_json = false
         in 1.2.0.
       </li>
     </ul>
-  
+
     <b class="header">1.2.0</b> &mdash; <small><i>May 13, 2015</i></small>
     &mdash; <a href="https://github.com/jashkenas/backbone/compare/1.1.2...1.2.0">Diff</a>
     &mdash; <a href="https://cdn.rawgit.com/jashkenas/backbone/1.2.0/index.html">Docs</a>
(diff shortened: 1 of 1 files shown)


rjatkinson2:
It's definitely useful to have this section on the top-level rather than needing to dive into delegateEvents to find it.

index.html
```diff
@@ -2930,6 +2931,34 @@ <h2 id="View">Backbone.View</h2>
       events that the view has <a href="#Events-listenTo">listenTo</a>'d.
     </p>
 
+    <p id="View-events">
+      <b class="header">events</b><code>view.events or view.events()</code>
+      <br />
+      The <b>events</b> hash (or function) can be used to specify DOM event listeners for your View. By explicitly defining the <b>events</b> property of your View, Backbone will automatically attach the event listeners at instantiation time, since it attempts to <a href="#View-delegateEvents">delegateEvents</a> right before invoking the <a href="#View-constructor">constructor/initialize</a>.
```
megawac: Can you wrap this text

DannyDelott:
Text has been wrapped to new lines. Good catch.

cjnething:
Wow this definitely needs clarification in the docs, please merge this

index.html
```diff
@@ -2930,6 +2931,39 @@ <h2 id="View">Backbone.View</h2>
...
 
+    <p id="View-events">
+      <b class="header">events</b><code>view.events or view.events()</code>
+      <br />
+      The <b>events</b> hash (or function) can be used to specify DOM event 
+      listeners for your View. By explicitly defining the <b>events</b> property 
+      of your View, Backbone will automatically attach the event listeners at 
+      instantiation time, since it attempts to 
+      <a href="#View-delegateEvents">delegateEvents</a> right before invoking the
+      <a href="#View-constructor">constructor/initialize</a>.
```
jridgewell: Small nitpick: it's called in the `#constructor`, but before `#initialize`. Mind changing this to just say "initialize"?

jridgewell:
How did we not have this before?

index.html
DannyDelott: Good catch. Fixed.

DannyDelott:
Corrected the anchor link text **constructor/initialize** to just **initialize**.

index.html
```diff
@@ -2930,6 +2931,40 @@ <h2 id="View">Backbone.View</h2>
       events that the view has <a href="#Events-listenTo">listenTo</a>'d.
     </p>
 
+    <p id="View-events">
+      <b class="header">events</b><code>view.events or view.events()</code>
+      <br />
+      The <b>events</b> hash (or function) can be used to specify DOM event listeners
```
akre54: "or function" => "or method"

index.html
```diff
@@ -2930,6 +2931,40 @@ <h2 id="View">Backbone.View</h2>
...
+      for your View. By explicitly defining the <b>events</b> property of your View,
+      Backbone will automatically attach the event listeners at instantiation time,
+      since it attempts to <a href="#View-delegateEvents">delegateEvents</a> right
```
akre54: This line are unnecessary. How about "Backbone will automatically attach the event listeners at instantiation time, right before invoking initialize", and move the `delegateEvents` reference earlier.

index.html
```diff
@@ -2930,6 +2931,40 @@ <h2 id="View">Backbone.View</h2>
...
+<pre>
+var InputView = Backbone.View.extend({
+
+  tagName: 'input',
+
+  events: {
+    "keydown" : "keyAction",
+  },
+
+  initialize: function(){ ... },
```
akre54: remove

index.html
```diff
@@ -2930,6 +2931,40 @@ <h2 id="View">Backbone.View</h2>
...
+
+  render: function(){ ... },
```
akre54: remove

index.html
akre54: "can be used to specify DOM event listeners for your View" => "can be used to declaratively specify the DOM events that will be bound to methods on your View with `delegateEvents`"

akre54:
Looks good, thanks. See inline comments and we'll merge.

index.html
megawac: I agree on `initialize` but I think `render` makes sense to leave in

index.html
DannyDelott: These are all great, quick question. Can you be a little more specific with which lines to delete here? Are you advocating for the removal of the tagName property, the events hash, initialize, and even the `var InputView ...` declaration?

I deleted the text about it also being a function that returns an events hash.

akre54: No just the initialize method from the example. The example should show only relevant info

DannyDelott:
All changes made. Thanks for handling this!

akre54:
Sorry one last thing -- squash down to a single commit (using rebase) and we'll get this in today. Thanks!

DannyDelott:
Squashed!

PR #3725, opened 2015-07-22
update docs according to issue #3694

### LINKED ISSUE #3694: docs for view.remove()
jeremy

IMO

http://backbonejs.org/#View-remove

this absolutely positively needs to read:

```
remove - view.remove()
Removes a view from the DOM, *removes/deletes the DOM element*, and calls stopListening to remove any bound events that the view has listenTo'd. 
```

as it is now, it doesn't make it clear that the el goes missing after ```remove()```

this call

this.$el.remove();

is hiding inside remove() and it's dangerous

one thing you could do is pass a boolean to remove() and decide whether or not to delete the DOM element

trizko:
Fixes #3694

### DIFF
--- a/index.html
+++ b/index.html
@@ -2926,7 +2926,7 @@ var Bookmark = Backbone.View.extend({
     <p id="View-remove">
       <b class="header">remove</b><code>view.remove()</code>
       <br />
-      Removes a view from the DOM, and calls
+      Removes a view and it's <tt>el</tt> from the DOM, and calls
       <a href="#Events-stopListening">stopListening</a> to remove any bound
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -2926,7 +2926,7 @@ <h2 id="View">Backbone.View</h2>
     <p id="View-remove">
       <b class="header">remove</b><code>view.remove()</code>
       <br />
-      Removes a view from the DOM, and calls
+      Removes a view and it's <tt>el</tt> from the DOM, and calls
```
akre54: Should be `its`

trizko: oops

PR #3729, opened 2015-07-24
v2: Can of Promises

au-phiware:
I'm a big fan of Backbone and of Promises, this is my take on combining the two.

I understand that this may be a can of worms and I have changed many, many functions and it may not be your cup of tea. For this reason I have also made sure all the tests pass (with changes, of course) and I have tried to be backwards compatible (the Todos app still works). I haven't added any new tests yet but I would like to, as use cases arise and as I start to use this code in my own projects.

I have many use cases in mind and have already started with my own major changes: for instances Router#execute will catch errors and trigger an error event (on the router). My intention is that route actions should return the promise(s) from other methods (like render), this will mean that no error gets swallowed or forgotten about.

I'm making this PR, ignorant of the opinions of others (I've only just realised that there are already discussions about this) but keen to learn those opinions and to get some feedback.

I have made some assumptions:
- `Promise`, as per ES6, is a global object. In my view it's very easy to include the standard [polyfill](https://www.promisejs.org/polyfills/promise-7.0.1.js).
- `jQuery.Deferred` is not `Promise` and needs to be avoided.
- Promises are useful outside of database interaction.

That last point is important to me. I am currently working on a big project that uses requirejs and pouchdb/couchdb. We use promises to orchestrate module dependencies as well as data from the server, whereby a Promise is created inside initialize and `then` is called inside render.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1511,8 +1511,9 @@
         try {
           var route = callback.apply(this, args);
           if (isPromise(route)) {
-            route['catch'](error);
+            return route['catch'](error);
           }
+          return route;
         } catch(e) {
           error(e);
         }
--- a/backbone.js
+++ b/backbone.js
@@ -1825,8 +1825,8 @@
 
     // Add a route to be tested when the fragment changes. Routes added later
     // may override previous routes.
-    route: function(route, callback) {
-      this.handlers.unshift({route: route, callback: callback});
+    route: function(route, callback, meta) {
+      this.handlers.unshift(_.defaults({route: route, callback: callback}, meta));
     },
 
     // Checks the current URL to see if it has changed, and if it has,
--- a/backbone.js
+++ b/backbone.js
@@ -433,7 +433,7 @@
     // Proxy `Backbone.sync` by default -- but override this if you need
     // custom syncing semantics for *this* particular model.
     sync: function() {
-      return Backbone.sync.apply(this, arguments);
+      return Promise.resolve(Backbone.sync.apply(this, arguments));
     },
 
     // Get the value of an attribute.
@@ -587,15 +587,13 @@
     fetch: function(options) {
       options = _.extend({parse: true}, options);
       var model = this;
-      var success = options.success;
-      options.success = function(resp) {
-        var serverAttrs = options.parse ? model.parse(resp, options) : resp;
-        if (!model.set(serverAttrs, options)) return false;
-        if (success) success.call(options.context, model, resp, options);
-        model.trigger('sync', model, resp, options);
-      };
-      wrapError(this, options);
-      return this.sync('read', this, options);
+      return wrapError(this.sync('read', this, options), this, options)
+        .then(function(resp) {
+          var serverAttrs = options.parse ? model.parse(resp, options) : resp;
+          if (model.set(serverAttrs, options))
+            model.trigger('sync', model, resp, options);
+          return model;
+        });
     },
 
     // Set a hash of model attributes, and sync the model to the server.
@@ -618,38 +616,35 @@
       // `set(attr).save(null, opts)` with validation. Otherwise, check if
       // the model will be valid when the attributes, if any, are set.
       if (attrs && !wait) {
-        if (!this.set(attrs, options)) return false;
+        if (!this.set(attrs, options)) return Promise.reject(this);
       } else {
-        if (!this._validate(attrs, options)) return false;
+        if (!this._validate(attrs, options)) return Promise.reject(this);
       }
 
       // After a successful server-side save, the client is (optionally)
       // updated with the server-side state.
       var model = this;
-      var success = options.success;
       var attributes = this.attributes;
-      options.success = function(resp) {
-        // Ensure attributes are restored during synchronous saves.
-        model.attributes = attributes;
-        var serverAttrs = options.parse ? model.parse(resp, options) : resp;
-        if (wait) serverAttrs = _.extend({}, attrs, serverAttrs);
-        if (serverAttrs && !model.set(serverAttrs, options)) return false;
-        if (success) success.call(options.context, model, resp, options);
-        model.trigger('sync', model, resp, options);
-      };
-      wrapError(this, options);
 
       // Set temporary attributes if `{wait: true}` to properly find new ids.
       if (attrs && wait) this.attributes = _.extend({}, attributes, attrs);
 
       var method = this.isNew() ? 'create' : (options.patch ? 'patch' : 'update');
       if (method === 'patch' && !options.attrs) options.attrs = attrs;
-      var xhr = this.sync(method, this, options);
+      var promise = wrapError(this.sync(method, this, options), this, options);
 
       // Restore attributes.
       this.attributes = attributes;
 
-      return xhr;
+      return promise.then(function(resp) {
+        // Ensure attributes are restored during synchronous saves.
+        model.attributes = attributes;
+        var serverAttrs = options.parse ? model.parse(resp, options) : resp;
+        if (wait) serverAttrs = _.extend({}, attrs, serverAttrs);
+        if (!serverAttrs || model.set(serverAttrs, options))
+          model.trigger('sync', model, resp, options);
+        return model;
+      });
     },
 
     // Destroy this model on the server if it was already persisted.
@@ -658,7 +653,6 @@
     destroy: function(options) {
       options = options ? _.clone(options) : {};
       var model = this;
-      var success = options.success;
       var wait = options.wait;
 
       var destroy = function() {
@@ -666,21 +660,15 @@
         model.trigger('destroy', model, model.collection, options);
       };
 
-      options.success = function(resp) {
-        if (wait) destroy();
-        if (success) success.call(options.context, model, resp, options);
-        if (!model.isNew()) model.trigger('sync', model, resp, options);
-      };
-
-      var xhr = false;
-      if (this.isNew()) {
-        _.defer(options.success);
-      } else {
-        wrapError(this, options);
-        xhr = this.sync('delete', this, options);
-      }
+      var isNew = model.isNew();
+      var promise = isNew ? Promise.resolve(model.attributes) :
+        wrapError(this.sync('delete', this, options), this, options);
       if (!wait) destroy();
-      return xhr;
+      return promise.then(function(resp) {
+        if (wait) destroy();
+        if (!isNew) model.trigger('sync', model, resp, options);
+        return model;
+      });
     },
 
     // Default URL for the model's representation on the server -- if you're
@@ -783,7 +771,7 @@
 
     // Proxy `Backbone.sync` by default.
     sync: function() {
-      return Backbone.sync.apply(this, arguments);
+      return Promise.resolve(Backbone.sync.apply(this, arguments));
     },
 
     // Add a model, or list of models to the set. `models` may be Backbone
@@ -1001,16 +989,14 @@
     // data will be passed through the `reset` method instead of `set`.
     fetch: function(options) {
       options = _.extend({parse: true}, options);
-      var success = options.success;
       var collection = this;
-      options.success = function(resp) {
-        var method = options.reset ? 'reset' : 'set';
-        collection[method](resp, options);
-        if (success) success.call(options.context, collection, resp, options);
-        collection.trigger('sync', collection, resp, options);
-      };
-      wrapError(this, options);
-      return this.sync('read', this, options);
+      return wrapError(this.sync('read', this, options), this, options)
+        .then(function(resp) {
+          var method = options.reset ? 'reset' : 'set';
+          collection[method](resp, options);
+          collection.trigger('sync', collection, resp, options);
+          return collection;
+        });
     },
 
     // Create a new instance of a model in this collection. Add the model to the
@@ -1023,13 +1009,11 @@
       if (!model) return false;
       if (!wait) this.add(model, options);
       var collection = this;
-      var success = options.success;
-      options.success = function(model, resp, callbackOpts) {
-        if (wait) collection.add(model, callbackOpts);
-        if (success) success.call(callbackOpts.context, model, resp, callbackOpts);
-      };
-      model.save(null, options);
-      return model;
+      return model.save(null, options)
+        .then(function(model) {
+          if (wait) collection.add(model, options);
+          return model;
+        });
     },
 
     // **parse** converts a response into a list of models to be added to the
@@ -1381,18 +1365,10 @@
       params.processData = false;
     }
 
-    // Pass along `textStatus` and `errorThrown` from jQuery.
-    var error = options.error;
-    options.error = function(xhr, textStatus, errorThrown) {
-      options.textStatus = textStatus;
-      options.errorThrown = errorThrown;
-      if (error) error.call(options.context, xhr, textStatus, errorThrown);
-    };
-
     // Make the request, allowing the user to override any Ajax options.
-    var xhr = options.xhr = Backbone.ajax(_.extend(params, options));
-    model.trigger('request', model, xhr, options);
-    return xhr;
+    var promise = options.promise = Promise.resolve(Backbone.ajax(_.extend(params, options)));
+    model.trigger('request', model, promise, options);
+    return promise;
   };
 
   // Map from CRUD to HTTP for our default `Backbone.sync` implementation.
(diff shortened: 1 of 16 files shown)


backbone.js
```diff
@@ -613,43 +629,48 @@
 
       options = _.extend({validate: true, parse: true}, options);
       var wait = options.wait;
+      var success = options.success || _.identity;
+      var error = options.error;
+      delete options.success;
+      delete options.error;
```
megawac: Don't modify the `options` object

backbone.js
```diff
@@ -613,43 +629,48 @@
...
 
       // If we're not waiting and attributes exist, save acts as
       // `set(attr).save(null, opts)` with validation. Otherwise, check if
       // the model will be valid when the attributes, if any, are set.
       if (attrs && !wait) {
-        if (!this.set(attrs, options)) return false;
+        if (!this.set(attrs, options)) return Promise.reject(this);
```
megawac: Does it make sense to reject when the model doesn't change. I'm not sure if this is the correct behaviour

backbone.js
au-phiware: err, but the master branch does modify the `options` object already... I guess you object to the `delete` lines?

backbone.js
```diff
@@ -658,29 +679,33 @@
...
       if (!wait) destroy();
-      return xhr;
+      return promise.then(function(resp) {
+        var promise = Promise.resolve(model);
+        if (wait) destroy();
+        if (success)
+          promise = Promise.resolve(success.call(options.context, model, resp, options));
+        if (!isNew)
+          promise = promise.then(function(delivered) {
+            model.trigger('sync', model, resp, options);
```
megawac: Does `promise` need to be redefined for this (I mean why not let a race (of sorts, the `sync` one will always win in all implementations I've seen) happen between `sync` and the returned promise `resolve`

backbone.js
```diff
@@ -693,7 +718,12 @@
         urlError();
       if (this.isNew()) return base;
       var id = this.get(this.idAttribute);
-      return base.replace(/[^\/]$/, '$&/') + encodeURIComponent(id);
+      var consume = function(base) {
+        return base.replace(/[^\/]$/, '$&/') + encodeURIComponent(id);
+      };
+      if (base instanceof Promise)
+        return base.then(consume);
```
megawac: Might want to make a helper for this. Also `instanceof` is not the correct way to check for `Promise`s as they should be implementation independent

See https://github.com/then/is-promise

backbone.js
au-phiware: oh, but I'm not modifying the object that is provided in that arguments.

backbone.js
```diff
@@ -1464,7 +1560,22 @@
...
-      if (callback) callback.apply(this, args);
+      var router = this;
+      var error = function(err) {
+        router.trigger('error', err, name, args, callback);
+      };
+      if (callback) {
+        try {
+          var route = callback.apply(this, args);
+          if (route instanceof Promise) {
+            route.catch(error);
```
megawac: use `['catch']` instead of `.catch` for all cases

megawac:
I'm sympathetic to this changeset and think a movement like this may be a good way to move backbone forward (along with @jmeas' proposed router changes) in v2. So a hesitant :+1:

backbone.js
au-phiware: Yes, I admit I wasn't sure on this point... I did want to return a Promise. Thinking about it now I think `reject` is wrong.

backbone.js
au-phiware: This could be written differently, but the point is that when an existing model is destroyed then the sync event will be triggered after the success callback. Providing a hook to the developer to resolve any rejection by the sync method before the event is triggered.

backbone.js
au-phiware: Good point. I used `instanceof` because I'm assuming ES6 Promises.

backbone.js
au-phiware: Good point. *grumbles something about reserved words*

backbone.js
au-phiware: As I have defined it, Model#save will return a `Promise` that resolves to the model in the successful case. I am unclear that calling save on an unmodified model is a success case. If it isn't then the returned `Promise` should be rejected with the model (or maybe `false`?) If it is a success case then the returned `Promise` should be resolved with the model. Alternatively, the current behaviour of returning `false` could be preserved, in which case, someone like me would wrap every call to save like this:

    Promise.resolve(model.save(attrs)).then(function(saved) {
      if (saved === false) doSomethingSpecial(model);
      doSomethingExpected(saved || model);
    }, handleError);

The fourth option (as I see it) is to return `Promise.resolve(false)` which is kind of saying:
> "when save is successful, you'll get a resolved `Promise`, in the normal case it'll be fulfilled with the model and in the exceptional case it won't, it'll be fulfilled with `false`, otherwise (in the case of failures) the `Promise` will be rejected."

backbone.js
au-phiware: Sorry, I got confused, the error callback provides the hook to resolve a rejected promise from `sync` (not the success callback). The success callback provides a hook to change the object (by reference) with which the returned `Promise` is fulfilled (and/or modify the model, if it wants) before the sync event is triggered. Event#trigger doesn't return anything particularly useful to the chain of promises so this is the only chance that the destroy method can participant in the promise chain after the call to sync and before sync is triggered.

Thinking about it some more, you could argue that the sync event doesn't care how or what the promise is fulfilled (i.e. `delivered` is not passed to Events#trigger), so what does it matter...

au-phiware: Actually, that's entirely your point isn't it? Sorry, I'm a bit slow...

So, to that point: the triggering of sync must be present in the returned chained of promises so that any errors can be caught by the caller.

au-phiware: This is super important to me, the other day I got bit by an error that was being thrown by a jquery plugin inside a [stickit](https://github.com/NYTimes/backbone.stickit) method that is called by a sync trigger. The error was swallowed deep inside a promise (of my own making) and was very hard to trace. If I had this code the error object would have been propagated up to the user and the user would have probably/hopefully been presented with an error instead of half a page of broken content. Of course the outcome is that I had to work around the buggy jquery plugin (there's no escape from that!) but there would have been a better developer experience at the very least.

So, I wrote a test: 403bc44a7b89c70a49d4c6f12675b06dbc0932e6

backbone.js
au-phiware: Is 0d4fe5b23f1023df17b7e6b13c7f44639f64de41 what you had in mind?

Also, shall go ahead and drop the polyfill under `test/vendor`?

backbone.js
megawac: My point was we can avoid returning a new promise and instead just return

```js
if (!isNew) {
          promise.then(function(delivered) {
            model.trigger('sync', model, resp, options);
         });
}
```

Note not redefining `promise = promise.then`

backbone.js
```diff
@@ -398,10 +401,20 @@
...
+      model.set(attrs, options);
+      model.changed = {};
+      var init = model.initialize.apply(model, args);
+      if (isPromise(init))
+        return init.then(_.constant(model));
+      return model;
+    };
+    if (isPromise(defaults))
+      return defaults.then(consume);
+    return consume(defaults);
```
megawac: I'm not sure about returning a promise from `initialize`

examples/backbone.localStorage.js
```diff
@@ -114,54 +114,43 @@ _.extend(Backbone.LocalStorage.prototype, {
...
-        break;
-      case "update":
-        resp = store.update(model);
-        break;
-      case "delete":
-        resp = store.destroy(model);
-        break;
+  var resp, errorMessage;
+
+  return new Promise(function(resolve, reject) {
```
megawac: Did this changeset break `backbone.localStorage`? For this to be accepted I don't think we should be breaking the major popular plugins in the ecosytem.

backbone.js
au-phiware: The resulting promise of that expression needs to be returned.

backbone.js
au-phiware: I went there

examples/backbone.localStorage.js
au-phiware: Nope, it doesn't break the existing backbone.localStorage. I changed it to use promises, that's all.

au-phiware:
I get 7 failed tests when I use either this branch or the jashkenas/master branch.

backbone.js
```diff
@@ -14,24 +14,24 @@
...
+    factory(root, exports, _, $, Promise);
 
   // Finally, as a browser global.
   } else {
-    root.Backbone = factory(root, {}, root._, (root.jQuery || root.Zepto || root.ender || root.$));
+    root.Backbone = factory(root, {}, root._, (root.jQuery || root.Zepto || root.ender || root.$), root.Promise);
   }
 
-}(function(root, Backbone, _, $) {
+}(function(root, Backbone, _, $, Promise) {
```
jmeas: I'm not a fan of requiring a particular Promise implementation. Many people will end up requiring both the library you have chosen here *and* whatever Promise library they actually use.

In my libs I prefer doing:

```js
if (typeof Promise !== 'undefined') {
  Backbone.Promise = Promise;
}
```

That way, if someone's using native Promises, and they wish to continue using native promises, they can. Or if they'd rather use RSVP, they can do that, too. And so on.

An extra install instruction – telling people to configure `Backbone.Promise` – is worth the added benefit of flexibility, I think.

backbone.js
```diff
@@ -1478,11 +1592,19 @@
...
+        router.routes = routes;
+        routes = _.keys(routes);
+        var route;
+        while ((route = routes.pop()) != null) {
+          router.route(route, router.routes[route]);
+        }
+      };
+      if (isPromise(routes))
+        return routes.then(consume);
+      consume(routes);
```
jmeas: Unnecessary imo...

backbone.js
```diff
@@ -1464,7 +1563,22 @@
...
+          var route = callback.apply(this, args);
+          if (isPromise(route)) {
+            route['catch'](error);
+          }
+        } catch(e) {
+          error(e);
+        }
+      } else {
+        error();
+      }
```
jmeas: You may want to add additional tests for this functionality.

Fwiw (although it may be irrelevant to this PR), if all of my desired changes to the router land then this method will no longer exist.

backbone.js
au-phiware: Yes, I agree... I felt dirty trying to pick an implementation from npm, bower, etc.

backbone.js
au-phiware: Are you talking about jmeas/backbone.base-router? I should like to check it out :)

megawac:
Also, there is significant overlap with https://github.com/jashkenas/backbone/pull/3651

backbone.js
jmeas: Ah, sorry, should have linked the PR. I was referring to https://github.com/jashkenas/backbone/pull/3660. That PR is heavily inspired by the ideas in BaseRouter :)

megawac: https://github.com/jashkenas/backbone/pull/3660

backbone.js
akre54: Agreed. Let's stick with the global Promise, or throw an error if one isn't defined (see @jridgewell's branch).

backbone.js
akre54: :-1: on this. Initialize is always synchronous, therefore this will always be resolved after you've created the instance -- in other words, exactly where it was resolved before. Revert this please.

backbone.js
akre54: Why do you care to delete these properties? It shouldn't matter if they stick around, no?

backbone.js
au-phiware: I have real world use cases where initialize is not synchronous. Of course, the constructor will not return a `Promise` if the user never returns a `Promise`... but I can tell that you feel strongly about this and I'm happy to agree to disagree.

akre54: Would love to hear those use cases if you feel like sharing.

On Mon, Jul 27, 2015, 9:06 AM Corin Lawson <notifications@github.com> wrote:

> In backbone.js
> <https://github.com/jashkenas/backbone/pull/3729#discussion_r35532739>:
>
> > -    this.changed = {};
> > -    this.initialize.apply(this, arguments);
> > +    var model = this, args = slice.call(arguments);
> > +    var defaults = _.result(this, 'defaults');
> > +    var consume = function(defaults) {
> > +      attrs = _.defaults({}, attrs, defaults);
> > +      model.set(attrs, options);
> > +      model.changed = {};
> > +      var init = model.initialize.apply(model, args);
> > +      if (isPromise(init))
> > +        return init.then(_.constant(model));
> > +      return model;
> > +    };
> > +    if (isPromise(defaults))
> > +      return defaults.then(consume);
> > +    return consume(defaults);
>
> I have real world use cases where initialize is not synchronous. Of
> course, the constructor will not return a Promise if the user never
> returns a Promise... but I can tell that you feel strongly about this and
> I'm happy to agree to disagree.
>
> —
> <https://github.com/jashkenas/backbone/pull/3729/files#r35532739>.
>
-- 

Adam K (mobile)

megawac: I think I'd rather let the user handle this themselves -- either with a custom wrapper or waiting for the promise to resolve

```js
defaults.then(attrs => new Backbone.Model(attrs)
.then(model => {

});
```

au-phiware: I have reverted the commit that affects the constructors, and I hinted at a use case in the last sentence of my OP. I would like to create an example to illustrate why I started on this course in the first place, but that will take time to prune back some real world code into something bite-sited :)

backbone.js
au-phiware: I don't want them called by ajax (i.e. jQuery or whatever it might be), because I want them to be apart of the promise chain.

backbone.js
au-phiware: I removed the dependency on any one promise implementation but still attempt to load it via require (or fallback to global)... Do you see any problems with that?

backbone.js
au-phiware: I've just seen @jridgewell's take on this: [backbone-promise](https://github.com/jridgewell/backbone/blob/6de1cbbf0c13bc26a07a389ac2c955e0bf7b58bc/backbone.js#L604). And I think I've been utterly confused on this... For some reason I had it in my head that `false` meant that there was no change to the attributes... :-$

backbone.js
akre54: Yes... if your project didn't include the `promise` library this bit of code would fail. It's much better to require the dev to provide a Promise polyfill or set one on `Backbone.Promise` directly instead.

jmeas: > but still attempt to load it via require (or fallback to global)... Do you see any problems with that?

Yeah, that is still going to be a problem for people. You should remove everything related to promises from both `package.json` and from the UMD wrapper.

akre54:
I think this pull really showcases some of the downsides of Promises (and asynchronicity in your code in general) in that once you use them in a few places, you practically have to make your entire API asynchronous.

I see very little reason for `defaults` or `render`, to give two examples, to be asynchronous. That's not the right place for network or user-facing code to exist. You're contorting Backbone's API &mdash; and by extension your code &mdash; to fit a small assortment of problems that are better handled elsewhere.

I'm fully in support of wrapping asynchronous calls (sync is the biggest one) in Promises, however.

backbone.js
au-phiware: Oh, I thought that the UMD wrapper would simply receive `undefined`... no probs

au-phiware:
Ha! Perhaps I should have added: "WARNING: May cause Scoliosis"

But on a serious note, I don't like the `isPromise` checks but I think there's power in giving the option to have a value, a function or a Promise at those points. IMO, writing with Promises demands a different way of thinking. Remember that a Promise does *not* represent an async call, instead it is simply a dependency on an async call, which could have been fulfilled at any poiny in the past. So it is not correct to assume that whereever there is a Promise there will be a network call or something else that hasn't happen.

So I agree that `defaults` is not the right place to be accessing the network, etc. But I can't dictate that it can't be time dependent on something

au-phiware:
Sorry, that close was a slip of the thumb... Please reopen if you are still interested in this PR.

akre54:
> Scoliosis

Ha. @paulmillr made a fork of Backbone a while back called Scoliosis (since renamed Exoskeleton). Maybe we could implement it there first.

>  Remember that a Promise does not represent an async call, instead it is simply a dependency on an async call, which could have been fulfilled at any poiny in the past.

Right of course. But it starts to get ugly fast when you mix sync and async code using promises. In my experience they tend to take over your entire application when really you should've been writing it reactively in the first place. 

> I don't like the isPromise checks but I think there's power in giving the option to have a value, a function or a Promise at those points.

There's power but there's also a great deal of sloppiness. You know what Ben Parker (and Teddy Roosevelt) said about power and responsibility...

> So I agree that defaults is not the right place to be accessing the network, etc. But I can't dictate that it can't be time dependent on something else.

Can you give an example?

godlark:
I think if we want to return a promise it would be good to pass parsed response to then's callback. Read more at https://github.com/jashkenas/backbone/pull/3779

prantlf:
The `sync` method in Backbone 1.x returns  promise using the interface of jQuery.Deferred. Be careful with switching to Promise/A interface in the result of this method. It would be a breaking change. You should schedule it for Backbone 2.x.

PR #3732, opened 2015-07-26
Use `_.bind` instead of `context` param of `_.sortBy`.

jdalton:
Since lodash v4 [will be removing](https://github.com/underdash/underdash/issues/4#issuecomment-118974230) the `context` param of methods Backbone can avoid a minor compat issue by deferring to `_.bind` for both cases in `sort`.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -977,16 +977,19 @@
     // normal circumstances, as the set will maintain sort order as each item
     // is added.
     sort: function(options) {
-      if (!this.comparator) throw new Error('Cannot sort a set without a comparator');
+      var comparator = this.comparator;
+      if (!comparator) throw new Error('Cannot sort a set without a comparator');
       options || (options = {});
 
-      // Run sort based on type of `comparator`.
-      if (_.isString(this.comparator) || this.comparator.length === 1) {
-        this.models = this.sortBy(this.comparator, this);
-      } else {
-        this.models.sort(_.bind(this.comparator, this));
-      }
+      var length = comparator.length;
+      if (_.isFunction(comparator)) comparator = _.bind(comparator, this);
 
+      // Run sort based on type of `comparator`.
+      if (length === 1 || _.isString(comparator)) {
+        this.models = this.sortBy(comparator);
+      } else {
+        this.models.sort(comparator);
+      }
       if (!options.silent) this.trigger('sort', this, options);
       return this;
     },


backbone.js
```diff
@@ -979,14 +979,14 @@
     sort: function(options) {
       if (!this.comparator) throw new Error('Cannot sort a set without a comparator');
       options || (options = {});
+      var comparator = _.isFunction(this.comparator) ? _.bind(this.comparator, this) : this.comparator;
 
       // Run sort based on type of `comparator`.
       if (_.isString(this.comparator) || this.comparator.length === 1) {
```
megawac: Should this be changed to `_.isString(comparator)`

jdalton: It doesn't matter functionally. I kept it `this.comparator` because the `length` check to the right still requires `this.comparator.length`.

backbone.js
```diff
@@ -979,14 +979,14 @@
     sort: function(options) {
       if (!this.comparator) throw new Error('Cannot sort a set without a comparator');
       options || (options = {});
+      var comparator = _.isFunction(this.comparator) ? _.bind(this.comparator, this) : this.comparator;
```
megawac: Would this be better as `_.iteratee(this.comparator, this)`?

backbone.js
akre54: Couldn't you hoist comparator length to above this block? It's inconsistent to refer to `this.comparator` in some places and `comparator` in others.

jdalton:
Ok I've tweaked the PR. The test fails seem unrelated. I checked out 1.2.1 and ran its `npm run-script test` and it failed phantomjs tests too.

backbone.js
jdalton: Yap, [there's another place](https://github.com/jashkenas/backbone/blob/1.2.1/backbone.js#L440) that relies on `iteratee(..., this)` but in that case I can advise users to just provide their callbacks pre-bound.

megawac: Why not just keep the context param in v4 and make it a simple proxy for bind?

jdalton: Because `iteratee` methods no longer have a `context` param punting for devs to just pass in iteratees pre-bound with `_.bind` or `Function#bind`.

megawac: Sure, but I think keeping this layer is useful for plugins/extensions built
on top of lodash that don't wish to break compatibility

jdalton: Passing pre-bound functions will work with Underscore too so that's the compat path I'll point to.

megawac: Sure but that sort of defeats the purpose of `iteratee` which is supposed to sugar a function. It forces devs to do `_.isFunction(iter) ? _.bind(iter, context) : _.iterate(iter);` or for them to override `iteratee`. Seems like a low hanger for a common case

jdalton: I'm *ok* with not handling the `context` case for `iteratee` in lodash since no `iteratee` based function in lodash accepts a `context` param. Users are free to customize `iteratee` to accept a `context` param if they want to expand it for their use.

paulfalgout:
Test'll be fixed here: https://github.com/jashkenas/backbone/pull/3743

paulfalgout:
A rebase'll fix travis

akre54:
No need for a rebase, I restarted the build.

jdalton:
:cool:

jdalton:
Ping @megawac

megawac:
I'm not happy about it but fine ;)

Issue #3734, opened 2015-07-27
Deferring updates until first save succeeds


dgsan:
What is the generally accepted method of delaying additional updates until the first save succeeds (with a model)? 

I'm trying to avoid an issue where save ends up creating additional entries server side because the first few saves all end up getting treated as `create` (POST) requests, I guess because of latency with the first save. I'll see something like POST `{first_name: "John"}`, POST `{first_name: "John", last_name: "Doe"}`, instead of the second request being a PUT for `first_name` and `last_name`, and then end up with multiple DB rows that are near identical as fields are sequentially filled in.

I tried the simple solution of deferring saves with deferred objects, but that just seems to lead to the hash sent back from the server's first save clobbering any local changes before the deferred save fires.

Thought I'd check before writing something complicated that queues/merges pending changes. Most of the related google results seem to be from several years ago (and haven't worked in tests).

tsjoberg:
The save event is an async event so without looking at your code you can try to use the models ```'sync'``` event and listen to that event prior to calling additional updates. 

http://backbonejs.org/#Events-catalog

You can also add a ```success``` callback to the save options as shown in the docs too which is called if 200 response is returned:

http://backbonejs.org/#Model-save

However you will probably need this and other functionality (ex. disabling submit button on ```request``` (or debouncing the actual save to make sure you send the latest data on multiple clicks) and re-enabling submit on successful save etc. These are just a few ideas but hopefully that helps.

akre54:
Please use the mailing list, StackOverflow, or the gitter channel for questions. Github issues are for maintaining the framework only.

That said, there are a few solutions to your problem, but the easiest, as @tsjoberg mentioned, is to [`debounce`](http://underscorejs.org/#debounce) your saves. You might also want to think about why this is happening in your code, and how to avoid it.

dgsan:
I just want to point out the the README.md says "To suggest a feature, report a bug, or general discussion: https://github.com/jashkenas/backbone/issues". I don't have any objections to using other forums, it just looked like the readme was suggesting asking stuff like this here.

akre54:
Ha, we may've outgrown that wording. It's been our policy for a few years that questions should live on a more appropriate medium. Thanks for pointing that out, I'll fix it ASAP.

PR #3735, opened 2015-07-27
Add alternative help channels to README

akre54:
Github issues are the wrong forum for general discussion and questions (see #3734). Let's direct folks to the right venues.

### DIFF
--- a/README.md
+++ b/README.md
@@ -16,8 +16,13 @@ Backbone supplies structure to JavaScript-heavy applications by providing models
 For Docs, License, Tests, pre-packed downloads, and everything else, really, see:
 http://backbonejs.org
 
-To suggest a feature, report a bug, or general discussion:
+To suggest a feature or report a bug:
 https://github.com/jashkenas/backbone/issues
+
+For questions on working with Backbone or general discussions:
+https://groups.google.com/forum/#!forum/backbonejs,
+http://stackoverflow.com/questions/tagged/backbone.js, or
+https://gitter.im/jashkenas/backbone
 
 Backbone is an open-sourced component of DocumentCloud:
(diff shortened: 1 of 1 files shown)


megawac:
Yes, definitely!

jridgewell:
:+1:

akre54:
> Should they be added to the index.html as well?

It [already exists](https://github.com/jashkenas/backbone/blob/cb47ea43401d108862b55e0f6b8806993fded1b5/index.html#L559-L566), no?

megawac:
#documentcloud on freenode -> gitter?

megawac:
And add SO?

akre54:
I tend to feel IRC is a better medium than Gitter, but the gitter channel is clearly much better maintained (mostly due to the notification emails, it seems).

megawac:
> I'm torn. I'd rather we favor open content over proprietary silos (i.e. suggest a web search over a StackOverflow)

Well the mailing list is google groups....

> I tend to feel IRC is a better medium than Gitter, but the gitter channel is clearly much better maintained 

I just like not missing messages in the channel. Don't think the irc channel has been active in a while

jridgewell:
StackOverflow, by far.

RusAlex:
IRC is important too. It may be in list next to gitter. They are not
competitors imho.

megawac:
Gitter can be accessed through IRC just FYI

On Tue, Jul 28, 2015 at 4:09 AM, Alexander Pletnev <notifications@github.com

jashkenas:
I'd prioritize IRC over Gitter ... just because, what are the odds of Gitter still being around in the same form in 5 years? But otherwise sounds great.

Issue #3720, opened 2015-07-18
Bump


jdalton:
This is a placeholder issue for @megawac or @jridgewell to hijack with checklists for the next bump.

I'm prepping for lodash v4 as part of the [Underdash proofing](https://github.com/underdash/underdash/issues/4#issuecomment-118974230) and will wait for the Backbone bump which removes `_.any` use in favor of `_.some`.

jridgewell:
### Relevant changes

- **#3680**
- **#3667**
- #3675
- #3678 
- #3646

### Things that should be considered

- **#3711**
- #3717

Bolded issue numbers are the important ones, non-bolds not so much...

Some one want to take a stab at writing the change log? Writing prose really isn't my thing. @paulfalgout?

paulfalgout:
Sure thing.  I'm all about taking on the small stuff.  I might wait on the considerations?

jridgewell:
> I might wait on the considerations?

Yup, those were just for the contribs to keep a checklist.

megawac:
We can publish a version before addressing all these, yeah?

jdalton:
Anything to do here or is it :shipit: time?

paulfalgout:
If we can pick the issues shipping, I'll go ahead and draft the log.

jdalton:
> If we can pick the issues shipping, I'll go ahead and draft the log.

Let's go with what's merged.

jridgewell:
@paulfalgout: I agree with @jdalton here. Let's just ship what we have, and we can merge the rest in the next version.

paulfalgout:
good deal.. I'll get on it shortly.. was a long weekend.

paulfalgout:
So....

jashkenas:
Bumped and published.

PR #3740, opened 2015-08-03
Draft change log for Backbone 1.2.2

paulfalgout:
For https://github.com/jashkenas/backbone/issues/3720

While this is short update, other than the underdash stuff, both additions are featurey and much less bugfixy so I'd suggest a minor point bump.

Style references:
- Listing multiple methods: https://github.com/jashkenas/backbone/blob/1.2.1/index.html#L4445
- Similar backbone event entry https://github.com/jashkenas/backbone/blob/1.2.1/index.html#L4539

Ponderings:
- Should the model-attribute-style predicate functionality be documented better elsewhere and just referenced here?
- The fairly generic underdash log entry could be removed, though it seems as if that is predominately why this release is shipping now.  If it stays, should we link out to Underdash's github?

Todo:
- [x] Add documentation for model-attribute-style
- [x] Add change log note for #3676

### DIFF
--- a/index.html
+++ b/index.html
@@ -895,11 +895,16 @@ proxy.on("all", function(eventName) {
 
 <pre>
 book.on({
-  "change:title": titleView.update,
   "change:author": authorPane.update,
+  "change:title change:subtitle": titleView.update,  
   "destroy": bookView.remove
 });
 </pre>
+    
+    <p>
+      For event map syntax to supply a <b>context</b> value for <tt>this</tt> when a 
+      callback is invoked, pass the optional second argument: <tt>model.on({ 'change': this.render }, this)</tt>
+    </p>
 
     <p id="Events-off">
       <b class="header">off</b><code>object.off([event], [callback], [context])</code><span class="alias">Alias: unbind</span>
@@ -1879,7 +1884,7 @@ alert(JSON.stringify(collection));
       <li><a href="http://underscorejs.org/#reject">reject</a></li>
       <li><a href="http://underscorejs.org/#every">every (all)</a></li>
       <li><a href="http://underscorejs.org/#some">some (any)</a></li>
-      <li><a href="http://underscorejs.org/#contains">contains (include)</a></li>
+      <li><a href="http://underscorejs.org/#contains">contains (includes)</a></li>
       <li><a href="http://underscorejs.org/#invoke">invoke</a></li>
       <li><a href="http://underscorejs.org/#max">max</a></li>
       <li><a href="http://underscorejs.org/#min">min</a></li>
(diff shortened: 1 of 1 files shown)


index.html
```diff
@@ -4272,6 +4272,24 @@ <h2 id="examples-tzigla">Tzigla</h2>
...
+    &mdash; <a href="https://cdn.rawgit.com/jashkenas/backbone/1.3.0/index.html">Docs</a>
+    <br />
+    <ul style="margin-top: 5px;">
+      <li>
+        Collection methods <tt>find</tt>, <tt>filter</tt>, <tt>reject</tt>, <tt>every</tt>, 
+        <tt>some</tt>, and <tt>partition</tt> now support a model-attributes-style predicate 
+        <tt>this.collection.reject({user: 'guybrush'})</tt>.
+      </li>
+      <li>
+        Backbone events now support multiple-event maps <tt>obj.on({'click focus': action})</tt>.
```
akre54: This was a bugfix. It was supported a few versions ago, but the events refactor caused a regression.

index.html
```diff
@@ -4272,6 +4272,24 @@ <h2 id="examples-tzigla">Tzigla</h2>
...
+      </li>
+      <li>
+        Various tweaks for Underdash proofing.
```
akre54: ??

akre54:
> Should the model-attribute-style predicate functionality be documented better elsewhere and just referenced here?

I would do both. What you have here is fine, just add to the Underscore section above. 

> The fairly generic underdash log entry could be removed, though it seems as if that is predominately why this release is shipping now.

According to whom?

index.html
megawac: I'd say `underdash` and lodash v4 future proofing (link to repos)

index.html
paulfalgout: Ah so a similar issue from `1.1.2` had this:
> This was previously an undocumented behavior from 1.1.2 removed in 1.2.0.

Would that be more accurate in this case to append to this, or is it that this functionality would have been expected from the documentation and was then broken in `1.2.0`.

megawac:
I'd like to see #3676 and #3732 merged before this release for future proofing `lodash v4`

paulfalgout:
> According to whom?
The changes for this update don't appear to be urgent enough for a release, except that they're ready to go and they're a blocker for the lodash v4 / underdash stuff mentioned here: https://github.com/jashkenas/backbone/issues/3720  So it seemed like that was the predominate reason for this release.

megawac:
Alright @paulfalgout could you just add something about #3676 and this looks :golf:

jdalton:
:shipit: !

paulfalgout:
Alright, latest draft is available @akre54 @jdalton @jridgewell @megawac @jashkenas   Tear it apart :crying_cat_face:  so we can :ship:

paulfalgout:
OK I also documented (to some extent) the previously undocumented event functionality.

jridgewell:
LGTM.

jashkenas:
Looks fine, but it doesn't really seem like enough meaty stuff to merit a whole **1.3** bump.

paulfalgout:
@jashkenas was worth a shot :-P
Was going off of:
```
if(meat > 0){
  version += 0.1;
}
```

paulfalgout:
:pineapple:

jridgewell:
@jashkenas or @braddunbar: mind pushing out a new version?

index.html
akre54: `click focus` may be confusing here. It seems like it is referring to DOM events rather than model events. Fix incoming.

jashkenas:
Done and done.

PR #3747, opened 2015-08-07
Use `some`, `includes`, `map` in tests instead of `any`, `include`, `pluck`.

jdalton:
Adjusts tests to follow up #3676.

### DIFF
--- a/test/collection.js
+++ b/test/collection.js
@@ -625,24 +625,26 @@
 
   test("Underscore methods", 19, function() {
     equal(col.map(function(model){ return model.get('label'); }).join(' '), 'a b c d');
-    equal(col.any(function(model){ return model.id === 100; }), false);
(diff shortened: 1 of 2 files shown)


megawac:
I'm not sure if we want to do some of this. I would be fine if we did `if _[function] <test>`

jdalton:
>  I would be fine if we did if `_[function] <test>`

I'm not sure what you mean.

megawac:
I mean instead of deleting the `includes`, `max`, `min` tests check if they exist on _ before executing the test

jdalton:
I would use `_.includes` but the version backbone tests doesn't have it. I can fix that though. The `_.max` min tests were adjusted to still test them just without the callback.

jdalton:
Ok updated.

PR #3752, opened 2015-08-10
Use _.result in more places.

jdalton:
Noticed a couple places `_.result` could be used and made that happen.


jridgewell:
I think the original code is much cleaner. We generally only use `_.result` when the `prop` can be a function or a normal value, not as a guard for nullish.

jdalton:
The plus here was handling the default value case.

PR #3753, opened 2015-08-10
Use more underscore methods and remove rogue unused param.

jdalton:
Use more underscore methods and remove rogue unused param.


jridgewell:
I'm ambivalent to the `#identity` -> `_.noop` change. But `#parse` and `#execute` were written with the unused params so that they'd be included in the annotated source.

Someone else want to weigh in?

akre54:
Agreed. Let's leave 'em be.

PR #3754, opened 2015-08-10
Remove _.matches proxy ref in docs

paulfalgout:
http://underscorejs.org/#matches doesn't work and is an alias (that may be deprecated)


jdalton:
It should actually be removed from the docs as it doesn't proxy to `matcher` or `matches`. However it does use `matches` internally.

paulfalgout:
Yep.  Didn't check if it was still in use, just noticed the broken link.  Removed.

PR #3758, opened 2015-08-13
v2: Collection#parse

jridgewell:

This is the revert of #3659 (which reverted #3644). So, really, it's just #3644 (which reverted #1407... reverts! :stuck_out_tongue:).

---

Third times the charm? :stuck_out_tongue_winking_eye:

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -757,6 +757,7 @@
     if (options.comparator !== void 0) this.comparator = options.comparator;
     this._reset();
     this.initialize.apply(this, arguments);
+    if (models && options.parse) models = this.parse(models, options);
     if (models) this.reset(models, _.extend({silent: true}, options));
   };
 
@@ -809,7 +810,6 @@
     // the core operation for updating the data contained by the collection.
     set: function(models, options) {
       options = _.defaults({}, options, setOptions);
-      if (options.parse && !this._isModel(models)) models = this.parse(models, options);
       var singular = !_.isArray(models);
       models = singular ? (models ? [models] : []) : models.slice();
       var id, model, attrs, existing, sort;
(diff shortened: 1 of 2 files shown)


d-reinhold:
[Line 837](https://github.com/jashkenas/backbone/pull/3758/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR837) in `Collection#set` is still checking for `options.parse` to determine if `Model#parse` should be called on the attrs, can that be removed as well? Because implementations of `Collection#parse` should be returning arrays of fully parsed models, right?

prantlf:
@d-reinhold, correction: Implementations of `Collection#parse` should be returning arrays of objects, which are expected to be consumed by the `Model#parse` method. If you "massage" the complete response in `Collection#parse`, what would you have the `Model#parse` for? How would you parse the response from the REST API returning a single object? Putting all functionality to `Collection#parse` would limit your framework to the REST API returning collections only..

prantlf:
@jridgewell, your suggestion does not cover the existing functionality.

New: `collection.set(collection.parse(models))`

As long as `collection.parse` does not return array of models, but an array of objects, which the models will be constructed with, there is no way to get `model.parse` called for every particular model. You lose the componentization - way to divide the scope of collection parsing (which is usually simple) and the scope of model parsing (which is usually complicated).

Existing: `collection.set(models, {parse: true})`

Unlike calling the `parse` method once from outside, this interface allows to propagate the parsing flag to model constructions. It allows componentization of the parsing code, even in non-homogeneous collections. Such separation encourages code sharing too; collection mixins, for example.

jgonggrijp:
I've seen some changing back and forth like this in Underscore as well. I'm not having it.

Issue #3545, opened 2015-03-23
Backbone Localstorage link broken in TODO app


Amit-Thawait:
The link referenced for `BACKBONE.LOCALSTORAGE.JS` in the Backbone TODO app is broken.

![backbone-todo](https://cloud.githubusercontent.com/assets/1220255/6777167/e22200ea-d16c-11e4-9be5-f6d188cd83ea.png)
 
![backbone-localstorage](https://cloud.githubusercontent.com/assets/1220255/6777200/43ac8466-d16d-11e4-9b0c-b62aa219c02f.png)

Amit-Thawait:
I think the correct link is http://backbonejs.org/docs/backbone-localstorage.html

If someone can confirm this then, I can update it.

akre54:
Ah. That is indeed weird. The difference between the two is `localStorage` vs `localstorage`. Good catch, let me look into it.

Issue #1720, opened 2012-10-10
Docs page links to 404 on to-do example


rizqidjamaluddin:
Affected page: http://backbonejs.org/docs/todos.html

The second link on the page returns a 404. Presumably the correct link reference should be to https://github.com/jeromegn/Backbone.localStorage.

PR #3759, opened 2015-08-14
Fixed a broken link.

ohbarye:
I found a broken link in [the document](http://backbonejs.org/docs/todos.html#section-1), so that fixed it. :)

### DIFF
--- a/docs/todos.html
+++ b/docs/todos.html
@@ -52,7 +52,7 @@
               </div>
               <p>An example Backbone application contributed by
 <a href="http://jgn.me/">Jérôme Gravel-Niquet</a>. This demo uses a simple
-<a href="backbone.localstorage.html">LocalStorage adapter</a>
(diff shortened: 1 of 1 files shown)


jridgewell:
Good eye!

akre54:
Previously: #3545 #1720

These docs are autogenerated by docco, so we shouldn't be editing them by hand.

I looked into this a while back when it last popped up and here's what I remember. The [canonical filename is `localStorage`](https://github.com/jashkenas/backbone/blob/1.2.1/examples/backbone.localStorage.js) (camelCased), but I think I recall at one point it might've been all lowercase (#2204 maybe?), which on case-insensitive filesystems (like OSX's default HFS+) reads in the file file, but spits out the wrong error. Weird. This definitely shouldn't still be happening.

ohbarye:
I'm sorry that I wasn’t aware that. I'll be careful.

akre54:
no harm no foul. thanks for catching it :)

Issue #3760, opened 2015-08-15
Collection#set doesn't intelligently merge unparsed models


d-reinhold:
`Collection#set` takes an option `parse` that, if set to true, will [call the collection's `parse` method on the input array](https://github.com/jashkenas/backbone/blob/master/backbone.js#L812), as well as the [model's `parse` method on individual objects in the input array] (https://github.com/jashkenas/backbone/blob/master/backbone.js#L837). This implies that it's ok for non-parsed data to be passed into `Collection#set`.

However, if non-parsed data is passed to `Collection#set`, [line 833](https://github.com/jashkenas/backbone/blob/master/backbone.js#L833) may fail to correctly merge the new data with existing data in the collection (in particular, if the non-parsed data is wrapped in a way that obscures the idAttribute).

Below is a failing test case that illustrates this issue:
```
  test("collection should merge in duplicate raw objects with {merge: true}", 1, function() {
    var Model = Backbone.Model.extend({
      parse: function(data) { return data.wrapper; }
    });

    var Col = Backbone.Collection.extend({model: Model});
    var col = new Col;

    col.set([{wrapper: {id: 1, name: 'Foo'}}], {parse: true, merge: true});
    var firstModel = col.first();

    col.set([{wrapper: {id: 1, name: 'Bar'}}], {parse: true, merge: true});
    var secondModel = col.first();

    strictEqual(firstModel, secondModel);
  });
```
 The `{merge: true}` option indicates that the updated model in the collection should be the same object reference, but instead it is a newly instantiated model. 

A potential fix might involve changing [line 833](https://github.com/jashkenas/backbone/blob/master/backbone.js#L833) from this:
```
        if (existing = this.get(attrs)) {
```

to this:
```
        var existing = this.get(attrs);
        if (options.parse) {
          var tempModel = new this.model(attrs, options);
          existing = existing || this.get(tempModel.attributes);
        }
        if (existing) {
```
although there may be a more elegant solution. If people agree that this is a bug, I can open a pull request.

jashkenas:
Perhaps. I'm not sure if this is something that is intended to be handled by `modelId` these days, or by `parse`? Anyone?

jridgewell:
Dupe of https://github.com/jashkenas/backbone/pull/3529. We closed that one, but I don't feel strongly one way or the other.

jashkenas:
Oh, right, so ids need to be parsed out at the collection level then...

d-reinhold:
[...]

Issue #2249, opened 2013-02-07
New model created unnecessarily before merging attrs into existing model


johtso:
It seems that when `.add()` is called with partial attributes corresponding to an existing model in a collection, even though the existing model just needs to be updated, [a new model is constructed](https://github.com/documentcloud/backbone/blob/master/backbone.js#L612) by `collection._prepareModel()`. (For example in the situation covered by this test: https://github.com/documentcloud/backbone/blob/master/test/collection.js#L921)

Am I right in thinking that `collection._prepareModel()` should only really be called if a new model is going to be added to the collection? It seems strange to have a new model initialized and have the cid counter incremented, when no new model is needed.

PR #3766, opened 2015-08-20
Models are now parsed before checking if they exist in a collection

d-reinhold:
See issue #3760

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -844,13 +844,17 @@
 
       // Turn bare objects into model references, and prevent invalid models
       // from being added.
-      var model;
+      var model, parsedModel;
       for (var i = 0; i < models.length; i++) {
         model = models[i];
 
         // If a duplicate is found, prevent it from being added and
         // optionally merge it into the existing model.
         var existing = this.get(model);
+        if (!existing && !this._isModel(attrs) && options.parse) {
+          parsedModel = new this.model(model, options);
+          existing = existing || this.get(parsedModel.attributes);
+        }
         if (existing) {
           if (merge && model !== existing) {
             var attrs = this._isModel(model) ? model.attributes : model;
@@ -866,7 +870,7 @@
 
         // If this is a new, valid model, push it to the `toAdd` list.
         } else if (add) {
-          model = models[i] = this._prepareModel(model, options);
+          model = models[i] = this._prepareModel(parsedModel || model, options);
           if (model) {
             toAdd.push(model);
             this._addReference(model, options);
@@ -1078,14 +1082,14 @@
 
     // Prepare a hash of attributes (or other model) to be added to this
     // collection.
-    _prepareModel: function(attrs, options) {
-      if (this._isModel(attrs)) {
-        if (!attrs.collection) attrs.collection = this;
-        return attrs;
+    _prepareModel: function(model, options) {
+      if (this._isModel(model)) {
+        model.collection = model.collection || this;
+      } else {
+        options = options ? _.clone(options) : {};
+        options.collection = this;
+        model = new this.model(model, options);
       }
-      options = options ? _.clone(options) : {};
-      options.collection = this;
-      var model = new this.model(attrs, options);
       if (!model.validationError) return model;
       this.trigger('invalid', this, model.validationError, options);
       return false;
--- a/test/collection.js
+++ b/test/collection.js
@@ -199,15 +199,36 @@
     equal(col.length, 1);
   });
 
-  test("merge in duplicate models with {merge: true}", 3, function() {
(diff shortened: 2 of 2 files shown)


jashkenas:
Instantiating models can also have side effects. I don't think we want to create extra ones just for the `parse`.

d-reinhold:
The only case where a model is instantiated early and subsequently only used for its parsed attributes is when the model is already in the collection, but the `idAttribute` is not in the root of the API response. Currently this case is not handled correctly in `master`: an extra model is instantiated, but instead of merging the new attributes onto the existing model, the existing model is removed from the collection, and the new one is pushed in. This PR fixes that problem, and in all other cases, the model is only instantiated/parsed once, just like before.

Certainly though I could be missing something; if you could provide a failing test case that illustrates your concerns, that'd be really helpful.

akre54:
This is how we used to do it, actually (#2249 and others). It's unexpected behavior, and a real pain to manage, when you start creating temporary models willy-nilly.

d-reinhold:
From my perspective, the unexpected behavior is that `Collection#fetch` doesn't work when your API wraps the `idAttribute` of your models. This PR fixes that. The code isn't pretty, but neither is `Collection#set` on `master` either, as @jashkenas [pointed out](https://github.com/jashkenas/backbone/issues/3760#issuecomment-132724960)

Sure, creating models 'willy-nilly' is bad. This creates a model in a very specific case that is currently unhandled. I guess I'd like to see a concrete example of how things could go wrong if this code was merged?

akre54:
Sure, and I hear ya. But there's a better way of fixing this than creating temporary models just to get their `idAttribute`. We can't (and likely shouldn't) enforce that creating a model is idempotent and side-effect-free, regardless of whether we think it's a good idea to put side effects in your `initialize` or not. It's just too surprising.

Thanks for your feedback and help on this so far. Let's try and get #3758 to a good state that fixes your use case.

d-reinhold:
@akre54 The `parsedModel` that is instantiated in this PR is also being instantiated when this code path executes in `master`. It's not a temporary model that is only used to get the `idAttribute`. All I'm doing differently is instantiating that model earlier (and explicitly), and using its attributes to update the existing model in the collection (if one exists), or adding the parsed model to the collection (if it wasn't there already). When the preconditions that trigger `parsedModel` to be instantiated occur in the `master` branch, the same model is be instantiated [here](https://github.com/jashkenas/backbone/blob/master/backbone.js#L869). I guess I did a poor job of explaining that, but this PR is really not doing any extra parsing or instantiating.

I'm all for a clean fix for this in v2; I already left feedback on #3758 last week. But V2 still feels pretty far out. I think this can be fixed in a backwards compatible way, do you disagree?

akre54:
[Line 856](https://github.com/jashkenas/backbone/pull/3766/files#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR856) says differently. If `{wrapper: {id: 123}}` fails the initial `get` line, it will create a temporary model, then use the temporary model's attributes to find it in the collection.

In your test case, you could use `modelId` to get the right model from the collection with zero changes to Backbone.

d-reinhold:
@akre54 Not quite. If `{wrapper: {id: 123}}` fails in the initial `get`, a new object `parsedModel` is instantiated early, correct. If a model with `id = 123` is not in the collection, `parsedModel` is added (and therefore is not being instantiated unnecessarily, it has to be instantiated at some point). If the model with `id = 123` *is* in the collection, we are able to find it, and so we call `existing.parse(attrs)` to update the existing model, and `parsedModel` is garbage collected. Seems like a waste.

However, the way `Collection#set` is implemented in `master`, the `get` for `{wrapper: {id: 123}}` will *always fail*, even if a model with `id = 123` is in the collection. So Backbone will proceed to the `add` step, and a duplicate model will be instantiated at that point! The existing model with `id=123` will get thrown out of the collection, and the duplicate is added instead. What should happen in this situation is the existing model gets updated with the new attributes, if there are any. That's what this PR does.

Do you see why, in both `master` and in this PR, the same number of models get instantiated in both cases? Obviously we'd like to have no duplicate models created, but this PR doesn't make the situation worse, and it makes the behavior consistent with when the passed-in models are not wrapped.

d-reinhold:
Regarding `Collection#modelId`, I guess it could be overridden to return `attrs.id || attrs.wrapper.id`, but then the collection would have to know about how the model is parsed. I guess that wouldn't be a terrible solution to this problem though.

akre54:
Yeah it's not the cleanest or clearest solution to have to duplicate logic between `parse` and `modelId` (we're thinking of better solutions, would love if you've got one), but I'd argue it's a lot better than creating a model unnecessarily. `modelId` is the way to go here.

Issue #3622, opened 2015-05-21
Documentation: #delegate and #undelegate


jridgewell:
Can't get to it right now, just needed something so we remember to document them later.

guahanweb:
Any particular doc generator you're using, or is this a hand-written documentation guide? Looks generated, but I don't see where the build script is in the fork.

akre54:
It's hand-written :)

Addressed in #3771

PR #3771, opened 2015-08-24
Document delegate/undelegate

akre54:
Addresses #3622.

### DIFF
--- a/index.html
+++ b/index.html
@@ -3001,6 +3001,12 @@ var InputView = Backbone.View.extend({
     </p>
 
     <p>
+      A single-event version of <b>delegateEvents</p> is available as <tt>delegate</tt>.
+      In fact, <b>delegateEvents</b> is simply a multi-event wrapper around <tt>delegate</tt>.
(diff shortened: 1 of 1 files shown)


philfreo:
Should `delegate` and `undelegate` appear in the documentation's left-nav as separate entries? I read in the changelog these methods were added but it was strange to not see them in the left nav under View.

akre54:
They're not really meant to get called by your average user, they're mostly just there to give plugin authors a consistent way of handling event binding. A one-paragraph note seems like it's consistent with the rest of the docs. Any reason we'd need to call them out specifically?

philfreo:
I just wasn't aware of any other methods that weren't "private" but that weren't documented, so it just seemed inconsistent

akre54:
Routing and history are the big ones, but there's also the bits about `render` and `template` and the events hash in `delegateEvents`. `delegate` and `undelegate` are kind of unique in Backbone: most "private" methods are prefixed with an underscore to show they're unstable implementation details. These are stable -- they're not going to change anytime soon.

I don't feel strongly though -- it just feels like added clutter.

PR #3779, opened 2015-08-27
model.fetch.then

godlark:
Recently I've made code similiar to:

book.fetch().then(function(resp) {
});

And I've expected to get a parsed response with my 'Book.parse' function but I've got a raw response.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -595,7 +595,9 @@
         model.trigger('sync', model, resp, options);
       };
       wrapError(this, options);
-      return this.sync('read', this, options);
+      return this.sync('read', this, options).then(function(response) {
+        return model.parse(response);
+      });
     },
 
     // Set a hash of model attributes, and sync the model to the server.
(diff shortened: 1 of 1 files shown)


akre54:
Closing in favor of #3729 and #3651. Please do a search before opening issues.

godlark:
Have you read the issue before closing? My pull request complain about how fetch function works. I DO NOT complain about using jquery.Deffered etc.

akre54:
`Backbone.sync` [proxies to `Backbone.ajax`](https://github.com/jashkenas/backbone/blob/1.2.2/backbone.js#L1410) behind the scenes. The `then` function you're getting from the return value of `fetch` is simply the return value from jQuery.ajax, a `jqxhr` (xhr wrapped in jQuery.Deferred.)

Backbone has no use for Promises at the moment. We don't even guarantee that the fetch function will return a `thenable` (i.e. a function with a `then` method) as your change requires, therefore we can't add add this change until we make some major plumbing decisions like the ones in #3729 and #3651. If you want this change, stop by those issues and register your input, they might have already added it.

godlark:
Ok, thanks for the reply :)

PR #3770, opened 2015-08-21
fix error when adding model to collection at an out of bounds index

ahumphreys87:
@jridgewell 

Fixes an issue noticed when running Marionette tests against 1.2.2. Adding a model at index 5 to a collection with a length of 3 throws an error. Previously this would add to the end of the collection.

PR #3780, opened 2015-08-28
Add out-of-bounds default to Collection#at

tcandens:
With this methods recent change of accepting negative arguments, it bothered me that undefined was being returned half the time. Although this might not be ideal, I feel like it add some consistency.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -972,6 +972,7 @@
 
     // Get the model at the given index.
     at: function(index) {
+      if (index > this.length) index = this.length - 1;
       if (index < 0) index += this.length;
       return this.models[index];
     },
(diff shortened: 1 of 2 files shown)


Florian-R:
Correct me if i'm wrong, but i think it has already been fixed in gh-3770

akre54:
Yup. Thanks @Florian-R

tcandens:
Looks like gh-3770 was a fix for *adding* models past the end of a collection. This would be simply for accessing collection models with collection.at( index ). This will return the last model when index is beyond length, keeping it consistent with the adding models behavior.

akre54:
My mistake (got thrown off by the "add" in the title...).

I still don't think we should change this. Collections are analogous to Arrays in javascript, and indexing beyond the bounds of an Array returns `undefined`.

It would be overly surprising behavior if we automatically clamped lookups for you (additions are fine, since you're likely just trying to add it to the end of the array already. If you want to get the last value, just use `col.last()`.)

PR #3784, opened 2015-09-03
Submitting a typo change in listenTo under Events.

tadwhitaker:
Love Backbone and just trying to tighten up this one word!


jashkenas:
Close, but the bold is actually referring to the name of the documented argument. It's just a little code wordplay.

Issue #3715, opened 2015-07-15
Documentation of options


Industrial:
Hi.

http://backbonejs.org/#Model-set

This (and many other methods of backbone classes) doesn't include documentation of all the arguments. I can not seem to find these options.

jridgewell:
Would you like to work on a PR?

Industrial:
Sure! :+1:

macgyver:
:+1: I had to look at the source code to realize that `.set` has a multi-variate signature

paulfalgout:
@akre54 is there work to be done here?  Looks like the PR for this issue was rejected: https://github.com/jashkenas/backbone/pull/3788/files  Not sure what else this issue would be useful for.

PR #3788, opened 2015-09-06
Updated model.set documentation

Niades:
Tried to make the model.set documentation more detailed, including the description of the `options` parameter, as requested in #3715.


akre54:
I don't know. I think the more common use case is to pass back a set of attributes from a server.

jashkenas:
Yep, this is expanding on two "features" — `silent` and `unset:true` — neither of which you actually want to use. That's why they're not highlighted.

PR #3803, opened 2015-09-23
Fix _removeModels regression

### LINKED ISSUE #3693: _removeModels regression bug
It looks like in 1.2.1, _removeModels was simplified, but introduced a regression bug.

```
    _removeModels: function(models, options) {
      var removed = [];
      for (var i = 0; i < models.length; i++) {
        var model = this.get(models[i]);
        if (!model) continue;
        // CODE REMOVED HERE

        var index = this.indexOf(model);
        this.models.splice(index, 1);
        this.length--;

        if (!options.silent) {
          options.index = index;
        }

        removed.push(model);
      }
      return removed.length ? removed : false;
    },
```

Removed the following lines after `if (!model) continue;`

```
        var id = this.modelId(model.attributes);
        if (id != null) delete this._byId[id];
        delete this._byId[model.cid];
```

What this means is that if you are listening to a remove event and you check for removal, you will get into an infinite loop because the event is now called before the id is removed from `this._byId`:

```
    if (current_related_model = collection.get(id)) collection.remove(current_related_model)
```

I think the _byId code should be put back so when events are fired, they represent the most up-to-date state.

jridgewell:
Fixes #3693.

This leaves open the question of whether events triggered on the model
during a ‘remove’ listener should also trigger on the collection. Just
something to revisit for V2.

```js
col.on('other', (model) => {
  // Should this be triggered?
});

col.on('remove', (model) => {
  // If the model is really "removed" (we can't `#get` it anymore)
  // by the time this listener is called, then I'd argue that this
  // shouldn't trigger the 'other' event on the collection...
  model.trigger('other');
});
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1088,6 +1088,12 @@
         this.models.splice(index, 1);
         this.length--;
 
+        // Remove references before triggering 'remove' event to prevent an
+        // infinite loop. #3693
+        delete this._byId[model.cid];
+        var id = this.modelId(model.attributes);
+        if (id != null) delete this._byId[id];
+
         if (!options.silent) {
           options.index = index;
           model.trigger('remove', model, this, options);
--- a/test/collection.js
+++ b/test/collection.js
@@ -298,12 +298,13 @@
     deepEqual(col.pluck('id'), [1, 2, 3]);
   });
 
-  test("remove", 10, function() {
+  test("remove", 11, function() {
     var removed = null;
     var result = null;
     col.on('remove', function(model, col, options) {
       removed = model.get('label');
       equal(options.index, 3);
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -1088,6 +1088,12 @@
         this.models.splice(index, 1);
         this.length--;
 
+        // Remove references before triggering 'remove' event to prevent an
```
dmitrigrabov: Does this work?

akre54:
All events maybe not, but a `remove` definitely should.

jridgewell:
I think this warrants a `v1.2.4`?

akre54:
Want to add a test that `col.on('other', ...)` doesn't trigger?

jridgewell:
It's in #3806, but I'm waiting till v2.

PR #3804, opened 2015-09-23
Set up subclassing using Object.create

kuychaco:
This change eliminates having to define and invoke a Surrogate function, makes the source code more readable, and reduces the size of the backbone.js framework. 

Object.create sets up the prototype delegation relationship without calling the "parent" constructor function. This is more straight forward than using a Surrogate constructor function just for the sake of setting up subclassing.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1857,9 +1857,8 @@
 
     // Set the prototype chain to inherit from `parent`, without calling
     // `parent` constructor function.
-    var Surrogate = function(){ this.constructor = child; };
-    Surrogate.prototype = parent.prototype;
-    child.prototype = new Surrogate;
+    child.prototype = Object.create(parent.prototype);
+    child.prototype.constructor = child;
 
     // Add prototype properties (instance properties) to the subclass,
     // if supplied.


jridgewell:
https://github.com/jashkenas/backbone/pull/3553 :wink:. Thanks, though!

kuychaco:
@jridgewell, some thoughts - while similar to [#3553](https://github.com/jashkenas/backbone/pull/3553), this isn't so much a "duplicate" as it is an alternative approach. With #3553 there were concerns with breaking backwards compatibility. Object.create is vanilla JS (part of the standard JS library since JavaScript 1.8.5), making backwards compatibility a non-issue. Thoughts?

jridgewell:
Using native `Object.create` won't work for IE8 or below, which is even more of a breaking change. `_.create` will fall back to the `Surrogate` method when needed.

akre54:
@kuychaco the "backwards compatibility" concerns from #3553 are about people using older versions of Underscore (which didn't have `_.create`), it's a separate concern from browser support. For what it's worth, Underscore [falls back to `Object.create`](https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L118) if it exists.

kuychaco:
Gotcha, thanks for clarifying!

PR #3806, opened 2015-09-25
Model events during "remove" do not trigger on Col

jridgewell:

On the heels of #3803, this prevents events triggered on the model during the "remove" event listener from bubbling up to the collection.

Why? Because the model is "removed" from the collection. You can't `#get` it any more, there're no more references. The collection shouldn't know anything about the model once that event is fired.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1102,20 +1102,14 @@
         var index = this.indexOf(model);
         this.models.splice(index, 1);
         this.length--;
-
-        // Remove references before triggering 'remove' event to prevent an
-        // infinite loop. #3693
-        delete this._byId[model.cid];
-        var id = this.modelId(model.attributes);
-        if (id != null) delete this._byId[id];
+        this._removeReference(model, options);
+        removed.push(model);
 
         if (!options.silent) {
           options.index = index;
           model.trigger('remove', model, this, options);
+          this._onModelEvent('remove', model, this, options);
         }
-
-        removed.push(model);
-        this._removeReference(model, options);
       }
       return removed.length ? removed : false;
     },
--- a/test/collection.js
+++ b/test/collection.js
@@ -310,8 +310,12 @@
     var result = null;
     col.on('remove', function(model, col, options) {
       removed = model.get('label');
+      model.trigger('other');
       equal(options.index, 3);
       equal(col.get(model), undefined, '#3693: model cannot be fetched from collection');
+    });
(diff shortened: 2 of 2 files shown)


jashkenas:
That makes perfect sense. You really don't think this should be merged?

jridgewell:
Not now, it's a breaking change. I figure if [this many](https://github.com/jashkenas/backbone/issues/3693) people are causing an infinite loop in the "remove" event, they're also relying on event bubbling.

platinumazure:
The `remove` event itself is still fired on the collection, though, right?

jridgewell:
Yes.

backbone.js
```diff
@@ -1102,20 +1102,14 @@
...
-        delete this._byId[model.cid];
-        var id = this.modelId(model.attributes);
-        if (id != null) delete this._byId[id];
+        this._removeReference(model, options);
+        removed.push(model);
 
         if (!options.silent) {
           options.index = index;
           model.trigger('remove', model, this, options);
+          this._onModelEvent('remove', model, this, options);
```
akre54: Why call `_onModelEvent` here? We know it will bypass all of the conditionals because we don't specifically handle `remove`.

jridgewell: Since it could be overridden. We're not exactly clear on what is overridable and what's not. #3703

akre54: :eyeroll:

The ones that are meant to be overridden are usually obvious, or their
utility becomes clear when you need them. They don't get in your way if you
don't need them.

I can't think of a reason why you'd override _onModelEvent. It seems to be
there mostly to make it easier to unbind later.

Theres no issue using _onModelEvent but I like the symmetry of the two
triggers in a row.

On Mon, Sep 28, 2015, 10:34 Justin Ridgewell <notifications@github.com>
wrote:

> In backbone.js
> <https://github.com/jashkenas/backbone/pull/3806#discussion_r40558531>:
>
> >
> >          if (!options.silent) {
> >            options.index = index;
> >            model.trigger('remove', model, this, options);
> > +          this._onModelEvent('remove', model, this, options);
>
> Since it could be overridden. We're not exactly clear on what is
> overridable and what's not. #3703
> <https://github.com/jashkenas/backbone/pull/3703>
>
> —
> <https://github.com/jashkenas/backbone/pull/3806/files#r40558531>.
>
-- 

Adam K (mobile)

PR #3810, opened 2015-09-28
Reduce without initial value doesn't work

just-boris:
Underscore's reduce as well as native Array.prototype.reduce can work without initial value.

See two similar calls:
```js
equal(_.reduce(col.models, function(a, b) {return a.id > b.id ? a : b}).id, 3); // works fine
equal(col.reduce(function(a, b) {return a.id > b.id ? a : b}).id, 3); // don't work
```

This behaviour seems to be broken since https://github.com/jashkenas/backbone/commit/7d646745618dc86ff74230a2063083963b7c343a, because this commit forces to call underscore with fixed number of arguments, but underscore detects this count and cannot switch to mode without initial value and starts with undefined.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -87,6 +87,9 @@
         return _[method](this[attribute], cb(iteratee, this), context);
       };
       case 4: return function(iteratee, defaultVal, context) {
+        if(arguments.length < 2) {
+          return _[method](this[attribute], cb(iteratee, this));
+        }
         return _[method](this[attribute], cb(iteratee, this), defaultVal, context);
       };
       default: return function() {
--- a/test/collection.js
+++ b/test/collection.js
@@ -631,10 +631,11 @@
     equal(coll.findWhere({a: 4}), void 0);
   });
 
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -87,6 +87,9 @@
         return _[method](this[attribute], cb(iteratee, this), context);
       };
       case 4: return function(iteratee, defaultVal, context) {
+        if(arguments.length < 2) {
+          return _[method](this[attribute], cb(iteratee, this));
+        }
```
jridgewell: Rather than add another condition here, I'd rather deopt `reduce` and friends into the `default` case below. Mind updating the [`collectionMethods`](https://github.com/jashkenas/backbone/blob/master/backbone.js#L1169-L1176) object, giving `reduce`, `inject`, `foldl`, `reduceRight` and `foldr` a value of `0`?

just-boris:
@jridgewell close this, I am going to continue discussion in #3820

Issue #3809, opened 2015-09-28
Add underscore methods `_.findIndex, `_.findLastIndex` and  `_.sortedIndex`


jacobbuck:
Finding myself writing:

```js
var index = collection.indexOf(collection.findWhere({foo: 'bar'}));
```

when this is so much better:

```js
var index = collection.findIndex({foo: 'bar'});
```

jridgewell:
Want to work on a PR?

jacobbuck:
Sure, just submitted one now. Though I'm not sure how to approach adding `_.sortedIndex` as it's iteratee and value arguments are in reverse order to the other underscore methods.

jridgewell:
Closed by #3811.

PR #3811, opened 2015-09-28
Add `findIndex` and `findLastIndex` underscore methods to Collection.

jacobbuck:
Related to #3809


megawac:
Could you not regen those files -- just modify backbone.js and the tests?

jacobbuck:
Sure thing, updated now.

Issue #3799, opened 2015-09-18
Router - passing an object with route params into handler


ntgn81:
Currently, matched parameters are passed into the action as an `array`

```javascript
routes: {
  "folder/:name-:mode": "openFolder"
},

openFolder: function(name, mode, queryString) {
  // do stuff
}
```

I would like the actions to be given an object with `name`, `params`, `queryString` instead, resembling express routing solution.

```javascript
routes: {
  "folder/:name-:mode": "openFolder"
},

openFolder: function(routeData) {
  /* routeData is now
  {
    name: "openFolder",
    params: {
      name: "valueOfName",
      mode: "valueOfMode"
    },
    queryString: "query=string&a=b"
  }
  */
}
```

It could be an option that can be enabled on BB.Router.
Is it a good idea to have this functionality in base BB.Router? Or it can be an external library?

I have tried it out, and it seems to be working well, with a new set of tests cloned from current router test to make sure nothing breaks.

https://github.com/ntgn81/backbone/commit/7b2c8b3dce5c19f5dc36ddd14c4f70d22108db45

Thanks!

akre54:
We could do that. Might be a nice addition for V2. Your implementation is a little heavyweight though. Any way we could slim it down?

ntgn81:
Thanks, @akre54! 

Can you please give me some pointers as to where the heavy code might be?

I think it could be cleaned up a bit if it was to replace the current logic. But this was done with backward compatibility in mind, so I just added the stuff on top of the existing logic.

akre54:
I would guess (without looking too much into it) that it should be possible to combine some of the logic with what we already have. I'll respond on the diff.

megawac:
Related v2 router changes https://github.com/jashkenas/backbone/pull/3660. /cc @jmeas

jmeas:
This is a great idea, I think!

For folks looking for this feature now, [BaseRouter](https://github.com/jmeas/backbone.base-router) returns a hash for named params and query params, though I can't recall if it works with splats.

As is evident by the commit referenced in the OP, the DSL implementation in Backbone is the biggest hindrance to this feature. I wasn't planning on making any changes to the DSL implementation in #3660 to keep concerns separate, but I'm hugely in favor of updating it. I had a vague plan to do a subsequent PR focused solely on the DSL, but I'm not sure if/when I'll get a chance to do that anytime soon.

ntgn81:
[...]

PR #3815, opened 2015-10-02
Router add use object params option

ntgn81:
Adding optional parameter to Router to make route callback handlers called with route params being an object instead of array.

https://github.com/jashkenas/backbone/issues/3799

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1455,6 +1455,16 @@
 
   // Set up all inheritable **Backbone.Router** properties and methods.
   _.extend(Router.prototype, Events, {
+    // If set to true, route handlers will be called with routeData object
+    //
+    //     this.route('search/:query/p:num', 'search', function(routeData) {
+    //       // for route 'search/backbone/p5?a=b'
+    //       // routeData.params.query === 'backbone'
+    //       // routeData.params.num === '5'
+    //       // routeData.queryString === 'a=b'
+    //     });
+    //
+    useParamsObject: false,
 
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
@@ -1467,12 +1477,26 @@
     //     });
     //
     route: function(route, name, callback) {
-      if (!_.isRegExp(route)) route = this._routeToRegExp(route);
+      // Pass paramNames into _routeToRegExp and let that function populate it
+      var paramNames = [];
+      // Need to know this so we don't try to create params object when the route is matched
+      var isRegExpRoute = _.isRegExp(route);
+
+      if (!isRegExpRoute) route = this._routeToRegExp(route, paramNames);
+
       if (_.isFunction(name)) {
         callback = name;
         name = '';
       }
       if (!callback) callback = this[name];
+
+      this._routeInfos = this._routeInfos || {};
+      this._routeInfos[route.toString()] = {
+        name: name,
+        isRegExpRoute: isRegExpRoute,
+        paramNames: paramNames
+      };
+
       var router = this;
       Backbone.history.route(route, function(fragment) {
         var args = router._extractParameters(route, fragment);
@@ -1511,13 +1535,18 @@
 
     // Convert a route string into a regular expression, suitable for matching
     // against the current location hash.
-    _routeToRegExp: function(route) {
+    _routeToRegExp: function(route, paramNames) {
       route = route.replace(escapeRegExp, '\\$&')
                    .replace(optionalParam, '(?:$1)?')
                    .replace(namedParam, function(match, optional) {
-                     return optional ? match : '([^/?]+)';
+                     if (optional) return match;
+                     paramNames.push(match.slice(1));
+                     return '([^/?]+)';
                    })
-                   .replace(splatParam, '([^?]*?)');
+                   .replace(splatParam, function(match) {
+                     paramNames.push(match.slice(1));
+                     return '([^?]*?)';
+                   });
       return new RegExp('^' + route + '(?:\\?([\\s\\S]*))?$');
     },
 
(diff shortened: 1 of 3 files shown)


backbone.js
```diff
@@ -1511,13 +1535,18 @@
 
     // Convert a route string into a regular expression, suitable for matching
     // against the current location hash.
-    _routeToRegExp: function(route) {
+    _routeToRegExp: function(route, paramNames) {
```
ntgn81: @akre54 
> I'm not so keen on modifying passed arguments. Let's bikeshed this decision a bit more (you may have been right in the first place).

I don't like this either, but this feels the least intrusive/most optimal since we are already RegExp parsing the input route string. And we can reuse `namedParam` and `splatParam` right where they are used.

Since `_routeToRegExp()` is only used from `route()`, how about moving it's body inside `route()` itself? Might be breaking change if someone is ovewritting `_routeToRegExp()` though.

backbone.js
```diff
@@ -1455,6 +1455,16 @@
 
   // Set up all inheritable **Backbone.Router** properties and methods.
   _.extend(Router.prototype, Events, {
+    // If set to true, route handlers will be called with routeData object
+    //
+    //     this.route('search/:query/p:num', 'search', function(routeData) {
```
ntgn81: @akre54 
> I don't know about the routeData object though. How about (params, queryString, name) as the arguments to the handler? Or (params, routeData) where routeData is queryString and name.

I think like `(params, routeData)` more. Having `routeData` as object gives option to expand the object with more stuff to help the route handler if needed, few things I could think of:
- `query` - object representation of queryString: 'x=true&y=5&z=abc' => `{x: "true", y: "5", z: "abc"}`.
  - Another option to enable queryString parsing maybe?
  - @jmeas is already doing this in [Backbone Router](https://github.com/jmeas/backbone.base-router#query) 
- `fragment`

backbone.js
akre54: > how about moving it's body inside route() itself?

I'm concerned about taking a nice small chunk of reusable, functionally pure code and pulling it out of a method and into the body of an already big method. 

I don't have a ton of time to work through this at the moment, but I should be freer next week. Try opening a pull and solicit feedback from other Backbone contributors. Looking forward to seeing this merged.

backbone.js
jmeas: :+1: to `routeData` over `1...n` args. I'd like to see Backbone's router give even more info in route callbacks, and a `1...n` arg API would get a little unwieldy with that many args.

ntgn81:
Any backbone guru have time to take a look at this and give me guidance on how to make this better? 

Thanks!

backbone.js
akre54: [This is what express uses](https://github.com/pillarjs/path-to-regexp), so passing the array is probably fine.

ntgn81: I was thinking about this too. What if this method, instead of returning RegExp, returns an object `{routeRegExp, paramNames}`. It would keep it functionally pure.

backbone.js
```diff
@@ -1467,12 +1477,26 @@
...
+      if (!isRegExpRoute) route = this._routeToRegExp(route, paramNames);
+
       if (_.isFunction(name)) {
         callback = name;
         name = '';
       }
       if (!callback) callback = this[name];
+
+      this._routeInfos = this._routeInfos || {};
+      this._routeInfos[route.toString()] = {
```
jridgewell: `#toString` shouldn't be necessary, it'll be coerced by the object.

backbone.js
```diff
@@ -1511,13 +1535,18 @@
 
     // Convert a route string into a regular expression, suitable for matching
     // against the current location hash.
-    _routeToRegExp: function(route) {
+    _routeToRegExp: function(route, paramNames) {
       route = route.replace(escapeRegExp, '\\$&')
                    .replace(optionalParam, '(?:$1)?')
                    .replace(namedParam, function(match, optional) {
-                     return optional ? match : '([^/?]+)';
+                     if (optional) return match;
```
jridgewell: I think we should be recording optional params as well.

backbone.js
```diff
@@ -1526,11 +1555,36 @@
...
         if (i === params.length - 1) return param || null;
         return param ? decodeURIComponent(param) : null;
       });
+
+      if (this.useParamsObject) {
+        // get routeInfo, which has array of param names
+        var routeInfo = this._routeInfos[route.toString()];
+
+        var routeData = {
+          name: routeInfo.name
```
jridgewell: I'm not sure the route's name is useful. Is there a usecase?

backbone.js
```diff
@@ -1526,11 +1555,36 @@
...
+        }
+        else {
+          routeData.params = _.object(routeInfo.paramNames, decodedParams);
+          var queryString = decodedParams[decodedParams.length - 1];
+          if (queryString) {
+            routeData.queryString = queryString;
+          }
+        }
+
+        return [routeData];
```
jridgewell: I'd much rather have these returned as separate parameters: `[paramsObject, queryString]`, instead of a nested object.

backbone.js
ntgn81: Yep, they are being recorded. At this point, this replace handles all named parameters, including optional ones, as previous replace already appends : to the name:

'named/optional/(y:z)'
  .replace(escapeRegExp, '\\$&')                    // 'named/optional/(y:z)''
  .replace(optionalParam, '(?:$1)?')                // 'named/optional/(?:y:z)?'
  .replace(namedParam, function(match, optional) {
    // this function gets called twice
    // 1st
    //   match: '(?:y'
    //   optional: '(?'
    // 2nd
    //   match: ':z'
    //   optional: undefined
  });

backbone.js
ntgn81: Good point. I will update. Thanks.

backbone.js
jridgewell: :+1:

backbone.js
ntgn81: I am currently using that right now in my app, where I hold on to current route data. And I use route.name to know what route it is.

Something along the line of `router.getCurrentRoute`, since it's not part of the route handler, `name` is what tells me what current route is.

backbone.js
ntgn81: So the handler would receive `(paramsObject, queryString)` as parameters, right?

We also discussed about this up here:
 https://github.com/jashkenas/backbone/pull/3815#discussion_r41050447

jridgewell: I need to read better, will continue discussion there.

backbone.js
jridgewell: I'd be ok with `(params, routeData)`. The most important piece of information here is the `params` object, hiding it inside `routeData` just adds friction.

ntgn81: :+1:

ntgn81:
Updated routed handler to receive `(params, routeData)` with tests.

Removed the route.toString in hash lookup too.

@jridgewell @akre54

wlepinski:
Is this PR planned to land on master?

PR #3816, opened 2015-10-03
Migrated tests to QUnit 2.0 syntax.

### LINKED ISSUE #3813: Unit tests are using QUnit 1.x syntax
The unit tests are using QUnit 1.x syntax (i.e., `module` global instead of `QUnit.module`, not using the `assert` object, etc.). And why should they do otherwise when QUnit 2.0 isn't out yet? However, as you probably know, most or all of the QUnit 2.0 syntax is in fact available now (as of QUnit 1.16, and we're consuming 1.18+). So it's possible to migrate everything over now, in order to make an eventual consumption of QUnit 2.0 easier.

Unless strong objections are raised in this discussion, I am happy to migrate the tests. I have a tool [platinumazure/qunit2-migrator][1] which makes the job a lot easier. I can do so all at once or in batches-- it doesn't matter.

  [1]: https://github.com/platinumazure/qunit2-migrator

platinumazure:
Fixes #3813.

I had to manually migrate `test/router.js` because my migrator apparently can't handle non-ASCII characters. Please do pay special attention to that one.

I don't know CoffeeScript that well, but it looks like the assertions in that library are simply falling back to the Node environment (and therefore its assert module). Thus I assume no modification needs to happen to that file (due to it not actually being QUnit tests). Please let me know if this is incorrect.

### DIFF
--- a/test/events.js
+++ b/test/events.js
@@ -1,41 +1,43 @@
 (function() {
 
-  module("Backbone.Events");
+  QUnit.module("Backbone.Events");
 
-  test("on and trigger", 2, function() {
+  QUnit.test("on and trigger", function(assert) {
+    assert.expect(2);
     var obj = { counter: 0 };
     _.extend(obj,Backbone.Events);
     obj.on('event', function() { obj.counter += 1; });
     obj.trigger('event');
-    equal(obj.counter,1,'counter should be incremented.');
+    assert.equal(obj.counter, 1, 'counter should be incremented.');
     obj.trigger('event');
     obj.trigger('event');
     obj.trigger('event');
     obj.trigger('event');
-    equal(obj.counter, 5, 'counter should be incremented five times.');
+    assert.equal(obj.counter, 5, 'counter should be incremented five times.');
   });
 
-  test("binding and triggering multiple events", 4, function() {
+  QUnit.test("binding and triggering multiple events", function(assert) {
+    assert.expect(4);
     var obj = { counter: 0 };
     _.extend(obj, Backbone.Events);
 
     obj.on('a b c', function() { obj.counter += 1; });
 
     obj.trigger('a');
-    equal(obj.counter, 1);
+    assert.equal(obj.counter, 1);
 
     obj.trigger('a b');
-    equal(obj.counter, 3);
+    assert.equal(obj.counter, 3);
 
     obj.trigger('c');
-    equal(obj.counter, 4);
+    assert.equal(obj.counter, 4);
 
     obj.off('a c');
     obj.trigger('a b c');
-    equal(obj.counter, 5);
+    assert.equal(obj.counter, 5);
   });
 
-  test("binding and triggering with event maps", function() {
+  QUnit.test("binding and triggering with event maps", function(assert) {
     var obj = { counter: 0 };
     _.extend(obj, Backbone.Events);
 
@@ -50,23 +52,23 @@
     }, obj);
 
     obj.trigger('a');
-    equal(obj.counter, 1);
+    assert.equal(obj.counter, 1);
 
     obj.trigger('a b');
-    equal(obj.counter, 3);
+    assert.equal(obj.counter, 3);
 
     obj.trigger('c');
-    equal(obj.counter, 4);
+    assert.equal(obj.counter, 4);
 
     obj.off({
       a: increment,
       c: increment
     }, obj);
     obj.trigger('a b c');
-    equal(obj.counter, 5);
+    assert.equal(obj.counter, 5);
   });
 
-  test("binding and triggering multiple event names with event maps", function() {
+  QUnit.test("binding and triggering multiple event names with event maps", function(assert) {
     var obj = { counter: 0 };
     _.extend(obj, Backbone.Events);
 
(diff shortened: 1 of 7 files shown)


platinumazure:
Also, as a side note, there isn't anything in `devDependencies` that would allow me to run `npm install` and `npm run lint`. It assumes a global `jsl` as a result, which I don't have.

akre54:
Might it be easier just to "import" / destructure the functions up top? That is..

```js
var module = Qunit.module,
    test = Qunit.test,
    expect = assert.expect,
    // ... etc
```

platinumazure:
And enable people to keep writing syntax the QUnit team doesn't support?

What exactly is wrong with how I've done it?

Also, if you look carefully, you'll see that the assert argument is
injected into each test, so there IS no way to get a module-scope version
of any functions belonging to that object.

Please read the [QUnit 2 upgrade guide][1] for more information.

  [1]: http://qunitjs.com/upgrade-guide-2.x

test/events.js
```diff
@@ -161,186 +168,202 @@
...
-    })
-    .trigger('a b');
-    ok(a);
-    ok(b);
-    equal(obj.counter, 2);
-  });
-
-  test("on, then unbind all functions", 1, function() {
+    obj
+      .on('all', function(event) {
```
captbaritone: Why the line break?

platinumazure: Personal style I guess? (I looked *everywhere* for a style guide, by the way- couldn't find one.) I couldn't stand trying to read/format code where chained method calls were indented to the same level.

I think I chose to line break to avoid having to indent the event handler function twice relative to the first line, so that the next method in the chain would be out one level relative to the function, but in one level relative to the main object.

That is, I was avoiding this pattern:

    obj.on('all', function (event) {
            // Needs two indents!?
        })
        .trigger('a b');

As I mentioned in the PR comments, there is no easy way to run the linter, nor is it included in `npm test`, nor does it seem to be run in Travis. If y'all care about code style in this repository, you have a funny way of showing it :-(

captbaritone: I don't think I've seen that style of chained methods before. It might be worth trying to keep the pull request scoped to just the QUnit changes and leave the code formatting for another PR.

That being said, take my advice with a grain of salt, I'm not a real contributor, just and interested user. I opened the tests for the first time today, and I could not agree with you more. The code "style" looks like it could use some love. Usually I'm not one to get too opinionated about code style, but I found the inconsistencies made it very hard to read.

platinumazure: Well, scoping the PR to relevant changes only is fair feedback. That said, I might wait until more folks ask for the same before making the change- not because you are "not a real contributor" but because amending and force-pushing a bunch of manual style changes is a pain in the rear end. Does that seem fair? Thanks again for the feedback!

akre54: Yeah lets switch this back. We can do style changes in a separate PR.

test/sync.js
```diff
@@ -11,211 +11,229 @@
...
+	assert.equal(data.title, 'The Tempest');
+	assert.equal(data.author, 'William Shakespeare');
+	assert.equal(data.length, 123);
   });
 
-  test("update with emulateHTTP and emulateJSON", 7, function() {
-    library.first().save({id: '2-the-tempest', author: 'Tim Shakespeare'}, {
+  QUnit.test("update with emulateHTTP and emulateJSON", function(assert) {
+    assert.expect(7);
+	library.first().save({id: '2-the-tempest', author: 'Tim Shakespeare'}, {
```
akre54: Fix this spacing. (what happened here?)

test/sync.js
```diff
@@ -11,211 +11,229 @@
...
+	assert.equal(this.ajaxSettings.data._method, 'PUT');
+	var data = JSON.parse(this.ajaxSettings.data.model);
+	assert.equal(data.id, '2-the-tempest');
+	assert.equal(data.author, 'Tim Shakespeare');
+	assert.equal(data.length, 123);
+  });
+
+  QUnit.test("update with just emulateHTTP", function(assert) {
+    assert.expect(6);
+	library.first().save({id: '2-the-tempest', author: 'Tim Shakespeare'}, {
```
akre54: Same here. Regex issue?

test/sync.js
```diff
@@ -11,211 +11,229 @@
...
+    assert.expect(3);
+	library.first().save({id: '2-the-tempest', author: 'Tim Shakespeare'});
+	library.first().destroy({
       emulateHTTP: true,
       emulateJSON: true
     });
-    equal(this.ajaxSettings.url, '/library/2-the-tempest');
-    equal(this.ajaxSettings.type, 'POST');
-    equal(JSON.stringify(this.ajaxSettings.data), '{"_method":"DELETE"}');
+	assert.equal(this.ajaxSettings.url, '/library/2-the-tempest');
```
akre54: de-indent

test/events.js
akre54: jsl is [available from npm](https://www.npmjs.com/package/jsl). Tests don't have a style guide, we enforce style through PRs and code maintenance (if you have any questions, just ask. most should be pretty readily apparent from the surrounding code).

platinumazure:
I'll fix indenting and style changes sometime this week. Any indent deviations were due to my not having a very good handle on indenting in Vim in my earlier file updates-- now that I have that down, it should be pretty easy to fix. Thanks!

platinumazure:
Okay, I think I've fixed everything. There were a few small style things that I left in for expedient patching, but these were corrections to follow existing style (e.g., explicit semicolons) rather than attempting to impose my own idea of line spacing or chained code. Hopefully these will be acceptable.

test/view.js
```diff
@@ -2,9 +2,9 @@
 
   var view;
 
-  module("Backbone.View", {
+  QUnit.module("Backbone.View", {
 
-    setup: function() {
+    beforeEach: function(assert) {
```
captbaritone: Do we need this `assert` callback if it's never used?

platinumazure: No, but it does no harm leaving it there. It can be useful if `beforeEach`
has an async operation.

captbaritone:
Commit is too large to open see all of it in GitHub's diff view and I cannot comment inline. That being said, I spent about an hour looking through the diff this evening and this is the closest thing I could find to "problem". Looks good!

`collection.js` new line number 1335: Missing indentation.

platinumazure:
Should I split this into multiple pull requests?

akre54:
Nope, looks great. Thanks.

jridgewell:
:clap:

PR #3818, opened 2015-10-05
Document using [] for Collections with zero models

captbaritone:
*Updated:*

The documented example shows passing `[]` to instantiate an empty collection
while still supplying an options object. This change makes the preceding
documentation consistent with that example.

I also noticed that we don't have any tests covering the previously recommended
approach of passing `null` (or any falsey value) for models. I've added a test
to ensure we continue to support that usage.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1822,7 +1822,7 @@ alert(library.get('dvd-1').id + ' ' + library.get('vhs-1').id);
       <b>initialize</b> function, it will be invoked when the collection is
       created. There are a couple of options that, if provided, are attached to
       the collection directly: <tt>model</tt> and <tt>comparator</tt>.<br />
-      Pass <tt>null</tt> for <tt>models</tt> to create an empty Collection with <tt>options</tt>.
+      Pass <tt>[]</tt> for <tt>models</tt> to create an empty Collection with <tt>options</tt>.
     </p>
 
 <pre>
--- a/test/collection.js
+++ b/test/collection.js
@@ -1437,6 +1437,22 @@ QUnit.test("#1894 - `sort` can optionally be turned off", function(assert) {
     assert.ok(collection.comparator === comparator);
   });
 
+  QUnit.test("Pass falsey for `models` for empty Col with `options`", function(assert) {
+    assert.expect(9);
+    var opts = {a: 1, b: 2};
+    _.forEach([undefined, null, false], function(falsey) {
+      var Collection = Backbone.Collection.extend({
+        initialize: function(models, options) {
+          assert.strictEqual(models, falsey);
+          assert.strictEqual(options, opts);
+        }
+      });
+
+      var col = new Collection(falsey, opts);
+      assert.strictEqual(col.length, 0);
+    });
+  });
+
   QUnit.test("`add` overrides `set` flags", function(assert) {
     var collection = new Backbone.Collection();
     collection.once('add', function (model, collection, options) {


index.html
```diff
@@ -1827,7 +1827,7 @@ <h2 id="Collection">Backbone.Collection</h2>
 
 <pre>
 var tabs = new TabSet([tab1, tab2, tab3]);
-var spaces = new Backbone.Collection([], {
+var spaces = new Backbone.Collection(null, {
```
akre54: Let's have the documented way be an empty array. We gracefully support falsy values.

test/collection.js
```diff
@@ -1355,6 +1355,14 @@
     ok(collection.comparator === comparator);
   });
 
+  test("Pass `null` for `models` for empty Col with `options`", 2, function() {
```
akre54: Could be falsey.

index.html
captbaritone: Sounds good. I'll open a second PR to change the docs rather than the example.

akre54: Use this PR. Just amend your commit

captbaritone: :+1:

test/collection.js
captbaritone: Should I discard the test, since it's no-longer relevant to the PR?

akre54: Or add `false` and `undefined` as options.

captbaritone: I've added `false`, `undefined` and `0`. Let me know if `0` is overdoing it a bit.

akre54: 0 is probably overdoing it. Keep it with false, undefined, and null, and I'll merge. Thanks!

captbaritone: Updated. Glad I could help. Thanks for your work.

captbaritone:
Updated.

captbaritone:
Updated to use custom `MyModel` instead of `Backbone.Model`. Thanks for catching that. Since `Backbone.Model` is the default model, the test was passing even if `options` was not passed. Now the test only passes when `options` is supplied.

index.html
```diff
@@ -1822,7 +1822,7 @@ <h2 id="Collection">Backbone.Collection</h2>
       <b>initialize</b> function, it will be invoked when the collection is
       created. There are a couple of options that, if provided, are attached to
       the collection directly: <tt>model</tt> and <tt>comparator</tt>.<br />
-      Pass <tt>null</tt> for <tt>models</tt> to create an empty Collection with <tt>options</tt>.
+      Pass <tt>[]</tt> for <tt>models</tt> to create an empty Collection with <tt>options</tt>.
```
akre54: I think it's fine to leave this as null (sorry missed this before).

test/collection.js
```diff
@@ -1437,6 +1437,17 @@ QUnit.test("#1894 - `sort` can optionally be turned off", function(assert) {
     assert.ok(collection.comparator === comparator);
   });
 
+  QUnit.test("Pass falsey for `models` for empty Col with `options`", function(assert) {
+    assert.expect(6);
+    _.forEach([undefined, null, false], function(falsey) {
+      var MyModel = Backbone.Model.extend();
+      var collection = new Backbone.Collection(falsey, {model: MyModel});
+
+      assert.ok(collection.model === MyModel);
```
akre54: How about testing the options instead of checking that it attaches properties? (This is currently potentially testing two behaviors).

```js
var opts = {a: 1, b: 2}
_.forEach([undefined, null, false], function(falsey) {
  var Collection = Backbone.Collection.extend({
    initialize: function(options) {
      assert.strictEqual(options, opts);
    }
  });

  var col = new Collection(falsey, opts);
  assert.strictEqual(col.length, 0);
});

```

captbaritone: Good idea. I've updated the code to use your example with one small change: I had to add the `models` argument to the `initialize()` method. Since I had the value, I figured I'd test it as well. Let me know if that feels excessive.

akre54: good catch :)

akre54:
Great. Thanks.

captbaritone:
@akre54 I'm a little confused by this commit https://github.com/jashkenas/backbone/commit/bb2c1b96183d0dd92d64b8ef06b7ba4d76f274db. 

My initial PR standardized the inconsistent documentation and example on `null` but you [commented](https://github.com/jashkenas/backbone/pull/3818#discussion_r41322545).

> Let's have the documented way be an empty array. We gracefully support falsy values.

So I undated the PR to use `[]` as the standard. You then merged my PR but followed up with https://github.com/jashkenas/backbone/commit/bb2c1b96183d0dd92d64b8ef06b7ba4d76f274db which brings us back to `null`.

We've satisfied my goal of consistency between documentation and example, so I have no real problem, but I'm a bit confused as to why we're back to `null`.

akre54:
That was my bad -- I thought I wrote a comment earlier. `[]` is obvious, `null` is clearer. `[]` doesn't need to be explicitly spelled out.

captbaritone:
@akre54 Got it. Thanks for the clarification!

PR #3820, opened 2015-10-07
Deopt Underscore methods that change on arguments

jridgewell:
All of our currently deoptimized Underscore methods (`pick`, `omit`, `invoke`, `without`, and `difference`) change behavior based on the number of arguments pass to the call. This both adds `reduce` (and friends).

Supersedes (and incorporates) #3810.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -87,6 +87,9 @@
         return _[method](this[attribute], cb(iteratee, this), context);
       };
       case 4: return function(iteratee, defaultVal, context) {
+        if(arguments.length < 2) {
+          return _[method](this[attribute], cb(iteratee, this));
+        }
         return _[method](this[attribute], cb(iteratee, this), defaultVal, context);
       };
       default: return function() {
--- a/backbone.js
+++ b/backbone.js
@@ -87,9 +87,6 @@
         return _[method](this[attribute], cb(iteratee, this), context);
       };
       case 4: return function(iteratee, defaultVal, context) {
-        if(arguments.length < 2) {
-          return _[method](this[attribute], cb(iteratee, this));
-        }
         return _[method](this[attribute], cb(iteratee, this), defaultVal, context);
       };
       default: return function() {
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -68,6 +68,9 @@
   // form param named `model`.
   Backbone.emulateJSON = false;
 
+  // A cached regex to detect if the Underscore function contains the `arguments`
+  // keyword.
+  var containsArguments = /arguments/;
```
akre54: this has to be `/\barguments\b/`, no?

backbone.js
```diff
@@ -76,6 +79,10 @@
   //
   // `Function#apply` can be slow so we use the method's arg count, if we know it.
   var addMethod = function(length, method, attribute) {
+    // In case we don't catch it, deopt Underscore functions that contain
+    // the arguments keyword.
+    if (length > 0 && containsArguments.test(_[method])) length = 0;
```
akre54: I didn't know `Re::test` automatically `toString`'d functions. Neat.

akre54:
This is a little gross but so is most of the perf rain dancing we're doing around those parts.

backbone.js
jridgewell: Yup. It'll soon be configurable by one of the well-known symbols.

backbone.js
jridgewell: Updated.

backbone.js
akre54: eh?

jridgewell: Scratch that, I thought `#test` was included in http://www.2ality.com/2015/09/well-known-symbols-es6.html#forwarding_calls_from_string_methods_to_their_parameters.

akre54:
>  explicitly deopts any method that depends on `arguments` from falling through the cracks

Do we really need to do this though?

just-boris:
That is need because raw underscore methods behave so.
Why optimisation of `reduce`-family methods matters? This methods are applied for collection once per time and are not called so often

jridgewell:
> Do we really need to do this though?

Necessarily, no. But'll it catch our screw-ups next time. And it might increase interop with underscore-like libraries that use `arguments` for other reasons.

akre54:
> underscore-like libraries that use arguments for other reasons

The only library this could refer to is Lo-dash. And afaik he doesn't do anything with `arguments`. Unless I'm mistaken?

jridgewell:
> The only library this could refer to is Lo-dash.

Not specifically, this could be any handwritten library.

> And afaik he doesn't do anything with arguments. Unless I'm mistaken?

I think lodash is pretty in-line with underscore on the functions Backbone uses.

jridgewell:
Anyone else object?

backbone.js
```diff
@@ -1166,8 +1173,8 @@
   // Underscore methods that we want to implement on the Collection.
   // 90% of the core usefulness of Backbone Collections is actually implemented
   // right here:
-  var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 4,
-      foldl: 4, inject: 4, reduceRight: 4, foldr: 4, find: 3, detect: 3, filter: 3,
+  var collectionMethods = { forEach: 3, each: 3, map: 3, collect: 3, reduce: 0,
```
akre54: Can you separate out these changes from the toString check? These should definitely get merged in.

jridgewell: Done.

Issue #3620, opened 2015-05-21
View Events Property Problem


LordMidi:
I've updated our project and noticed a different behaviour. With the update from 1.1.2 to 1.2.0 I can't set the events property of a view in the "initialize" function anymore - or at least the callbacks don't get triggered. Do I set the events directly as the view property in the extend context it still works. But I need to create the selectors for the events dynamically, therefore I build the events property in "initialize".

Any ideas?

jridgewell:
Ah, seems in all the commits for #3003, we removed the `#delegateEvents` call after calling `#initialize`.

As best as I can follow, it was [removed](https://github.com/akre54/backbone/commit/c5c0d2ef0c50be1468c470f48cf1b2fd6005b593) because of an early implementation of `#setElement` in that PR had to do [argument juggling](https://github.com/akre54/backbone/commit/986540f46f07d53257a293013938863f7b083a48), which wasn't the cleanest API. However, that implementation was [rewritten](https://github.com/akre54/backbone/commit/7180d5ce0ae68fa538ea0d1f278a2636bd403bc4), so I don't see why this can't be [supported again](https://github.com/jashkenas/backbone/blob/1.1.2/backbone.js#L1044-L1050). @akre54?

akre54:
I think the relevant pull is #3060. You can also build the events in the `events` method:

```js
var View = Backbone.View.extend({
  events: function() {
    return getEventsFor(this);
  }
});

```

jridgewell:
> I think the relevant pull is #3060.

:+1: Makes sense to me, just wanted to make sure this wasn't an oversight in #3003.

jridgewell:
Also, you can use `#delegate` [directly](https://github.com/jashkenas/backbone/blob/1.2.0/backbone.js#L1271-L1273).

marian-r:
I am extending Views with mixins dynamically inside the initialize method, which now does not work if some event handler is defined in mixin. Although, I'm OK with calling `delegateEvents()` method manually, it would be nice to document this BC break.

akre54:
What do you mean mixins?

marian-r:
As far as I know Mixin is a pattern when you insert functions from one object into another. Maybe I'm wrong about the naming. Situation with event handlers which I was talking about is the following:
```js
var myMixin = {
    eventHandler: function () {
        console.log("I'm from mixin");
    }
};

var MyView = Backbone.View.extend({
    events: {
        'click #myButton': 'eventHandler'
    },
[...]

PR #3827, opened 2015-10-20
Add a preinitialize method to allow for true instance properties and ES6 classes

benmccormick:
This is my attempt to allow for better support for ES6 classes through a new method, localProps.

`localProps` allows you to define a function (or an object), that is evaluated in the constructor for Models/Collections/Views/Routers, with the result being mixed in to each local object.

As a happy side effect this allows a declarative format for instance properties on an object (instead of using `this.foo = 'bar'` statements in the constructor).

This allows us to write something like 

```
class DocumentRow extends Backbone.View {

    localProps() {
        return {
          tagName:  "li",
          className: "document-row",
          events: {
            "click .icon":          "open",
            "click .button.edit":   "openEditDialog",
            "click .button.delete": "destroy"
          }
        };
    }

    initialize() {
        this.listenTo(this.model, "change", this.render);
    }

    render() {
        //...
    }
}
```

instead of adding properties to the prototype separate from the class declaration or using non-standard features like decorators.

See #3560 for more background on why this is useful/necessary

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -394,6 +394,7 @@
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
+    this.preinitialize.apply(this, arguments);
     this.cid = _.uniqueId(this.cidPrefix);
     this.attributes = {};
     if (options.collection) this.collection = options.collection;
@@ -421,6 +422,10 @@
     // The prefix is used to create the client id which is used to identify models locally.
     // You may want to override this if you're experiencing name clashes with model ids.
     cidPrefix: 'c',
+
+    // preinitialize is an empty function by default. You can override it with a function
+    // or object.  preinitialize will run before any instantiation logic is run in the Model.
+    preinitialize: function(){},
 
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
@@ -756,6 +761,7 @@
   // its models in sort order, as they're added and removed.
   var Collection = Backbone.Collection = function(models, options) {
     options || (options = {});
+    this.preinitialize.apply(this, arguments);
     if (options.model) this.model = options.model;
     if (options.comparator !== void 0) this.comparator = options.comparator;
     this._reset();
@@ -784,6 +790,11 @@
     // The default model for a collection is just a **Backbone.Model**.
     // This should be overridden in most cases.
     model: Model,
+
+
+    // preinitialize is an empty function by default. You can override it with a function
+    // or object.  preinitialize will run before any instantiation logic is run in the Collection.
+    preinitialize: function(){},
 
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
@@ -1221,6 +1232,7 @@
   // if an existing element is not provided...
   var View = Backbone.View = function(options) {
     this.cid = _.uniqueId('view');
+    this.preinitialize.apply(this, arguments);
     _.extend(this, _.pick(options, viewOptions));
     this._ensureElement();
     this.initialize.apply(this, arguments);
@@ -1243,6 +1255,10 @@
     $: function(selector) {
       return this.$el.find(selector);
     },
+
+    // preinitialize is an empty function by default. You can override it with a function
+    // or object.  preinitialize will run before any instantiation logic is run in the View
+    preinitialize: function(){},
 
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
@@ -1469,6 +1485,7 @@
   // matched. Creating a new one sets its `routes` hash, if not set statically.
   var Router = Backbone.Router = function(options) {
     options || (options = {});
+    this.preinitialize.apply(this, arguments);
     if (options.routes) this.routes = options.routes;
     this._bindRoutes();
     this.initialize.apply(this, arguments);
@@ -1483,6 +1500,10 @@
 
   // Set up all inheritable **Backbone.Router** properties and methods.
   _.extend(Router.prototype, Events, {
+
+    // preinitialize is an empty function by default. You can override it with a function
+    // or object.  preinitialize will run before any instantiation logic is run in the Router.
+    preinitialize: function(){},
 
     // Initialize is an empty function by default. Override it with your own
     // initialization logic.
--- a/test/collection.js
+++ b/test/collection.js
@@ -661,6 +661,31 @@
     assert.equal(coll.one, 1);
   });
 
+  QUnit.test('preinitialize', function(assert) {
+    assert.expect(1);
+    var Collection = Backbone.Collection.extend({
+      preinitialize: function() {
+        this.one = 1;
+      }
+    });
+    var coll = new Collection;
+    assert.equal(coll.one, 1);
+  });
+
+  QUnit.test('preinitialize occurs before the collection is set up', function(assert) {
+    assert.expect(2);
+    var Collection = Backbone.Collection.extend({
+      preinitialize: function() {
+        assert.notEqual(this.model, FooModel);
+      }
+    });
+    var FooModel = Backbone.Model.extend({id: 'foo'});
+    var coll = new Collection({}, {
+      model: FooModel
+    });
+    assert.equal(coll.model, FooModel);
+  });
+
   QUnit.test('toJSON', function(assert) {
     assert.expect(1);
     assert.equal(JSON.stringify(col), '[{"id":3,"label":"a"},{"id":2,"label":"b"},{"id":1,"label":"c"},{"id":0,"label":"d"}]');
--- a/test/model.js
+++ b/test/model.js
@@ -63,6 +63,36 @@
     assert.equal(model.get('value'), 2);
   });
 
+
+  QUnit.test('preinitialize', function(assert) {
+    assert.expect(2);
+    var Model = Backbone.Model.extend({
+
+      preinitialize: function() {
+        this.one = 1;
+      }
+    });
+    var model = new Model({}, {collection: collection});
+    assert.equal(model.one, 1);
+    assert.equal(model.collection, collection);
+  });
+
+  QUnit.test('preinitialize occurs before the model is set up', function(assert) {
+    assert.expect(6);
+    var Model = Backbone.Model.extend({
+
+      preinitialize: function() {
+        assert.equal(this.collection, undefined);
+        assert.equal(this.cid, undefined);
+        assert.equal(this.id, undefined);
+      }
+    });
+    var model = new Model({id: 'foo'}, {collection: collection});
+    assert.equal(model.collection, collection);
+    assert.equal(model.id, 'foo');
+    assert.notEqual(model.cid, undefined);
+  });
+
   QUnit.test('parse can return null', function(assert) {
     assert.expect(1);
     var Model = Backbone.Model.extend({
--- a/test/router.js
+++ b/test/router.js
@@ -95,6 +95,10 @@
       '*anything': 'anything'
     },
 
+    preinitialize: function(options) {
+      this.testpreinit = 'foo';
+    },
+
     initialize: function(options) {
       this.testing = options.testing;
       this.route('implicit', 'implicit');
@@ -179,6 +183,11 @@
   QUnit.test('initialize', function(assert) {
     assert.expect(1);
     assert.equal(router.testing, 101);
+  });
+
+  QUnit.test('preinitialize', function(assert) {
+    assert.expect(1);
+    assert.equal(router.testpreinit, 'foo');
   });
 
   QUnit.test('routes (simple)', function(assert) {
--- a/test/view.js
+++ b/test/view.js
@@ -61,6 +61,28 @@
     assert.strictEqual(new View().one, 1);
   });
 
+  QUnit.test('preinitialize', function(assert) {
+    assert.expect(1);
+    var View = Backbone.View.extend({
+      preinitialize: function() {
+        this.one = 1;
+      }
+    });
+
+    assert.strictEqual(new View().one, 1);
+  });
+
+  QUnit.test('preinitialize occurs before the view is set up', function(assert) {
+    assert.expect(2);
+    var View = Backbone.View.extend({
+      preinitialize: function() {
+        assert.equal(this.el, undefined);
+      }
+    });
+    var _view = new View({});
+    assert.notEqual(_view.el, undefined);
+  });
+
   QUnit.test('render', function(assert) {
     assert.expect(1);
     var myView = new Backbone.View;


jridgewell:
I'm not sure I like this any better than the [property methods](https://github.com/jashkenas/backbone/issues/3560#user-content-methods) approach...

benmccormick:
@jridgewell hey Justin.  I get it because I don't think its perfect/beautiful either.  

This way has a bit less boilerplate than methods for everything, and works for Backbone libs/extensions that rely on properties without using `_.result` (admittedly they probably should be doing that)

I'd love it if somebody came along and tossed out a better solution for a clean syntax with pure ES6 classes.  If this PR is a start for something better then that's a win in my mind.  But if there isn't something better, I think this is an improvement.

akre54:
> I'm not sure I like this any better than the property methods approach...

Agreed. And this has the added deficit of not sharing objects across prototypes. An app with a lot of objects is going to balloon in memory very quickly.

I've been declaring ES2016 classes broken until we fix prototypes and `super` in the constructor. Backbone's approach (and the long history of others that came before) is much more flexible and sensible.

benmccormick:
> An app with a lot of objects is going to balloon in memory very quickly.

I think this is overstated  How much additional memory overhead is this really going to add in a normal case, if this is for configuration props?.  Again though, is there a better way?

> I've been declaring ES2016 classes broken until we fix prototypes and super in the constructor. Backbone's approach (and the long history of others that came before) is much more flexible and sensible.

I understand this mindset.  But, I'm in favor of allowing devs to use the spec'd classes here for the sake of standardization.  A common class standard allows better tooling (allow things like http://ternjs.net/ to better interpret Backbone constructs for instance) as well as lowering the learning curve when moving between Backbone/Angular 2/ React / etc.  So while I'm very willing to accept that this isn't the best way of doing this (I hope somebody has something better), I don't like the idea of giving up on allowing ES6 classes as a usable option.

jridgewell:
> This way has a bit less boilerplate than methods for everything, and works for Backbone libs/extensions that rely on properties without using _.result (admittedly they probably should be doing that)

It's not much boilerplate. And for things like `Model#defaults` and `View#attributes` it'll actually save developers from themselves.

> Agreed. And this has the added deficit of not sharing objects across prototypes. An app with a lot of objects is going to balloon in memory very quickly.

I'm wary about this argument. Yes they'll be bigger, but I don't think it'll cause huge problems.

> So while I'm very willing to accept that this isn't the best way of doing this (I hope somebody has something better), I don't like the idea of giving up on allowing ES6 classes as a usable option.

I'm actually coming around to this. If we update the code to pass in `options`, I'd be in favor. We can use this to as the sanctioned method to attach properties to the instance _before_ the other methods are called on creation. Kind of like a `#preinitialize`.

I know there are issues requesting this, and that can be solved with this, but my issue searching skills are failing me. Something along the lines of changing the `tagName` based on input options.

Also, it'd help since ~75% of my own code does `_.extend(this, _.pick(options, ['option1', ...])` in my constructor or `#initialize` (depending on where in the creation process I need it).

- - -

So, add `options`, and :+1:.

backbone.js
```diff
@@ -394,6 +394,7 @@
   var Model = Backbone.Model = function(attributes, options) {
     var attrs = attributes || {};
     options || (options = {});
+    _.extend(this, _.result(this, 'localProps'));
```
akre54: We shouldn't need `result` for this, no? There's never a case `localProps` would be an object over just sticking them on the prototype, yeah?

jridgewell: I think forcing it to be a method is the right choice.

blikblum:
How about change the name from localProps to instanceProps ?

backbone.js
benmccormick: My thought was around the secondary use case (in my mind) of adding properties to the instance.  I was thinking a user might want to do that with a few static properties.  But I'm realizing now that the reference would be shared anyway and that would undermine the purpose. So I'm on board with that.

benmccormick:
@jridgewell just making sure I'm understanding you correctly, you want the options passed in to the constructor passed to the localProps/instanceProps function?

@blikblum I'd be ok with instanceProps.  It would also be an opportunity to clarify the current docs.  Right now the model `extend` docs start with this:

>  To create a Model class of your own, you extend Backbone.Model and provide instance properties, as well as optional classProperties to be attached directly to the constructor function.

It's referring to the fact that the first argument to extend is added as instance properties on the prototype, but thats super confusing I think to most people.  Naming this instanceProps and clarifying that extends is mixing data onto the prototype and not an instance would likely be helpful.

jridgewell:
>  just making sure I'm understanding you correctly, you want the options passed in to the constructor passed to the localProps/instanceProps function?

Yup.

> It would also be an opportunity to clarify the current docs

Separate PR?

> I'd be ok with instanceProps

I think this is going in the wrong direction. Defining a more generic `#preinitialize` would cover this use case and avoid semantic arguments about what it should and shouldn't do. That'll solve issues like #3620, #3685, and https://github.com/jashkenas/backbone/pull/3060#issuecomment-119616052.

```js
var Model = function(data, options) {
  var attrs = attributes || {};
  options || (options = {});
  this.preinitialize.apply(this, arguments);
  // ...
  this.initialize.apply(this, arguments);
};
```

benmccormick:
@jridgewell I'm on board with that.  Generalizing it makes sense to me.  Thanks for the wider context.

Ok I'm going to edit the PR to rename this to preinitialize across the board, make it method only, and change it to applying with the constructor options as opposed to extending the prototype.

So my example above will look like:

```
class DocumentRow extends Backbone.View {

    preinitialize() {
        _.extend(this, {
          tagName:  "li",
          className: "document-row",
          events: {
            "click .icon":          "open",
            "click .button.edit":   "openEditDialog",
            "click .button.delete": "destroy"
          }
        });
    }

    initialize() {
        this.listenTo(this.model, "change", this.render);
    }

    render() {
        //...
    }
}
```

Which is more explicit for readers anyway.  I like it.

braddunbar:
So, I don't quite understand how this is different than the following.

```js
class MyView extends Backbone.View {
  constructor (attributes, options) {
    _.extend(this, …)
    super(attributes, options)
    this.listenTo(…)
  }
}
```

I find `initialize` to be problematic as a replacement for calling super, with or without `class` sugar. Adding another lifecycle method as an alternative to calling methods when necessary seems like a step in the wrong direction to me.

jridgewell:
> So, I don't quite understand how this is different than the following.

You can't use `this` before `super()` in ES6 classes. :disappointed:

```js
class MyView extends Backbone.View {
  constructor (attributes, options) {
    _.extend(this, …) // throws error
    super(attributes, options)
    this.listenTo(…)
  }
}

class MyView extends Backbone.View {
  constructor (attributes, options) {
    super(attributes, options)
    _.extend(this, …) // too late, dom element has already been created and events bound.
    this.listenTo(…)
  }
}
```

> I find initialize to be problematic as a replacement for calling super

I actually agree for the current case: `#initialize` should be deprecated when ES6 classes become the norm. But for this case, we can't do a "before constructor" (`#preinitialize), only an "after" (`#initialize`).

braddunbar:
> You can't use this before super() in ES6 classes.

Well that's…surprising. I don't understand why since one can already do `Super.call(this)`, but thanks for telling me! :smile_cat: 

In that case we're left with getters and setters I suppose? I haven't seen them mentioned yet in this thread or #3560.

```js
class MyView extends Backbone.View {
  get tagName () { return '…' }
  …
}
```

I understand it's preferable to declare these things directly on the prototype (I would rather that as well) but I don't know that `preinitialize` is an improvement in that regard.

jridgewell:
> In that case we're left with getters and setters I suppose? I haven't seen them mentioned yet in this thread or #3560.

https://github.com/jashkenas/backbone/issues/3560#issuecomment-91597621. :stuck_out_tongue: 

But that won't work for runtime options:

```js
class View extends Backbone.View {
  constructor(options) {
    super(options);
    _.extend(this, _.pick(options, 'type'));
  }

  tagName() {
    return this.type === 'emote' ? 'span' : 'em';
  }
}
```

Yes, you could pass in `tagName` in `options`, but then your view's logic isn't self contained.

braddunbar:
Thanks @jridgewell! (That's a long thread.) I guess I'll leave this alone then, as I don't have any better ideas. ¯\\\_(ツ)\_/¯

jridgewell:
I was being cheeky, sorry.

> I guess I'll leave this alone then, as I don't have any better ideas.

I don't mean to force my way here, I just think it's a pretty simple solution and it'll work well with code I usually write.

benmccormick:
@jridgewell well I proposed an idea that I wasn't completely happy with in the hopes that it would produce a better idea, and I feel like that happened, so its all good from my POV

backbone.js
```diff
@@ -1192,7 +1203,9 @@
   // Creating a Backbone.View creates its initial element outside of the DOM,
   // if an existing element is not provided...
   var View = Backbone.View = function(options) {
+
```
jridgewell: Random newline?

benmccormick: @jridgewell oops.  Fixed.

akre54:
(replying to @jridgewell) My objection is that with this pull classes now have three places to put create-time logic: the `constructor`, `preinitialize`, and `initialize`. This feels like a hacky solution that is a failure of ES6 classes, and adds extra confusion for those who just want to use normal `extend`.

benmccormick:
@akre54 theres default constructor behavior, and hooks to run code before and after the default behavior.  Or you can override the default behavior.

Thats a reasonable and coherent set of options, and its a common pattern to provide a component that has life cycle functions/events, and hooks to add behavior before or after those events. 

Marionette Views allow you to override `render`, or put stuff in `onBeforeRender` or `onRender`.

React has similar before and after hooks for the life cycle events of their components (componentWillUpdate/componentDidUpdate).

Ember does similar stuff, components have `willInsertElement` and `didInsertElement` events.  

I think people will be able to grasp the difference here if it is documented.

jridgewell:
> My objection is that with this pull classes now have three places to put create-time logic: the `constructor`, `preinitialize`, and `initialize`.

With this, there's never a reason to override `constructor`. Anything you would put there, put in either `preinit` or `init`.

> This feels like a hacky solution that is a failure of ES6 classes, and adds extra confusion for those who just want to use normal extend.

I'm not happy with the ES6 class spec either, but I don't think that should stop us from adding support for it. And this solves issues like https://github.com/jashkenas/backbone/pull/3060#issuecomment-119616052.

megawac:
After thinking about it some more, I like preinitialize a lot as a way to avoid having to manage `constructor` call order. 

Can you add a couple tests @benmccormick

benmccormick:
@megawac Tests added, and passing.  Sorry for the delay, I missed your original message somehow.

megawac:
@jridgewell @akre54 I'm very happy with this PR. Any more criticisms or is it good to merge? When merged can we try to get a 1.4 out asap?

jridgewell:
LGTM, let's merge at the end of the day if no one chimes in.

akre54:
Now that static property support has landed in Chrome, do we need this anymore?

jridgewell:
[Yup](https://github.com/jashkenas/backbone/issues/3560#issuecomment-91601515). Static properties are set _after_ the `super()` call, which means our view has already been created.

akre54:
Right right. I'm [referencing the comment above it](https://github.com/jashkenas/backbone/issues/3560#issuecomment-91597621). (By static properties I mean the [static keyword](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static), not property initializers. Those haven't, to my knowledge, landed in the spec yet).

[Example](http://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015%2Ces2015-loose%2Creact%2Cstage-0%2Cstage-1%2Cstage-2%2Cstage-3&experimental=false&loose=false&spec=false&code=class%20MyView%20extends%20Backbone.View%20%7B%0A%20%20static%20events%20%3D%20%7B%0A%20%20%20%20'click%20.foo'%3A%20'handleClick'%0A%20%20%7D%0A%20%20static%20tagName%20%3D%20function()%20%7B%0A%20%20%20%20return%20this.checked%20%3F%20'div'%20%3A%20'span'%0A%20%20%7D%0A%7D).

jridgewell:
Ohh. I'm kinda partial to this solution over the `instOrStaticVar`. @megawac?

benmccormick:
@akre54 to be clear, in your example, the `static events` part isn't currently part of the spec right?  Thats combining the static keyword with the property initializer proposal.  

If that is right, I'm not sure what has changed here?  You can use methods to define the various things here with or without the static keyword.

typhonrt:
Yes the [class properties](https://babeljs.io/docs/plugins/transform-class-properties/) / [stage-2 proposal](https://github.com/tc39/proposal-class-public-fields). Check out the example @akre54 posted with just the ES2015 preset; [things fail](http://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015&experimental=false&loose=false&spec=false&code=class%20MyView%20extends%20Backbone.View%20%7B%0A%20%20static%20events%20%3D%20%7B%0A%20%20%20%20'click%20.foo'%3A%20'handleClick'%0A%20%20%7D%0A%20%20static%20tagName%20%3D%20function()%20%7B%0A%20%20%20%20return%20this.checked%20%3F%20'div'%20%3A%20'span'%0A%20%20%7D%0A%7D). Now click the presets section to the right of the evaluate checkbox and [add stage-2](http://babeljs.io/repl/#?babili=false&evaluate=false&lineWrap=false&presets=es2015%2Cstage-2&experimental=false&loose=false&spec=false&code=class%20MyView%20extends%20Backbone.View%20%7B%0A%20%20static%20events%20%3D%20%7B%0A%20%20%20%20'click%20.foo'%3A%20'handleClick'%0A%20%20%7D%0A%20%20static%20tagName%20%3D%20function()%20%7B%0A%20%20%20%20return%20this.checked%20%3F%20'div'%20%3A%20'span'%0A%20%20%7D%0A%7D).

benmccormick:
@typhonrt I don't think we're saying different things.  

If I understand correctly static methods are currently *part of the spec*, are supported in the latest version of every major environment and will be supported everywhere eventually.  

As far as I can tell (and it seems you agree) static properties are a *proposal* that may or may not eventually end up in the spec, and can currently be used through Babel.  

I understand how static properties potentially solve the problem of this PR.  I don't understand how static methods do.

typhonrt:
Indeed, though at stage-2 / draft it's quite likely to [proceed onward](http://www.2ality.com/2015/11/tc39-process.html) and is a fairly popular feature. An article to peruse on [ES6+ with React](https://babeljs.io/blog/2015/06/07/react-on-es6-plus); see the "property initializers" / "arrow functions" sections.

davidmpaz:
Hi,

is there any ETA to get this PR via #4028 merged? This is more than one year old already and from what I read from related threads and posts, most people agreed on while not the perfect solution provides necessary support for current specifications.

I would like to start using Backbone since it fits better than a full bloated framework like Angular or Vue.js but the fact that I can not have members (properties) initialized in my constructors, because of problems stated already, is hindering me severely.

Could you please decide on adding it?

Thanks in advance
David

benmccormick:
@davidmpaz I've resigned myself to this possibly never getting released.  I don't think Backbone is getting significant attention from any of the maintainers these days (see recent issue activity).  I'd recommend working off a fork if this is super important to you, or if you just need the data modeling, using a different data library like [MobX](https://mobx.js.org/).

davidmpaz:
I thought on introduce it bit by bit to have time to check how is accepted, starting with Views using it also with TypeScript by the way. Just wanted to have goodies (bit more structure, events, jQuery scoped access to DOM) from Backbone.View.

But having your comments in mind as well as issues activity I will probably quit, I can not risk on a fork for this. I really don't like one man orchestra shows :) and I also don't want to rely on decorators as long as they don't became standard, which is the alternative I like the most for this situation.

I will discuss this with colleagues and see what we can do. I was strongly hoping to not end re-implementing wheels by my self ;) Backbone was looking like a perfect match for our needs.

Thanks for the hints

PR #3833, opened 2015-10-22
Fix 1.2.2 Model#id regression

### LINKED ISSUE #3778: Potential regression when setting `id` directly from 1.2.1 to 1.2.2
As part of upgrading from 1.2.1 to 1.2.2, we noticed an issue in our app where we were directly setting `id` on a model. I don't think this is the best way to set the id, but it is what it is.

```js
model.id = SOME_ID
```

After setting additional properties via `set` later in our code, `model.id` started to return `undefined`.

Below is a minimal example. The expected behavior is that both `console.log`'s will output "123". In 1.2.2 calling `set` is wiping out `model.id` (appears to be from this commit: https://github.com/jashkenas/backbone/commit/be7fc959641a3ccaa362110023ea7d697bc3f93b)

You can run these fiddles and see the output in the console.

1.2.1: https://jsfiddle.net/jktms1xx/2/
```sh
> 123
> 123
```

1.2.2: https://jsfiddle.net/g1am1otw/1/
```sh
> 123
> undefined
```

---

We've resolved this issue by changing our code to call `model.set('id', SOME_ID)` - but we were wondering if setting `model.id` directly should continue to work. Thanks!

cc @jbrad

jridgewell:
Fixes #3778.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -506,7 +506,7 @@
       }
 
       // Update the `id`.
-      this.id = this.get(this.idAttribute);
+      if (this.idAttribute in attrs) this.id = this.get(this.idAttribute);
 
       // Trigger all relevant attribute changes.
       if (!silent) {


braddunbar:
:+1:

akre54:
Can we get a test case for this?

PR #3834, opened 2015-10-22
Fix _onModelEvent regression

### LINKED ISSUE #3662: .trigger('change')

### LINKED ISSUE #3717: Manually triggering event on model that is part of collection causes error
Hello -- we were upgrading backbone in our app and ran into an unexpected error.  Here is the simplest repro for it:
```
var c = new Backbone.Collection();
var m = new Backbone.Model();

c.add(m);
m.trigger('change');  //  error thrown
```
Was there an intentional change in how events are handled?

Thanks!

jridgewell:
Fixes #3662. Fixes #3717

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1148,14 +1148,16 @@
     // events simply proxy through. "add" and "remove" events that originate
     // in other collections are ignored.
     _onModelEvent: function(event, model, collection, options) {
-      if ((event === 'add' || event === 'remove') && collection !== this) return;
-      if (event === 'destroy') this.remove(model, options);
-      if (event === 'change') {
-        var prevId = this.modelId(model.previousAttributes());
-        var id = this.modelId(model.attributes);
-        if (prevId !== id) {
-          if (prevId != null) delete this._byId[prevId];
-          if (id != null) this._byId[id] = model;
+      if (model) {
+        if ((event === 'add' || event === 'remove') && collection !== this) return;
+        if (event === 'destroy') this.remove(model, options);
+        if (event === 'change') {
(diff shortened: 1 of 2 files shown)


megawac:
Could you add a test

backbone.js
```diff
@@ -1148,14 +1148,16 @@
...
     _onModelEvent: function(event, model, collection, options) {
-      if ((event === 'add' || event === 'remove') && collection !== this) return;
-      if (event === 'destroy') this.remove(model, options);
-      if (event === 'change') {
-        var prevId = this.modelId(model.previousAttributes());
-        var id = this.modelId(model.attributes);
-        if (prevId !== id) {
-          if (prevId != null) delete this._byId[prevId];
-          if (id != null) this._byId[id] = model;
+      if (model) {
```
megawac: should we do `model instanceof Backbone.Model` instead?

jridgewell: I don't think it'll be necessary. The issue was people triggering `change` without the required parameters (just `trigger('change')`).

jridgewell:
Test added.

QuotableWater7:
thanks guys! :+1:

QuotableWater7:
I just looked at the code - does this change mean that triggering `change` manually would fail silently?

QuotableWater7:
ah my bad, the change makes sense to me now - thanks again.

PR #3837, opened 2015-10-26
Always return removed array from #_removeModels

### LINKED ISSUE #3835: Collection.remove() Requires Updated Documentation or Fix
According to the documentation for `collection.remove()`:

"Remove a model (or an array of models) from the collection, and return them."

It seems the part about returning the removed models is not always the case in the latest stable release, 1.2.3 (as well as the Edge version). What I expected based on the documentation is to get back an array, always, even if the array empty. (This would allow me to do things such as `collection2.add(collection1.remove(someArrayOfModels))` without having to worry about the return type for `remove()`) The problem for me is that `remove()` is actually returning different value types depending on what I pass in as arguments, which you will see below how this can become an issue.

Open up a browser console with Backbone loaded and attempting the following:

```
// Start with this for each test case
var collection = new Backbone.Collection([new Backbone.Model(), new Backbone.Model()]);
console.log(collection.length); //should be 2
```

Now trying the following operations (be sure to re-create the collection each time):
```
collection.remove(collection.at(0)); //Returns a single model instance
```
```
collection.remove([ collection.at(0), collection.at(1)]) //Returns an array of two model instances
```
```
collection.remove() //Returns 'undefined'
```
```
collection.remove([]) //Returns 'false'
```
The last one where I get `false` from `collection.remove([])` is where this really becomes an issue, since when I do `collection.add(false)` it adds a new item to the collection, even though I didn't want to add anything. (try it, and you'll see the collection's `length` property will increase by 1)

Some recommendations I have to resolve this:

- Update the documentation to clarify the various return types that `remove()` returns back
- Update `collection.remove()` to always return an array, regardless of input arguments
- Update `collection.add()` to not add anything if `false` is passed as the first argument

jridgewell:
Fixes #3835.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -809,7 +809,7 @@
       var singular = !_.isArray(models);
       models = singular ? [models] : _.clone(models);
       var removed = this._removeModels(models, options);
-      if (!options.silent && removed) this.trigger('update', this, options);
+      if (!options.silent && removed.length) this.trigger('update', this, options);
       return singular ? removed[0] : removed;
     },
 
@@ -1117,7 +1117,7 @@
         removed.push(model);
         this._removeReference(model, options);
       }
-      return removed.length ? removed : false;
+      return removed;
     },
 
     // Method for checking whether an object should be considered a model for
--- a/test/collection.js
+++ b/test/collection.js
@@ -325,7 +325,7 @@
   });
 
   QUnit.test("remove", function(assert) {
-    assert.expect(11);
+    assert.expect(12);
     var removed = null;
     var result = null;
     col.on('remove', function(model, col, options) {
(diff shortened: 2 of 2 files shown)


akre54:
:+1:

Cleod9:
@jridgewell Before we wrap this up, I might be wrong on this but based on the source code's `singular` variable if you were to remove a single model by passing in `remove(model)` wouldn't the function end up returning that single model instead of an array?

jridgewell:
> wouldn't the function end up returning that single model instead of an array?

That's intentional. I meant we shouldn't be returning `false`, only the single model or the array of removed  (or none) models.

Cleod9:
Ok, just wanted to confirm. If it's intentional and documented I'm good. Thanks for the fix!

PR #3840, opened 2015-10-28
Deopt Underscore methods that use `arguments`

jridgewell:
To prevent issues like #3810, we can test each underscore method for the presence of `arguments`. If found, we explicitly deoptimize the wrapping method, so that the call arguments exactly match the wrapper's.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -68,6 +68,9 @@
   // form param named `model`.
   Backbone.emulateJSON = false;
 
+  // A cached regex to detect if the Underscore function contains the `arguments`
+  // keyword.
+  var containsArguments = /\barguments\b/;
   // Proxy Backbone class methods to Underscore functions, wrapping the model's
   // `attributes` object or collection's `models` array behind the scenes.
   //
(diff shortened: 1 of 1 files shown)


akre54:
:-1: this is a sledgehammer solution to a minor (and currently non-existent) problem. I don't like the precedence merging this would set.

backbone.js
```diff
@@ -76,6 +79,10 @@
   //
   // `Function#apply` can be slow so we use the method's arg count, if we know it.
   var addMethod = function(length, method, attribute) {
+    // In case we don't catch it, deopt Underscore functions that contain
+    // the arguments keyword.
+    if (length > 0 && containsArguments.test(_[method])) length = 0;
```
megawac: No need for length>0 and I would create the regex on this line

`if (/\barguments\b/.test(_[method])) length = 0;`

-1 on this overall. I'd prefer we just be more careful with this in the future. Instead change this patch to set `reduce: 0, reduceRight: 0`

jridgewell: > Instead change this patch to set reduce: 0, reduceRight: 0

https://github.com/jashkenas/backbone/pull/3820

jridgewell:
Two thumbs down so far. Root issues is fixed, so I'm fine with leaving this one alone.

PR #3843, opened 2015-10-30
Avoid Object.prototype collisions with defaults

### LINKED ISSUE #3842: Object.prototype properties conflict with `_.defaults`
I run into an issue today, as Firefox added support for a new `watch` function to each Object.prototype, while I was unfortunate to use 'watch' as a name of an attribute of my Model.
The problem is caused by Model's constructor line:
```
attrs = _.defaults({}, attrs, _.result(this, 'defaults'));
```
and the fact that lodash's _.defaults implementation sees that `typeof {}.watch` is not `'undefined` so it does not want to override it with the one found in `attrs`.
Not sure if the issue is with Backbone, Lodash, or Firefox, but one thing is sure: while I can (and do) avoid today's magic properties (like `constructor`, `valueOf` etc.) I can not really anticipate what word will be forbiden in future, and don't like the idea of updating my app in future whenever one of my favorite words get's banned.

jridgewell:
This is definitely an underscore concern, but it's easier to demo in Backbone code.

Fixes #3842.

### DIFF
--- a/test/model.js
+++ b/test/model.js
@@ -34,6 +34,12 @@
     assert.equal(model.collection, collection);
   });
 
+  QUnit.test('Object.prototype properties are overridden by attributes', function(assert) {
+    assert.expect(1);
+    var model = new Backbone.Model({hasOwnProperty: true});
+    assert.equal(model.get('hasOwnProperty'), true);
+  });
+
   QUnit.test('initialize with attributes and options', function(assert) {
     assert.expect(1);
     var Model = Backbone.Model.extend({
@@ -55,19 +61,6 @@
     });
     var model = new Model({value: 1}, {parse: true});
     assert.equal(model.get('value'), 2);
-  });
-
-  QUnit.test('initialize with defaults', function(assert) {
-    assert.expect(2);
-    var Model = Backbone.Model.extend({
-      defaults: {
-        firstName: 'Unknown',
-        lastName: 'Unknown'
-      }
-    });
-    var model = new Model({'firstName': 'John'});
-    assert.equal(model.get('firstName'), 'John');
-    assert.equal(model.get('lastName'), 'Unknown');
   });
 
   QUnit.test('parse can return null', function(assert) {
@@ -428,7 +421,7 @@
   });
 
   QUnit.test('defaults', function(assert) {
-    assert.expect(4);
+    assert.expect(9);
     var Defaulted = Backbone.Model.extend({
       defaults: {
         one: 1,
@@ -438,6 +431,9 @@
     var model = new Defaulted({two: undefined});
     assert.equal(model.get('one'), 1);
     assert.equal(model.get('two'), 2);
+    model = new Defaulted({two: 3});
+    assert.equal(model.get('one'), 1);
+    assert.equal(model.get('two'), 3);
     Defaulted = Backbone.Model.extend({
       defaults: function() {
         return {
@@ -449,6 +445,15 @@
     model = new Defaulted({two: undefined});
     assert.equal(model.get('one'), 3);
     assert.equal(model.get('two'), 4);
+    Defaulted = Backbone.Model.extend({
+      defaults: {hasOwnProperty: true}
+    });
+    model = new Defaulted();
+    assert.equal(model.get('hasOwnProperty'), true);
+    model = new Defaulted({hasOwnProperty: undefined});
+    assert.equal(model.get('hasOwnProperty'), true);
+    model = new Defaulted({hasOwnProperty: false});
+    assert.equal(model.get('hasOwnProperty'), false);
   });
 
   QUnit.test('change, hasChanged, changedAttributes, previous, previousAttributes', function(assert) {
@@ -989,8 +994,8 @@
 
   QUnit.test('`previous` for falsey keys', function(assert) {
     assert.expect(2);
-    var model = new Backbone.Model({0: true, '': true});
-    model.set({0: false, '': false}, {silent: true});
+    var model = new Backbone.Model({'0': true, '': true});
+    model.set({'0': false, '': false}, {silent: true});
     assert.equal(model.previous(0), true);
     assert.equal(model.previous(''), true);
   });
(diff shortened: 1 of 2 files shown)


test/model.js
```diff
@@ -428,27 +421,24 @@
...
     Defaulted = Backbone.Model.extend({
-      defaults: function() {
-        return {
-          "one": 3,
-          "two": 4
-        };
-      }
+      defaults: { hasOwnProperty: true, }
     });
-    model = new Defaulted({two: undefined});
```
akre54: Why'd you get rid of this test case?

akre54:
I dunno. If this is a problem with `_.defaults` we should fix it there.

test/model.js
jridgewell: It's testing the same thing as the one above it, no?

akre54: No... it's testing that passing `undefined` is thrown away, not that the key isn't passed at all.

test/model.js
```diff
@@ -34,6 +34,12 @@
     assert.equal(model.collection, collection);
   });
 
+  QUnit.test("initialize with attributes", function(assert) {
```
akre54: Something more descriptive? Like `Object.prototype values are respected`?

test/model.js
akre54:
```js
var defaults = {foo: true};

_.defaults({}, defaults, {foo: undefined}); // { foo: true }
_.defaults({}, defaults, {bar: undefined}); // { foo: true, bar: undefined }

_.extend({}, defaults, {foo: undefined}); // { foo: undefined }
_.extend({}, defaults, {bar: undefined}); // { foo: true, bar: undefined }
```

jridgewell: > No... it's testing that passing undefined is thrown away, not that the key isn't passed at all.

Isn't that [this test](https://github.com/jridgewell/backbone/blob/defaults-prototype-collision/test/model.js#L425-L433)?

> `var defaults = {foo: true};`

With my "defaults" implementation:

```js
defaults = {foo: true};

defaults({}, defaults, {foo: undefined}); // { foo: true }
defaults({}, defaults, {bar: undefined}); // { foo: true, bar: undefined }
```

The only difference with mine is it won't respect properties that are already on the `dest` object:

```js
_.defaults({foo: false}, {foo: true}); // { foo: false }
defaults({foo: false}, {foo: true}); // { foo: true }
```

akre54: > Isn't that this test?

Sorry, missed that in the diff. 

> The only difference with mine is it won't respect properties that are already on the dest object

Then I'm confused. That changes _.default's behavior completely, no?

jridgewell: We're only using defaults with an "empty" destination object, so it works the same. Externally, this wouldn't be acceptable.

akre54: Yeah but I dunno. Either we should fix this in Underscore or we shouldn't fix this at all. What's good for the goose is good for the gander...

jridgewell: If we made it return a new instance instead of modifying the destination, it would work perfectly. We're kind of headed down that path with https://github.com/jashkenas/underscore/pull/2232.

jdalton:
I think it's fine to have this in Backbone temporarily until it can be addressed in a released version of Underscore.

jdalton:
For reference here's the fix in lodash https://github.com/lodash/lodash/commit/8a3842b788e920e905827bebc82447c9ce97b29d. It preserves the existing defaults iteration order and behavior of not overwriting non-undefined values.

akre54:
I dunno. I feel like this is something that should be fixed in Underscore, if at all. Who names their keys `hasOwnProperty`? Attribute keys and collection ids really shouldn't be arbitrary like this.

jdalton:
> I dunno. I feel like this is something that should be fixed in Underscore, if at all. Who names their keys `hasOwnProperty`?

It's less those classic `Object.prototype` method names and more of the new ones like `watch` in Firefox. I'm cool with punting on it for now as an option too.

akre54:
To be sure, `watch` is a pretty dumb decision on FF's part, but it's not insurmountable. Unless there's a really good reason I'd rather not introduce a subtle backwards compatibility bug like this.

jdalton:
Yep, the likely issues I see are `watch`, `unwatch`, `constructor`, `toString` and `toSource`, but since this hasn't been tackled yet (several years in) I can see this as lower priority.

jridgewell:
Repinging this: I've updated the PR to use `_.extends` and `_.defaults` to work around the conflicts. What does everyone else think?

jridgewell:
Ping @akre54 since you're reviewing things today. :wink:

akre54:
Hows the perf hit when there are a few keys?

jridgewell:
http://jsperf.com/defaults-extend/2

jridgewell:
Ping. I want to get this one in.

akre54:
For the record, I think this (`_.defaults(_.extend({}, defaults, attrs), defaults)`) is a really ugly hack for a minor to nonexistent problem, but the perf hit isn't terrible. Hope this is the last of it.

test/model.js
```diff
@@ -989,8 +994,8 @@
 
   QUnit.test('`previous` for falsey keys', function(assert) {
     assert.expect(2);
-    var model = new Backbone.Model({0: true, '': true});
-    model.set({0: false, '': false}, {silent: true});
+    var model = new Backbone.Model({'0': true, '': true});
```
akre54: Whys this needed? JS automatically quotes strings

jridgewell: It was a liter warning for inconsistent string usage. Must have committed it by accident. :relaxed:

captbaritone: Linting rules are not perfect :disappointed:

jridgewell: All good, they're much better now than before.

Sendoushi:
Hi, 

I've  chosen some  interesting links you may find helpful,  here is the full list <http://crintyfesta.ballerblogger.com/e4vnuj>

Alberto Martinez

PR #3845, opened 2015-11-02
Test case for model.id

jridgewell:
Re: https://github.com/jashkenas/backbone/pull/3833#issuecomment-152842054


akre54:
This seems like a terrible behavior, but fine, if it's consistent.

jridgewell:
We can kill it in V2. :wink:

PR #3850, opened 2015-11-05
fixed typo

adictovirtual:
fixed typo in the examples


akre54:
This file comes from [Backbone.LocalStorage](https://github.com/jeromegn/Backbone.localStorage). Might be worth fixing there too.

PR #3851, opened 2015-11-05
missing () invocation

adictovirtual:
missing history () invocation

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1839,7 +1839,7 @@
   });
 
   // Create the default Backbone.history.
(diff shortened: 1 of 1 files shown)


captbaritone:
The `()` is optional when not passing any arguments, so this comes down to a style choice.

I did a quick check and this is the only time we construct and object with no arguments, so whichever way we go, it's guaranteed to be consistent. ;)

akre54:
Yup yup. Looks a little cleaner when we don't need it. Thanks @captbaritone

PR #3862, opened 2015-11-20
Performance with stopListening

ashlux:
When listening to a lot of models, stopListening(model) is very expensive due to checking whether listeningTo is empty. I've added a test that demonstrates the problem. I might be missing it, but I see no benefit to just leaving _listeningTo = {}.

```javascript
if (_.isEmpty(listeningTo)) this._listeningTo = void 0; 
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -242,7 +242,6 @@
 
       listening.obj.off(name, callback, this);
     }
-    if (_.isEmpty(listeningTo)) this._listeningTo = void 0;
 
     return this;
   };
--- a/test/events.js
+++ b/test/events.js
@@ -158,6 +158,27 @@
     b.trigger('event2');
   });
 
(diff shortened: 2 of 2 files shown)


ashlux:
Anyone have any input on this?

megawac:
Agreed. Remove all `console.info` calls

akre54:
We don't have any perf tests for the rest of the app. Any reason to do it here?

megawac:
Yeah, not a fan of the tests but I think that line can be dropped from backbone.js

akre54:
Yeah works for me. @ashlux, want to cut it down to the single line and we'll merge?

ashlux:
See https://github.com/jashkenas/backbone/pull/3883 instead.

akre54:
In the future, just amend your pull request (with `git rebase -i`). No need to open another pull request.

PR #3864, opened 2015-11-24
Change `_.invoke` to `_.map`.

jdalton:
This makes way for `_.invoke` to be renamed `_.mapInvoke` or other in lodash v4 (due out in Jan 2016).

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1011,7 +1011,9 @@
 
     // Pluck an attribute from each model in the collection.
     pluck: function(attr) {
-      return _.invoke(this.models, 'get', attr);
+      return _.map(this.models, function(model) {
+        return model.get(attr);
+      });
     },
 
     // Fetch the default set of models for this collection, resetting the


akre54:
Why did this get merged @jridgewell? Backbone has never been aligned with Lo-Dash's release schedules.

Also, I'd like to state categorically that I believe @jdalton's actions in regard to my comments in lodash/lodash#1658 to be ethically wrong and incredibly poor form. Deleting and editing comments and suppressing differing opinions is not the way to behave in open source or in life. I'm posting here so he can't edit or delete this.

jridgewell:
> Why did this get merged @jridgewell? Backbone has never been aligned with Lo-Dash's release schedules.

Same as https://github.com/jashkenas/backbone/pull/3747, https://github.com/jashkenas/backbone/pull/3732, https://github.com/jashkenas/backbone/pull/3678, and https://github.com/jashkenas/backbone/pull/3675. It doesn't hurt anything in our codebase, and will only help devs using Backbone.

akre54:
Those all came in the past few months, initiated by John-David. We don't gain anything by sticking to Lo-Dash's releases, and in fact we lose a great deal of readability and flexibility. If he wants to maintain compatibility that's fantastic.

No comments on the rest of my post?

jridgewell:
> We don't gain anything by sticking to Lo-Dash's releases

We're not sticking to any of his releases? Backbone's not gonna release a new version just for this, or any of the other PRs.

> and in fact we lose a great deal of readability and flexibility

Most of them are 1 lines changes that are frankly more idiomatic in JS (`Array` has `#some` not `#any`, `Function` has `#bind`). This one's no different. And we help some dev down the line who wants to use the latest lodash with the latest Backbone. I see that as a win, even if we added two lines.

> No comments on the rest of my post?

Every time you two are in a thread, it gets heated. I personally think it colors you against his ideas. But lodash is de-facto his and he's free to oversee it that way. It would be inappropriate in Underscore, though.

jridgewell:
As a side, we could make this even better with `return this.map(attr);`

jdalton:
> As a side, we could make this even better with `return this.map(attr);`

Oh snao!

megawac:
@jridgewell that won't work as `map` iterates over models not model attributes

jdalton:
@megawac The unit tests pass. I'll open a PR and if it's incorrect we should add a test to ensure it catches it.

PR #3865, opened 2015-11-24
Add #pluck to collectionMethods

jridgewell:


megawac:
`Collection#map` is over models not model attributes

megawac:
Sorry I may of been mistaken. Really sorry about that @jridgewell

jridgewell:
Damn, it won't work directly for `#pluck`. But @jdalton opened https://github.com/jashkenas/backbone/pull/3866 that cleans this up.

jdalton:
> it won't work directly for `#pluck`.

Do you have any insight into why it doesn't work?

jridgewell:
Because we would have to pass down a special `(model) => model.get(attr)` function into `_.pluck`, but `_.pluck` internally calls `_.property` coercing the getter into it's source string.

PR #3867, opened 2015-11-24
Ensure `attr` is a string when passed to `this.map`.

jdalton:
This PR ensures `attr` is a string when passed to `this.map`. I've had folks rely on the undocumented behavior of letting iteratees slip through if `attr` is not coerced.


jridgewell:
That's fine with me, but I'll defer until another :+1:.

PR #3875, opened 2015-12-01
Fixed Collection#set regression when parse returns falsy value

### LINKED ISSUE #3871: Collection.set behavior changed in 1.2.3
This change https://github.com/jashkenas/backbone/commit/a082d2fdf9d0509ae7082296e41a0b7efe6878d2#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR797
changes the behavior if collection's `parse` method return falsy value.

Before: empty collection
After: collection with 1 "blank" model

Is it and intended change? My personal consideration is that having 1 empty model in this case is unexpected and unwanted behavior.

Flamefork:
Restored previous code for this logic.
Added a regression test.

Fixes #3871

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -824,7 +824,7 @@
       if (options.parse && !this._isModel(models)) models = this.parse(models, options);
 
       var singular = !_.isArray(models);
-      models = singular ? [models] : models.slice();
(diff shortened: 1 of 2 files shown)


craigmichaelmartin:
:+1:

PR #3884, opened 2015-12-11
make _validate respect silent option

craigmichaelmartin:
This is useful both for `isValid` - where the validity of a model can easily be ascertained without sounding the invalid alarm - as well as for `set` where silence is respected even when validating.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -718,13 +718,16 @@
     },
 
     // Run validation against the next complete set of model attributes,
-    // returning `true` if all is well. Otherwise, fire an `"invalid"` event.
+    // returning `true` if all is well. Otherwise, return false and
+    // fire an `"invalid"` event (if not silent).
     _validate: function(attrs, options) {
       if (!options.validate || !this.validate) return true;
       attrs = _.extend({}, this.attributes, attrs);
       var error = this.validationError = this.validate(attrs, options) || null;
       if (!error) return true;
-      this.trigger('invalid', this, error, _.extend(options, {validationError: error}));
+      if (!options.silent) {
+        this.trigger('invalid', this, error, _.extend(options, {validationError: error}));
+      }
       return false;
     }
 
--- a/test/model.js
+++ b/test/model.js
@@ -1367,6 +1367,31 @@
     assert.ok(model.isValid());
   });
 
+  QUnit.test('#3884 - _validate respects silent option so isValid can return false without triggering invalid.', function(assert) {
+    assert.expect(2);
+    var model = new Backbone.Model();
+    model.validate = function() { return 'some problem'; };
+    model.on('invalid', function() {
+      throw 'invalid';
+    });
+    assert.ok(!model.isValid({silent: true}));
+    assert.throws(model.isValid);
+  });
+
+  QUnit.test('#3884 - _validate respects silent option so set can be silent even with validation.', function(assert) {
+    assert.expect(1);
+    var model = new Backbone.Model();
+    model.validate = function() { return 'some problem'; };
+    model.on('invalid', function() {
+      throw 'invalid';
+    });
+    model.on('change:foo', function() {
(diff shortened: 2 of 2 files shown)


will-mooney-3:
:+1:

jridgewell:
:+1:, someone else want to weigh in?

akre54:
I'm :+1: anytime we can move towards fewer side effects, but aren't we trying to get rid of the `silent` option? Maybe this should take the inverse? i.e. `{trigger: true}`?

craigmichaelmartin:
I'm happy to implement `trigger` instead for `_validate`, but then we introduce a state where events are not triggered by default. Currently all events fire unless silenced, right? This would create the opposite state where now some events are only fired when `trigger`ed. EG, model `set` would accept a `trigger` (respected in `_validate` for firing `invalid`) and a `silent` (respected for not firing `change` and `sort`). I really like the idea of replacing silent with trigger, but only if its codebase-wide. Let me know how you want me to proceed.

akre54:
The only place I know of where we require `trigger` by default is in [`Router#navigate`](http://backbonejs.org/#Router-navigate).

> Currently all events fire unless silenced, right?

Most cases where we have a `silent` option are in methods like `Collection#set` and `Model#set`, which are mutative and meant to have side effects.

I mostly don't like that the name `isValid()` seems to imply that it returns a boolean and that calling it is side effect-free, when in fact it triggers an event on failure. We could split out the `_validate` method into `_validate` and `_validateAndTrigger`, say, but I'm :-1: on adding another silent flag.

jridgewell:
> The only place I know of where we require trigger by default is in Router#navigate.

That everyone hates... Trigger by default.

akre54:
For sure. I think it's the wrong default in that instance.

In my idea scenario, `set` and friends would trigger an event, `isValid()` wouldn't.

craigmichaelmartin:
1. I completely agree that `isValid` should be silent by default. This PR was intended to be non breaking. Are you ok with this breaking change? #3896 is a backward incompatible implementation for this problem.

2. Are you ok with model `set` and friends being able to trigger an event even with `{silent: true})`?

Both of these were addressed with options being considered in _validate.

xerona:
:+1: lgtm

jashkenas:
I think there's a reason why `silent` isn't being checked here — probably because it would muddle cases where the `silent` is being passed for something else.

For example — you want to set a model's attributes silently, but you do *not* want to set an invalid model.

Probably not worth the breaking change.

akre54:
A very good point

On Sun, Dec 20, 2015, 08:02 Jeremy Ashkenas <notifications@github.com>
wrote:

> I think there's a reason why silent isn't being checked here — probably
> because it would muddle cases where the silent is being passed for
> something else.
>
> For example — you want to set a model's attributes silently, but you do
> *not* want to set an invalid model.
>
> Probably not worth the breaking change.
>
> —
> <https://github.com/jashkenas/backbone/pull/3884#issuecomment-166115129>.
>
-- 

Adam K (mobile)

craigmichaelmartin:
> probably because it would muddle cases where the `silent` is being passed for something else.

Yeah, though I think `silent` is already muddled - it's not intuitive as to what events are supposed to be silenced; in your example, why the `change` would be silent, but not the `invalid`).

I'd be happy to kill the silent option if that was not meant tongue in cheek. 

That said, a specific use case that started this was to find a way for `isValid` to be called without side-effects. Maybe something like a purer `isValid` and a loud `ensureValid`?

craigmichaelmartin:
What is the status of this PR as a candidate for 1.4? I think it is the right step until (hopefully) [removing](https://github.com/jashkenas/backbone/pull/3906) the silent option altogether in 2.0 :fingers-crossed: 😄

jgonggrijp:
Closing in favor of #3906.

PR #3885, opened 2015-12-11
trust npm to set up $PATH

craigmichaelmartin:
npm automatically sets up $PATH to look in node_modules/.bin


megawac:
Yep, karma used to not set the `bin: {}` stuff to let npm to identify the clis. Want to send the same patch in for underscore?

craigmichaelmartin:
Interesting! Sure thing!

PR #3886, opened 2015-12-12
Update wording to "space-separated" to be consistent

ClimbsRocks:
This is both grammatically correct, and consistent with the language in the rest of the comments.


ClimbsRocks:
it's a super minor change, but just checking in on this!

PR #3892, opened 2015-12-16
Non-controversial ESLint whitespace rules

captbaritone:
In an attempt to bring our ESLint rules closer to those in Underscore, I've
started by adding some whitespace rules which don't involve any significantly
opinionated changes.


akre54:
:+1:

PR #3893, opened 2015-12-16
Add ESLint rule: key-spacing

captbaritone:


akre54:
I dunno. I like lining up properties and RHS expressions.

@jridgewell except for the really uncontroversial ones, can we leave these open a little longer for debate?

captbaritone:
There are other linting rules which support enforcing lined up properties: http://eslint.org/docs/rules/key-spacing.html#2-vertically-align-values-align-value

I gave them a try and they seem to try to force that style of alignment in a number places that don't make sense. They also get very confused with things like:

       var modelMethods = {keys: 1, values: 1, pairs: 1, invert: 1, pick: 0,
           omit: 0, chain: 1, isEmpty: 1};

I think the upshot is we can have an enforced rule or we can allow for lined up properties, but not both, which is a shame, since in some places the aligned values look very nice.

jridgewell:
I didn't think this was controversial... Sorry.

test/collection.js
```diff
@@ -202,7 +202,7 @@
   QUnit.test("add; at should add to the end if the index is out of bounds", function(assert) {
     assert.expect(1);
     var col = new Backbone.Collection([{id: 2}, {id: 3}]);
-    col.add(new Backbone.Model({id: 1}), {at:   5});
```
akre54: These I'm fine with, but the vertically aligned ones would be nice to have.

test/router.js
```diff
@@ -74,28 +74,28 @@
...
-      "contacts":                   "contacts",
-      "contacts/new":               "newContact",
-      "contacts/:id":               "loadContact",
-      "route-event/:arg":           "routeEvent",
-      "optional(/:item)":           "optionalItem",
-      "named/optional/(y:z)":       "namedOptional",
-      "splat/*args/end":            "splat",
-      ":repo/compare/*from...*to":  "github",
-      "decode/:named/*splat":       "decode",
+      "noCallback": "noCallback",
```
akre54: Can we revert this one?

backbone.js
```diff
@@ -1425,9 +1425,9 @@
   var methodMap = {
     'create': 'POST',
     'update': 'PUT',
-    'patch':  'PATCH',
```
akre54: Can we revert this too? Or does it break the linter?

captbaritone: We would have to either disable the rule, or set it to only warn (and then tolerate the fact that it's always going to throw warnings).

I'm fine either way. If we want to do either of the above, just let me know and I'll open a PR which reverts only the cases where the values were aligned.

PR #3894, opened 2015-12-16
Large batch of ESLint cleanups

captbaritone:
In an attempt to batch these changes together, this pull request includes a number of new rules (all taken from Underscore's `.eslintrc`).

Each commit reflects the changes needed for a single rule, so I would recommend reviewing by commit.

@akre54, @megawac c4b27bd May be controversial in the same way that https://github.com/jashkenas/backbone/pull/3893 was.

If we want some but not all of these, I'm happy to go back to submitting them as individual pull requests.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -4,7 +4,10 @@
     "node": true,
     "amd": true
   },
-
+  "globals": {
+    "attachEvent": true,
+    "detachEvent": true
+  },
   "rules": {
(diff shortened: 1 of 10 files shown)


megawac:
:+1:

captbaritone:
Since `no-multi-spaces` seems a bit debatable (#3899), I've rebased this pull request to not include that change.

PR #3895, opened 2015-12-17
Remove unused method mapping cases

Shutnik:


megawac:
I'm = to this, but I kind of like them for consistency

akre54:
Yeah let's leave em in for consistency's sake. I'd argue the alternative would be more confusing.

Shutnik:
@megawac @akre54 This is dead code, there are no exampes and arguments named using magic ball. Why we can't comment that lines for ’consistency‘ and exchlude them from minificated version? Why we can't revert that megre request when real usecases will appear?

jridgewell:
> Why we can't revert that megre request when real usecases will appear?

We won't remember, the de-opt path (`default`) will catch the methods using 2 and 4 parameters.

PR #3896, opened 2015-12-17
remove side effects from isValid

craigmichaelmartin:
A different (and better, and breaking) implementation solving for a subset of https://github.com/jashkenas/backbone/pull/3884 based on @akre54's feedback.


jgonggrijp:
Closing in favor of #3906.

PR #3898, opened 2015-12-18
Add ESLint rule: wrap-iife

captbaritone:
This pull request is a sub-set of #3894.


akre54:
It looks like we can [use the default](http://eslint.org/docs/rules/wrap-iife) of "outside" to keep the current style (which I personally prefer).

captbaritone:
@akre54 Sure. This will bring us away from parity with Underscore. I don't personally have any problem with that, but is worth mentioning. Should we consider switching Underscore to the "outside" style as well to maintain consistency, or is stasis more valuable than consistency?

akre54:
This is the height of bikeshedding so I'd say it's fine to leave it. We can be consistent in other, more important, areas :)

captbaritone:
Quick clarification: The style we are using in the test is actually the "inside" style, not "outside".

jridgewell:
I prefer our style, `(function() {})();`

PR #3899, opened 2015-12-18
Add ESLint rule: no-multi-spaces

captbaritone:
This pull request is a sub-set of #3894.

### DIFF
--- a/test/model.js
+++ b/test/model.js
@@ -163,7 +163,7 @@
     assert.ok(!a.isNew(), "any defined ID is legal, negative or positive");
     a = new Backbone.Model({'foo': 1, 'bar': 2, 'baz': 3, 'id': 0});
     assert.ok(!a.isNew(), "any defined ID is legal, including zero");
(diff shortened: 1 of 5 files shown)


test/model.js
```diff
@@ -163,7 +163,7 @@
     assert.ok(!a.isNew(), "any defined ID is legal, negative or positive");
     a = new Backbone.Model({'foo': 1, 'bar': 2, 'baz': 3, 'id': 0});
     assert.ok(!a.isNew(), "any defined ID is legal, including zero");
-    assert.ok( new Backbone.Model({          }).isNew(), "is true when there is no id");
```
akre54: This is separate from the rest (but a possible good use of this rule...) I'll make the change separately now.

akre54:
I'm not a fan of this change, it's a little too greedy with cases that actually should be permissible. I wonder if ESLint could have a more fine-grained option than this.

test/model.js
captbaritone: :+1:

PR #3904, opened 2015-12-20
Add ESLint rule: no-unused-expressions

captbaritone:
We could add this rule with no changes to the code if we also added the "allowShortCircuit" option. However, we have a mixture of:

    a = a || default;

And:

    a || (a = default)

Only the first option avoids the "unused expression", and I personally find it more readable, so I've proposed the standardization.


akre54:
Let's leave this one be. I prefer our style.

PR #3906, opened 2015-12-21
kill the `silent` option

craigmichaelmartin:
This does away with the`silent` option in deference to passing flag options for the event callbacks to look at. This is a result of @jashkenas's [feedback](https://github.com/jashkenas/backbone/pull/3884#issuecomment-166115129). In order to reduce boilerplate code, some flags are automatically passed to the callback where the default behavior currently relies on `silent`. I'll update tests if this approaches looks reasonable.

Open to feedback!

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -479,7 +479,6 @@
 
       // Extract attributes and options.
       var unset      = options.unset;
-      var silent     = options.silent;
       var changes    = [];
       var changing   = this._changing;
       this._changing = true;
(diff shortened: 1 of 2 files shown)


megawac:
I have mixed feelings about killing this option. It can be useful when numerous [VirtualCollections](https://github.com/p3drosola/Backbone.VirtualCollection) are spawned from some Collection and you don't want to rebuild the indexes of each virtual collection.

I don't think the options replacing `silent` in this pr improve the code or the API

jaapz:
What is especially better about watching for flags in your callbacks instead of just passing the silent option? I don't see why it's inherently better to do it in the proposed new way. This would be quite an API break and I don't see why it would be removed in favor of this other way that as far as I can see is not a major improvement.

What is the rationale behind this change?

craigmichaelmartin:
Thanks for the feedback! @jaapz My thought is that it ensures consistently and without exception that state changes occur, making the system more intuitive and extendable. It also forces component logic to be encapsulated back in the component, doing away with dictating a sweeping assumption for all units. Lastly, though I'm not sure would be valuable, it would allow finer grain hooks (ie, the previously permanent silent events).

This at least has been the gains from my experience. I have found the `silent` option to be a temptatious anti-pattern, that is convenient at the outset, but does not scale well, and neither enhances understanding of my units, compared with more specific flags. (Ironically, the only place I've [wanted](https://github.com/jashkenas/backbone/pull/3884) [it](https://github.com/jashkenas/backbone/pull/3896) is for what sounds like should be a side-effect free `isValid` - but which does not use it, and where I saw how vague `silent` is and un-intuitively respected - not out of bad implementation, but simply that `silent` is too general). This has just been my experience, though, and I'm very open to those more familiar with the event driven paradigm.

@megawac I'm a little confused. The same could still be accomplished, but you're opposed to having to add `if(options && options.silent) return;` to do so?

craigmichaelmartin:
Perhaps a more succinct articulation: models should be unaware of the views, and so should not have the confidence to act exceptionally and not trigger an event. Instead, they can pass along flags describing the event for whoever is out there telling them they probably don't care.

craigmichaelmartin:
Hm, I see this was [done](https://github.com/jashkenas/backbone/pull/2113) by @braddunbar for Model#set and [restored](https://github.com/jashkenas/backbone/commit/fe251a477da230f26c317a5b72905641ece39fef) by @jashkenas soon after. Is there some background on this?

prantlf:
The `silent` option is a part of the 1.x interface. It would break existing applications. Including mine. Read #2113  for more information.

Issue #3857, opened 2015-11-13
Hash is incorrect in safari,if pass path using unicode string to navigate method 


leexiaosi:
In Safari , pass a path using unicode string to `navigate` method  like this

``` js
Backbone.history.navigate('/city/北京',{trigger:true});
```

The hash is incorrect. It output

``` js
location.hash
//> #/city/%17%AC
```

I tracked the code, it caused by `_updateHash` method.

The call stack created by  `History.prototype.navigate` method like this
- navigate
  -  decodeFragment
  - _updateHash

The fragment passed to `_updateHash`,  was decoded by `decodeFragment`, so the hash was incorrect。

I also searched the commit history, the `decodeFragment` added in "1.2.0"

PS: My english is not good, although I am trying to improve

jridgewell:
Seems to be working for me. Can you provide a jsfiddle showing the error?

leexiaosi:
[https://jsfiddle.net/leexiaosi/uyp3o6ek/](https://jsfiddle.net/leexiaosi/uyp3o6ek/)

**notice!!!** : please open in safari.

### the record of safari
![safari](https://cloud.githubusercontent.com/assets/516562/11164567/62f9ddf6-8b2f-11e5-8aa8-631c1aa84422.gif)

### the record of chrome
![chrome](https://cloud.githubusercontent.com/assets/516562/11164583/c66d1b82-8b2f-11e5-9d88-ecba16ae6167.gif)

paulfalgout:
This seems fairly similar to what is described here: https://github.com/jashkenas/backbone/issues/3830

rrott:
Looks like Safari are buggy in the implementation of assignment to window.location.hash =(
北京 in unicode should be %u53**17**%u4E**AC** but not **%17%AC**

I faced the same issue with Ukrainian chars as well:
>[Log] cityName is : їє (index.html, line 35)
>[Log] locationHash is : #city/WT (index.html, line 36)

It affects all non-latin chars

addinig encodeURI() in _updateHash() before assigning a new hash fixes the issue in Safari, but in all other browsers links will look like 
/index.html#city/%E5%8C%97%E4%BA%AC

>location.hash = '#' + encodeURI(fragment);
https://github.com/jashkenas/backbone/blob/master/backbone.js#L1835

I am not sure if it's ok to send such pull requests =(

rrott:
Found a better solution. Will update you once tests are done and green

okovpashko:
[...]

Issue #3830, opened 2015-10-21
History._updateHash with correctly encoded fragment breaks URI in Safari Browser


jirgn:
## FYI
When coding my first Backbone / Marionette Project I run into a strange URI encoding Problem. This means I am far from being a Backbone Professional and maybe I'm getting something wrong.

## Problem Description
I'm having the following route in my Marionette Router, that handles searches with GET style query parameters.

    var Router = Marionette.AppRouter.extend({
        appRoutes: {
            'search(?*query)' : 'search'
        }
    });

When trying to persist my Application State, I need the URL Fragment to be able to handle special Characters like ö,ä,ü correctly. Therefore I encode the Query (e.g. ?loc-city[]=Berlin&loc-cty[]=München) and call

    router.navigate('search?loc-city%5B%5D=M%C3%BCnchen&loc-city%5B%5D=Berlin');
 
This works perfectly in most Browsers. But when testing on Safari, I encountered, that in the navigation routine, the fragment will be decoded and then given to Backbone.History._updateHash, which updates the windows location. At this Point Safari (Desktop and Mobile Version) encodes the given fragment in ISO and the query then will be

    ?loc-city[]=Berlin&loc-city[]=M%FCnchen // notice the %FC which is the url encoded iso char

which cannot be consumed by Javascript/Backbone anymore, and break the Code with an URIError (when calling decodeUriComponent in Backbone.Router._extractParameters).

Because I find not much Information on the Web, I call for your help here, if there's an obvious bug anywhere, or if i got something wrong, please give me a hint. 
I already checked the encoding on all sides, the documents are utf-8, the server delivers the correct encoding as well and the headers also say so.

jridgewell:
@braddunbar, mind taking a look?

jridgewell:
@jirgn: Looked into this today, and I'm not able to replicate. I'll reopen if you can provide a jsfiddle that causes the error.

PR #3909, opened 2015-12-26
fixed Safari related issue with wrong data put to location.hash

rrott:
Fix for the issues https://github.com/jashkenas/backbone/issues/3857  and https://github.com/jashkenas/backbone/issues/3830
I'm not sure if everything was covered by tests previously, but changing the code does not affected old tests.
I've added Safari related test case to make them green when karma runs with safari.
> ./node_modules/.bin/karma start --browsers=Safari

The only problem now is that location.hash in Safari is encoded while in other browsers it is the same as in url. But anyway it works with no errors.
safari

chrome

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1801,7 +1801,7 @@
       // If hash changes haven't been explicitly disabled, update the hash
       // fragment to store history.
       } else if (this._wantsHashChange) {
-        this._updateHash(this.location, fragment, options.replace);
+        this._updateHash(this.location, fragment);
         if (this.iframe && fragment !== this.getHash(this.iframe.contentWindow)) {
           var iWindow = this.iframe.contentWindow;
 
@@ -1813,7 +1813,7 @@
             iWindow.document.close();
           }
 
-          this._updateHash(iWindow.location, fragment, options.replace);
+          this._updateHash(iWindow.location, fragment);
         }
 
       // If you've told us that you explicitly don't want fallback hashchange-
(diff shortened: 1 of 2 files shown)


jridgewell:
@braddunbar: mind reviewing?

rrott:
Can someone please have a look?

jaapz:
Am I right that this PR removes the possibility to choose between replacing and not replacing the current hash?

Because as far as I can see, replace is now always called, which means the user cannot go back in the history anymore. I don't think that is acceptable behavior.

okovpashko:
Have you tried to fail your test? 
Because the [test](https://github.com/jashkenas/backbone/blob/master/test/router.js#L194) for unicode symbols has already been written and it passes in Safari. 
As I figured out window.location is [emulated](https://github.com/jashkenas/backbone/blob/master/test/router.js#L19) by `<a>` HTML element and assigning not URL encoded hash to it is worked correctly.

Flamefork:
I believe this could be fixed by #4025 fix.

Passing not encoded UTF string to `location.hash` is treated by Safari differently from other browsers, but  probably this usage should be avoided anyway. Passing URLEncoded string to `location.hash` and to `Backbone.history.navigate` behaves the same now:

This (changed from the original) test passes using current `master`:
```javascript
  QUnit.test('#2667 - Hashes with unicode in them. Safari related issue', function(assert) {
    assert.expect(2);
    var route = '#' + encodeURIComponent('北їñ');

    location.replace('http://example.com' + route);
    assert.equal(location.hash, route);

    Backbone.history.navigate(route);
    assert.equal(location.hash, route);
  });
```

jridgewell:
> I believe this could be fixed by #4025 fix.


Sweet.

megawac:
@Flamefork want to send a PR with that test case? Thanks

Flamefork:
@megawac I'm just not sure if it's useful: with the `encodeURIComponent` change, the first check is obvious and doesn't check anything in Backbone, and the second one ~duplicates this one: https://github.com/jashkenas/backbone/blob/master/test/router.js#L1071
What do you think?

PR #3911, opened 2015-12-29
Avoid nested ternary

captbaritone:
While looking into adding the "no-nested-ternary" ESLint rule, I realized that this logic could be refactored a bit.

This nested ternary was introduced in pull request #3875 in order to handle falsy returns from `parse()`.

This refactor avoids the nested ternary and also makes the intent clearer by associating the empty array directly with `parse()`'s return.

It also more narrowly handles the falsy case (we never intended to allow you call `set(0)` for example.)

The only down side I can see is that if `parse()` returns falsy, we will unnecessarily slice/clone the empty array we just created.

(@jridgewell)

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -822,10 +822,12 @@
       if (models == null) return;
 
       options = _.defaults({}, options, setOptions);
-      if (options.parse && !this._isModel(models)) models = this.parse(models, options);
+      if (options.parse && !this._isModel(models)) {
+        models = this.parse(models, options) || [];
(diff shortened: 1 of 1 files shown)


akre54:
Slicing the empty array doesn't seem like a big deal to me (nor a particularly common case). Looks good. Thanks.

jridgewell:
@akre54: Did you mean to merge this?

akre54:
I did :sleeping:

Issue #3905, opened 2015-12-21
Model with id "undefined"


duskpoet:
If you put model with id "undefined" in collection first, .get method of collection will always return this model.

https://jsfiddle.net/duskpoet/rg74hL4p/

jridgewell:
This actually comes from the `#set` model, where we try to determine if `{id: 1}` is in the collection already. The set uses `#get({id: 1})`, which tries `this._byID[model.cid]`, and since the model doesn't have a `cid` property, it fetches the `undefined` model.

Can we [please](https://github.com/jashkenas/backbone/search?q=byId&type=Issues&utf8=%E2%9C%93) split up `byId` and `byCid`?

akre54:
Under what circumstances would you have an id of `"undefined"`?

This sounds like a bug in your code, and there are better places for bugfixes than in Backbone proper.

duskpoet:
I have categories, coming from external API, and one of them is... "undefined". I'd like to keep them as `id`, cause I need to get models from collection by category often.

craigmichaelmartin:
PR https://github.com/jashkenas/backbone/pull/3936 will allow for this. Examples are contrived, until they're not. I'm in favor of [adding tests back](https://github.com/jashkenas/backbone/pull/3953) to ensure backbone is un-opinionated about what is 'reasonable' and ensures this case is tested for.

duskpoet:
There is also a similar case with id: '[object Object]'
https://jsfiddle.net/duskpoet/yg1mnvhz/

craigmichaelmartin:
I've update my PR for this case. Given the nature of accepting such a wide range of values for `get`, i think these are judicious measures. I don't think Adam is going to agree though :P

akre54:
@duskpoet that's really not how Model ids are supposed to be used.

Your best bet is to do `collection.findWhere({category: "undefined"})` for your first case and fix the bug for your second.

PR #3916, opened 2016-01-02
Model with id "undefined" 

spasa:
Add fix get method in collection (commit https://github.com/spasa/backbone/commit/347e8694af6221a45bb0382be7a94d37ef90db2c). This is fix for issue https://github.com/jashkenas/backbone/issues/3905.

As you can see in commit, we first check if obj is model. If it is, then we can check if model is in collection by cid as well.


spasa:
I will close this pull request, because in tests, this error occurred "Newline required at end of file but not found". I will fix this and create a new pull request

PR #3919, opened 2016-01-04
Ability to add multiple events in one line

BigFax:
Add the ability to bind multiple events in one line inside backbone views.

Instead of one event per line to call the same function
```javascript
events: { 
    'click a[data-anchor]': 'scroll',
    'wheel': 'scroll',
    'keydown': 'scroll'
}
```
All events can be written on the same line (separated by commas)
```javascript
events: { 
    'click a[data-anchor], wheel, keydown': 'scroll'
}
```

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1282,11 +1282,13 @@
       if (!events) return this;
       this.undelegateEvents();
       for (var key in events) {
-        var method = events[key];
+        var method = events[key], combinedEvents = key.split(',');
         if (!_.isFunction(method)) method = this[method];
         if (!method) continue;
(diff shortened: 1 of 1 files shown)


akre54:
Dupe of #2611, #1362, and others. Please do a search before opening issues.

BigFax:
It's a request, not an issue :p

I read these conversations before.
But I am not offering the same code, i thought it was a good thing to propose a new version to take care of white spaces as @jashkenas mentioned it in #1362 conversation and with less lines of code.
Does it really affect performances ? I don't think so. 
Why not add a new feature which keep the same behavior in this case ? :)

PR #3920, opened 2016-01-07
Fix undefined attrs in modelId

Frumba:
When I use set method on a model it automatically trigger 'change' event and in my case I don't have previousAttributes.
The problem is that the method _onModelEvent send 'undefined' attrs to the modelId which end by crashing.


jridgewell:
How do you not have previous attributes? That's done [in `#set`](https://github.com/jashkenas/backbone/blob/master/backbone.js#L488).

PR #3927, opened 2016-01-14
Silent true note for collection.remove

megawac:


craigmichaelmartin:
How and where `silent` is applied is unintuitive and inconsistent. How about we [kill it](https://github.com/jashkenas/backbone/pull/3906) :smile: But, yeah, while it exists, this helps illuminate where it is respected :+1:

PR #3930, opened 2016-01-17
add `has` method to collection

craigmichaelmartin:
Adds a `has` method to the collection prototype. Analogous to Model#[has](https://github.com/jashkenas/backbone/blob/master/backbone.js#L451)

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -973,6 +973,11 @@
       return this._byId[obj] || this._byId[id] || this._byId[obj.cid];
     },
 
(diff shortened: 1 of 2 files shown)


captbaritone:
Would the mixed in `contains` not provide this functionality?

captbaritone:
Never mind. ;) I was on my phone an misread the change. This checks by key not by value.

akre54:
Sure, why not.

PR #3932, opened 2016-01-19
remove unnecessary logic

craigmichaelmartin:
I'd also be in favor of changing the comment for `modelId` if there's interest, from `Define how to uniquely identify models in the collection` to `Returns the unique identification of a model for the collection` ..even though that's only half the story since it's stored by cid as well.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1074,7 +1074,7 @@
 
     // Define how to uniquely identify models in the collection.
     modelId: function(attrs) {
-      return attrs[this.model.prototype.idAttribute || 'id'];
+      return attrs[this.model.prototype.idAttribute];
     },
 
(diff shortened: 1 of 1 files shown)


jridgewell:
This one's intentional to support polymorphic `model` factories.

craigmichaelmartin:
Oh, cool! I didn't even know about that :P Thanks for the explanation!

jridgewell:
I can't seem to find the right issue right now, but it's somewhere the `git blame` history...

akre54:
Shouldn't we have a test that fails just in case someone tries to remove it?

jridgewell:
@craigmichaelmartin? :smile:

craigmichaelmartin:
Haha, I was already working on it :)

PR #3933, opened 2016-01-20
add tests solidifying behavior for modelId with polymorphic models

craigmichaelmartin:
cf https://github.com/jashkenas/backbone/pull/3932


akre54:
Thanks

PR #3935, opened 2016-01-24
add a collection modelIdAttribute

craigmichaelmartin:
Define an optional `modelIdAttribute` for collections. If present it is used as the key by which models are uniquely identified in a collection. Useful for when using a model factory.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -381,6 +381,8 @@
   // want global "pubsub" in a convenient place.
   _.extend(Backbone, Events);
 
+  var defaultIdAttribute = 'id';
+
   // Backbone.Model
   // --------------
 
@@ -415,7 +417,7 @@
 
     // The default name for the JSON `id` attribute is `"id"`. MongoDB and
     // CouchDB users may want to set this to `"_id"`.
-    idAttribute: 'id',
+    idAttribute: defaultIdAttribute,
 
     // The prefix is used to create the client id which is used to identify models locally.
     // You may want to override this if you're experiencing name clashes with model ids.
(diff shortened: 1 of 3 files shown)


backbone.js
```diff
@@ -381,6 +381,8 @@
   // want global "pubsub" in a convenient place.
   _.extend(Backbone, Events);
 
+  var defaultIdAttribute = 'id';
```
craigmichaelmartin: If exposed, could be an easy, one-time configuration for couchDB and mongoDB users.

craigmichaelmartin:
Or, `modelIdAttribute` could be defined like:

```javascript
// Returns the key by which the model is uniquely
// identified in the colleciton.
modelIdAttribute: function(attrs) {
  return this.model.prototype.idAttribute || defaultIdAttribute;
}
```

and `modelId` like:

```javascript
// Returns the value by which the model is uniquely
// identified in the colleciton.
modelId: function(attrs) {
  return attrs[_.result(this, 'modelIdAttribute')];
}

jridgewell:
I'm not sure this really gives us anything over providing a `#modelId`, e.g. https://github.com/jashkenas/backbone/issues/3760 is still a problem.

craigmichaelmartin:
Correct, its intent was an easier api for collections with polymorhpic models by exposing the key the collection uses to uniquely identify models. As it is now, one must override the function to change a key.

jridgewell:
> As it is now, one must override the function to change a key.

But your [suggested changes](https://github.com/jashkenas/backbone/pull/3935#issuecomment-174292901) use methods, and now we have two where just one will do. I'd love to simplify the uniqueness issues with a collection, but I don't think this is the right way.

craigmichaelmartin:
Hm, I'm fine with that. I guess I don't see the reason not to offer more hooks than less :P Thanks for your feedback!

PR #3936, opened 2016-01-24
optimize collection get

craigmichaelmartin:


jridgewell:
This looks good to me, but we're slightly changing the returns from prioritizing `cid` over `#modelId`. Someone else want to weigh in?

craigmichaelmartin:
That makes sense. What would you think of something like this?
```javascript
get: function(obj) {
  if (obj == null) return void 0;
  return this._byId[obj] ||
    (this._isModel(obj)
        ? (this._byId[this.modelId(obj.attributes)] || this._byId[obj.cid])
        : this._byId[this.modelId(obj)]);
}
```

akre54:
Yeah I like your second form a bit more. It seems like `get(id)` should be the preferred form, with `get(attrs)` a close second.

jridgewell:
I'm pro second form too, but I'd break it up into an if-else instead of a ternary.

craigmichaelmartin:
Sure thing!

akre54:
How about just

```js
get: function(obj) {
  if (obj == null) return void 0;
  return this._byId[obj] || this._byId[this.modelId(obj.attributes || obj)] || this._byId[obj.cid];
}
```

Can anyone see a need for the `isModel` check?

craigmichaelmartin:
I added `isModel` to account for the edge case of a model with an `undefined` id. A get for an id that doesn't exist would return the undefined model.

craigmichaelmartin:
Hm, this code would also. If it's worth accounting for that, I can update my code. Otherwise, I'll go with what you've proposed.

akre54:
Let's at least add a test for that behavior. My code seems to pass all tests currently, but if it's going to break somewhere we should figure that out.

craigmichaelmartin:
OK, I have the test case illustrating the edge case. I actually described it wrong above, sorry! If I change to what you've proposed it fails. I like what you wrote, but it's up to you whether supporting it is worth the extra call.

jridgewell:
This fixes #3905? Whoot!

akre54:
Oh thats what you meant by `undefined`? I really don't think that's worth fixing (see my comment on #3905), or even testing for.

Any other potential bugs?

jridgewell:
> I really don't think that's worth fixing (see my comment on #3905)

I disagree. `cid`s are a semi private identifier and we shouldn't be looking for it on a normal object. Devs should be passing the `id` instead. This fixes the code to enforce that.

akre54:
3905 is about the string "undefined". Where does that have anything to do with cids?

akre54:
We could split out `_byId` from `_byCid` (we did in the past), but I don't see how that is relevant to the current discussion.

jridgewell:
> #3905 is about the string "undefined". Where does that have anything to do with cids?

The reason `undefined` is a problem id is because we try to grab `cid` from a non-model object (which doesn't contain `cid`).

> We could split out _byId from _byCid (we used to do just this)

I'm also cool with that (and let's get rid of `#cidPrefix` while we're at it), but that'd be another PR.

akre54:
Show me a test?

On Thu, Feb 4, 2016, 12:50 Justin Ridgewell <notifications@github.com>
wrote:

> #3905 <https://github.com/jashkenas/backbone/issues/3905> is about the
> string "undefined". Where does that have anything to do with cids?
>
> The reason undefined is a problem id is because we try to grab cid from a
> non-model object (which doesn't contain cid).
>
> We could split out _byId from _byCid (we used to do just this)
>
> I'm also cool with that (and let's get rid of #cidPrefix while we're at
> it), but that'd be another PR.
>
> —
> <https://github.com/jashkenas/backbone/pull/3936#issuecomment-179964788>.
>
-- 

Adam K (mobile)

craigmichaelmartin:
> 3905 is about the string "undefined". Where does that have anything to do with cids?

Because as you pointed out a few minutes ago on another [PR](https://github.com/jashkenas/backbone/pull/3843/files#r51905979) (:smile:), JS casts keys to strings, so `this._byId[obj.cid]` for an obj without a cid property is equivalent to `this.byId["undefined"]`.
<hr>
I believe the point is: `get` is sloppy in that it consults non-models for a model property. @duskpoet was kind enough to point out one case where this could lead to errors. @jridgewell has articulated this [line by line](https://github.com/jashkenas/backbone/issues/3905#issuecomment-166362336), and I have provided one [example](https://github.com/craigmichaelmartin/backbone/blob/optimize-collection-get/test/collection.js#L109-L112). You are right that this test, and any additional examples (eg, an object with a cid property colliding with a model's cid) would be equally contrived, but nevertheless serve to illustrate the point that `get` should be judicious and understand that cids are a semi private identifiers and not try looking for it on a normal object.

akre54:
> evaluating this._byId[obj.cid] for an obj without a cid property is equivalent to this.byId["undefined"].

But I'm confused. Shouldn't the call through `modelId` catch any of those edge cases for you?

`get`'s contract is to take a string (`id`, `cid`), a model object with `id` or `cid` properties, or an attributes object that is transformed through `modelId`. Under what circumstances would you pass an object that didn't conform to any of those?

I disagree about `cid` being semi private. Sometimes It's the best way to reference an object that hasn't been persisted to storage yet.

jridgewell:
> But I'm confused. Shouldn't the call through modelId catch any of those edge cases for you?

If the model isn't in there, we check for a `cid` regardless of `#modelID`'s return value.

> I disagree about cid being semi private. Sometimes It's the best way to reference an object that hasn't been persisted to storage yet.

In which case you have a model instance, and you're free to pass it to `#get` to check if it's stored in the collection.

craigmichaelmartin:
Or you could pass in the `cid` property. 

My impression was that we should only assume `cid` is meaningful in an object if that object is a model, and this made `get` more judicious. However, if Adam want's to preserve {cid: 'something'} functionality, I could understand that, and we could do `obj.cid && _.byId[obj.cid]` and ditch the isModel check. Thoughts?

akre54:
> If the model isn't in there, we check for a cid regardless of #modelID's return value.

And who cares? There will only be an `"undefined"` entry in `_byId` if you have a bug in your code that gives one of your models the id `"undefined"`. Again, not a real problem.

> In which case you have a model instance, and you're free to pass it to #get to check if it's stored in the collection.

We're getting off topic but typically you may just have a reference to the string `cid` (otherwise you'd just use the JS reference to the model). We can discuss `cids` separately but given the current requirements of this method, this change should work, right?

akre54:
> we could do obj.cid && _.byId[obj.cid] and ditch the isModel check

We could do that.

jridgewell:
`cid in obj` would be good, too.

craigmichaelmartin:
Thanks! I ended up using `obj.cid` because `in` fails for non objects. I also added an ending `|| void 0` to account for if obj.cid is falsey that it returns undefined. If you don't mind potentially returning 0, false, '', etc, I can remove it. I'm guessing Adam isn't going to it :P Just let me know.

test/collection.js
```diff
@@ -106,6 +106,16 @@
     assert.equal(collection.get(1).id, 1);
   });
 
+  QUnit.test('get with "undefined" id first', function(assert) {
+    var collection = new Backbone.Collection([{id: 'undefined'}, {id: 1}]);
```
akre54: Let's leave this out, we shouldn't be officially supporting this.

backbone.js
```diff
@@ -980,11 +980,13 @@
...
-    // Get a model from the set by id.
+    // Get a model from the set by id, cid, model object with id or cid
+    // properties, or an attributes object that is transformed through modelId.
     get: function(obj) {
       if (obj == null) return void 0;
-      var id = this.modelId(this._isModel(obj) ? obj.attributes : obj);
-      return this._byId[obj] || this._byId[id] || this._byId[obj.cid];
+      return this._byId[obj] ||
+        this._byId[this.modelId(obj.attributes || obj)] ||
+        obj.cid && this._byId[obj.cid] || void 0;
```
akre54: I think this is fine without the final guard. A 0 cid might be a worthwhile case to support, but I'd argue you're almost always going to be getting those `cids` from *somewhere* (i.e. an existing model) which [can never have a cid of 0](https://github.com/jashkenas/underscore/blob/1.8.3/underscore.js#L1373-L1379).

craigmichaelmartin:
Sure thing. Removed the final guard ensuring `undefined` for no matches (and the test), and removed the test I add for an id with value "undefined" as well as the existing one. Let me know if there is any other changes.

akre54:
Great, just crossing fingers this doesn't cause some weird edge cases :)

backbone.js
```diff
@@ -980,11 +980,13 @@
...
 
-    // Get a model from the set by id.
+    // Get a model from the set by id, cid, model object with id or cid
+    // properties, or an attributes object that is transformed through modelId.
     get: function(obj) {
       if (obj == null) return void 0;
-      var id = this.modelId(this._isModel(obj) ? obj.attributes : obj);
-      return this._byId[obj] || this._byId[id] || this._byId[obj.cid];
+      return this._byId[obj] ||
+        this._byId[this.modelId(obj.attributes || obj)] ||
```
akre54: Note: this now disallows an object with `attributes` object to be passed in to `get`. We usually tell folks to stay away from sticking objects on the model attributes, but this a potential breaking change.

craigmichaelmartin: Yeah, do you want to allow this?

akre54: I think it should be fine, just noting the downside. Worst case if it's *really* bad we can add a patch release. My preference is to favor sane defaults over guarding against every possible edge case.

craigmichaelmartin: :+1:

PR #3940, opened 2016-01-28
Fixing object notation for Events.once

roycetucker:
Upon upgrading, I found that Events.once was no longer passing context through when using object notation.

```javascript
this.once({closed: this.onClosed, shown: this.onShown}, this);
```

Other event handling forms were still functioning properly:

```javascript
this.on({closed: this.onClosed, shown: this.onShown}, this);
this.once('closed', this.onClosed, this);
this.once('shown', this.onShown, this);
```

I couldn't find a changeset explaining why `undefined` was being passed through instead of the callback object, but doing so drops the context when two arguments were used.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -309,7 +309,7 @@
   Events.once = function(name, callback, context) {
     // Map the event into a `{event: once}` object.
     var events = eventsApi(onceMap, {}, name, callback, _.bind(this.off, this));
-    return this.on(events, void 0, context);
+    return this.on(events, callback, context);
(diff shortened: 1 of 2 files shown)


megawac:
Mind adding a test case to https://github.com/jashkenas/backbone/blob/master/test/events.js

jridgewell:
Yah, we'll need a test case. Thanks!

megawac:
Great catch, thanks @roycetucker

PR #3943, opened 2016-01-31
Minor. Change urlError to urlNeededError and make message more verbose.

sam-at-github:
I'm sure one of the first errors newbs will come across is the urlError Error. It was for me. This minor commit just makes this error a little bit more verbose and easier to understand for newbs like me.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -690,7 +690,7 @@
       var base =
         _.result(this, 'urlRoot') ||
         _.result(this.collection, 'url') ||
-        urlError();
+        urlNeededError();
       if (this.isNew()) return base;
       var id = this.get(this.idAttribute);
       return base.replace(/[^\/]$/, '$&/') + encodeURIComponent(id);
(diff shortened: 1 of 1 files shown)


jridgewell:
Ping: @akre54, what do you think?

akre54:
I don't see how this is any clearer? `A "url" property or function must be specified` vs `No url specified.`

They say the same thing?

sam-at-github:
> I don't see how this is any clearer? A "url" property or function must be specified vs No url specified.

Yeah but that is not an accurate summary of the change.. Newb perspective. It may be the same to you because your experienced with the software. When I first saw this message it sounded a bit cryptic.

akre54:
Sure, I'm happy to merge in something that makes it clearer. But that change is no clearer than the text it replaces.

sam-at-github:
Didn't want to be too specific otherwise you'd lose generality, but It's clearer ...

Sendoushi:
Greetings! 

I've recently found some  amazing information, you  might like and I think it is worth  reading! Just  take a look <http://ltysyndushi.mattbuysgold.com/e4lhif>


Rushing, Alberto Martinez

PR #3945, opened 2016-01-31
Use strictEqual in Collection#has Tests

jridgewell:
I'm getting failures when testing locally that `assert.notOk` isn't a function. :cry:

Testing the boolean return values is probably intended anyways.

### DIFF
--- a/test/vendor/qunit.css
+++ b/test/vendor/qunit.css
@@ -1,27 +1,27 @@
 /*!
- * QUnit 1.17.1
- * http://qunitjs.com/
+ * QUnit 1.21.0
+ * https://qunitjs.com/
  *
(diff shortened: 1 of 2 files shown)


jdalton:
`notOk` exists in at least QUnit v1.18.0.

http://api.qunitjs.com/notOk/.

jridgewell:
It's a fresh install... Let me try removing `node_modules` and reinstalling.

jridgewell:
The issue is we're serving a vendored QUnit when testing in a browser locally (the `test/index.html` page). Upgrading it to match our npm version fixes it.

jdalton:
Just use the npm one rewire the paths to `node_modules/qunit/qunit.js` or smth.

PR #3947, opened 2016-02-01
Minor update to docs. 

sam-at-github:
Qualify the behavior of fetch({reset: true}). Link doc on reset event to reset method.

P.S. I know index.html is served from the gh-pages branch, but could not find instruction on where to commit doc and looks like master is ahead of gh-pages. Also not sure whether HTML was being built from anything - doesn't look like it. Couldn't find instruction.


megawac:
No worries, this PR is correct. We cherry-pick commits to the gh-pages branch when we're ready

Sendoushi:
Dear,   This  it what you like, you just have to take a look! Info here <http://faspytrasla.creativecateringbydesign.org/e4kjcmac>

Warmest, Alberto Martinez

PR #3953, opened 2016-02-05
add tests for undefined id

craigmichaelmartin:
I would like to re-propose these test. These were removed from https://github.com/jashkenas/backbone/pull/3936 under the impression that there is no need to ensure support for unreasonable values. I believe backbone should be un-opinionated about what values are legitimate. Examples are contrived, until they're not. The second test in particular is important to ensure _byId doesn't evaluate for an `undefined` return.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -984,7 +984,7 @@
     // properties, or an attributes object that is transformed through modelId.
     get: function(obj) {
       if (obj == null) return void 0;
-      return this._byId[obj] ||
+      return (_.isString(obj) || _.isNumber(obj)) && this._byId[obj] ||
         this._byId[this.modelId(obj.attributes || obj)] ||
         obj.cid && this._byId[obj.cid];
     },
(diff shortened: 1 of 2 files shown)


craigmichaelmartin:
Added a guard to only treat `obj` as an id if it is a string or number.

akre54:
> I believe backbone should be un-opinionated about what values are legitimate. Examples are contrived, until they're not.

`get` has a contract. You may call that opinionated, but we've said [again and again](https://github.com/jashkenas/backbone/search?q=arbitrary&type=Issues&utf8=%E2%9C%93) that developers shouldn't be using arbitrary values as ids. Particularly for a hot code path like `get`, it'd be best to keep the logic as un-cluttered as possible.

craigmichaelmartin:
`get` has a contract that accepts a wide range of values. I am a fan of ensuring it is prudent about the path it takes given the specific value that is passed in, and without opinion following that path to the end.

That said, I understand where you are coming from. Thanks!

Issue #3941, opened 2016-01-29
Router.navigate executes although trigger:false with encoded hash fragment in Firefox


bobmop:
`Router.navigate()` will execute with `trigger:false` if the has fragment will contain url encoded characters in FireFox (tested with V39, V43).

correct behaviour in Backbone 1.1.2: https://jsfiddle.net/g9zmwmdk/10/
wrong behaviour in Backbone 1.2.3: https://jsfiddle.net/g9zmwmdk/9/

I thinkl the new behaviour was introduced in https://github.com/jashkenas/backbone/compare/1.1.2...1.2.0#diff-0d56d0d310de7ff18b3cef9c2f8f75dcR1764

akre54:
Yup, specifically in #3434 as a fix for #3426. I'm not super familiar with this code but I think the solution is to not to pre-encode your string. See if that works for you?

bobmop:
Thank you for the feedback.

I get a path from an api which is already encoded and navigate() using this path without extra encoding.

I think i found the "broken" code:

After `navigate()` is called correctly, `Backbone.History.checkUrl()` is called by an event listener and will compare the cached (decoded) fragment with `getFragment()`, which is not decoded by Backbone itself. The result is another fragment than the cached one and `loadUrl()` including the configured handler is executed.

I think Backbone has to be fixed, e.g. calling this.decodeFragment() in this.getFragment().

akre54:
If you want to open a pull with your proposed solution and some tests that fail on master we'd be happy to take a look.

bobmop:
here is a proposal including a test: #3955

PR #3955, opened 2016-02-05
#3941 return decoded hash in getHash()

bobmop:
...to compare correctly to cached fragment in navigate()

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1630,7 +1630,7 @@
     // in Firefox where location.hash will always be decoded.
     getHash: function(window) {
       var match = (window || this).location.href.match(/#(.*)$/);
-      return match ? match[1] : '';
+      return match ? this.decodeFragment(match[1]) : '';
     },
 
     // Get the pathname and search params, without the root.
--- a/test/router.js
+++ b/test/router.js
@@ -407,6 +407,17 @@
     assert.strictEqual(router.rest, 'has space');
   });
 
+  QUnit.test('#3941 - should return the decoded hash in getFragment().', function(assert) {
+    assert.expect(2);
+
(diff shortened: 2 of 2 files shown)


akre54:
@braddunbar ?

braddunbar:
This seems fine to me. I'd say it could even be moved up into `getFragment` since `getPath` calls `decodeFragment` too.

bobmop:
The methods `getPath`and `getHash` are also called separately. I can't figure out the influence...

megawac:
I'm fine with this but it may be a minor breaking change. I'd consider it a bug fix personally though

barceyken:
We added this to our project too :+1:

hbomb9000:
This seems to break browsers 'back' navigation button. However, it does fix the above stated problem.

jashkenas:
> This seems to break browsers 'back' navigation button. However, it does fix the above stated problem.

Uh.

platinumazure:
It'd be great to have normalized hash decoding behavior, but the bit about breaking a browser's "back" button is highly concerning. Does anyone know why that happens? Is it fixable, or is this the wrong direction?

Daniel-Alonso-Exabeam:
This will not work if `match[1]` contains a "%" character that is not used in an encoded character sequence, because the native `decodeURI` function will throw an error. This can happen in two cases:
1. If the encoded fragment passed into `Backbone.history.navigate` contains a "%25". 
2. In Firefox, `this.location.href` returns the url in an encoded format, but "%" characters are not encoded for some reason (there are no "%25").

jgonggrijp:
Rejecting this as it is a breaking change *and* it seems to cause worse problems than it aims to solve.

PR #3958, opened 2016-02-06
Update README to include high-level, easy-to-comprehend introduction

adhir-lalla:
I have written a clear, easy-to-comprehend introduction for Backbone.js. I believe this introduction, free of complex technical terms, will swiftly and thoroughly inform anyone about the use-cases, merit, and advantages, including the prominent features, of Backbone.js. Seasoned developers, new programmers, and even nontechnical visitors will benefit from the clear language, exposing Backbone.js to a broader range of users.

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,10 @@
 
 Backbone supplies structure to JavaScript-heavy applications by providing models with key-value binding and custom events, collections with a rich API of enumerable functions, views with declarative event handling, and connects it all to your existing application over a RESTful JSON interface.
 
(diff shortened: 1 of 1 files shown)


akre54:
I like it but it's a little long (and I think index.html is the better place for this kind of wording).

We already have the [Getting Started](http://backbonejs.org/#Getting-started) section on the website. Is that not clear enough?

adhir-lalla:
Thanks for the feedback. Indeed the "Getting Started" section adds clarity

Issue #3961, opened 2016-02-08
Collection#set - add events sends options.index that correspond to wrong index


jrdb:
As pointed out by @akre54 in issue #3960, `options.index` was deprecated for `'add'` events in 0.9.9.

However, under certain cases `'add'` events generated from a call to **Collection#set** passes a stale `options.index` that was assigned during the generation of `'remove'` events in **Collection# _removeModels**.  

Since the options argument of event handlers are not well documented, you should be able to rely on anything passed, including options.index.

The intent was that `options.index` would only be sent if `at` was specified.  However `options.index` is not unset if `at` is unspecified.  If `_removeModels` has set `options.index`, and `at` is unset, then `'add'` events will get a stale `options.index` (from _removeModels).

Since the same `options` object is reused across all events (add and remove),  `_removeModels` needs to cleanup after itself after all `'remove'` events are triggered.

`delete options.index;` 

---

Bug Preconditions:
-   options.remove = true 
-   options.at == null
-   passed in attrs has elements to add and collection has elements to remove

Bug Manifestation:
- _removeModels will set options.index when sending out its triggers.
- If options.at == null, the set method does not clear out the options.index set by _removeModels before triggering the add events. In that case, the last options.index from the remove events is reused in the add events.

`

```
    var array = _.map([1,2,3,4], function(val) {
        return {property:val};
    });

    var collection = new Backbone.Collection(array);

    collection.on('add', function(model, collection, options){
        // BUG!!: options.index (0) != collection.indexOf(model) (3)
        console.error(
            'BUG!!: options.index ('
            + options.index +
            ') != collection.indexOf(model) ('
            + collection.indexOf(model) +
            ')'
        );
    });

    array = collection.toArray();
    array.shift();
    array.push({property:5});
    collection.set(array);
```

`

PR #3964, opened 2016-02-08
#3961 - add events sends options.index that correspond
to wrong index

jrdb:
fix and test for #3961 - add events sends options.index that correspond
to wrong index

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1140,6 +1140,7 @@
         removed.push(model);
         this._removeReference(model, options);
       }
+      if(models.length > 0 && !options.silent) delete options.index;
       return removed;
     },
 
--- a/backbone.js
+++ b/backbone.js
@@ -1140,7 +1140,7 @@
         removed.push(model);
         this._removeReference(model, options);
       }
-      if(models.length > 0 && !options.silent) delete options.index;
+      if (models.length > 0 && !options.silent) delete options.index;
       return removed;
     },
 
--- a/test/collection.js
+++ b/test/collection.js
@@ -1996,21 +1996,21 @@
   });
 
 
-  QUnit.test("#3961 - aadd events sends options.index that correspond to wrong index", function(assert) {
-  	var numModels = 4;
-  	var models = _.each(['a','b','c','d'], function(val) {
-  		return new Backbone.Model({id:val});
-  	});
+  QUnit.test('#3961 - add events sends options.index that correspond to wrong index', function(assert) {
+    var numModels = 4;
+    var models = _.each(['a', 'b', 'c', 'd'], function(val) {
+      return new Backbone.Model({id:val});
+    });
 
-	var collection = new Backbone.Collection(models);
+    var collection = new Backbone.Collection(models);
 
-	models.shift(); // remove first element;
-	models.push(new Backbone.Model({id:'e'}));
+    models.shift(); // remove first element;
+    models.push(new Backbone.Model({id:'e'}));
 
-	collection.on('add', function(model, coll, options){
-		assert.equal(options.index, undefined);
-	});
-	collection.set(models);
+    collection.on('add', function(model, coll, options){
+      assert.equal(options.index, undefined);
+    });
+    collection.set(models);
   });
 
 })();
--- a/test/collection.js
+++ b/test/collection.js
@@ -1995,4 +1995,22 @@
     assert.equal(fired, false);
   });
 
+
+  QUnit.test("#3961 - aadd events sends options.index that correspond to wrong index", function(assert) {
+  	var numModels = 4;
+  	var models = _.each(['a','b','c','d'], function(val) {
+  		return new Backbone.Model({id:val});
+  	});
+
+	var collection = new Backbone.Collection(models);
+
+	models.shift(); // remove first element;
+	models.push(new Backbone.Model({id:'e'}));
+
+	collection.on('add', function(model, coll, options){
+		assert.equal(options.index, undefined);
+	});
+	collection.set(models);
+  });
+
 })();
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -1140,6 +1140,7 @@
         removed.push(model);
         this._removeReference(model, options);
       }
+      if (models.length > 0 && !options.silent) delete options.index;
```
akre54: Why the `options.silent` check? If it's just to check whether `add` will fire, shouldn't we always remove it?

Also, you don't have to delete this prop (`delete` is slow). Just set it to `undefined`.

jrdb: > Why the options.silent check? If it's just to check whether add will fire, shouldn't we always remove it?

If options.silent == true, then options.index was never set. The check saves the expensive delete operation.

> Also, you don't have to delete this prop (delete is slow). Just set it to undefined.

Since you are passing options to an event callback, you don't know how the callback will iterate over the options object.  If you set it to `void 0` and the event callback does a `for(key in options)`, then `options.index` will still show up, `options.hasOwnProperty('index') == true`.  

The event callbacks could then get any of the following cases:
1. `options.index = value`
2. `options.index = undefined`
3. `options` does not own property `index`

I'm not dogmatic about it, I'll make the change if you are ok sacrificing consistency on the last two cases for performance. I think you can make a reasonable argument that the callbacks should be able to handle all 3 cases.  Then it would make sense to remove the options.silent check.

jrdb: Could also add `index:void 0` as part of the `setOptions` which would eliminate case 3 and then a `options.index = void 0;` in `_removeModels` would result in consistent behavior.

akre54: Why would you be iterating the options object looking for `options.index`?

jrdb:
If you expose something as part of an API, you shouldn't constrain how people look at what was exposed regardless of hypothetical use-cases.  I think most people would choose consistency over inconsistency if consistency comes almost for free.

In terms of cost:

N = num elements to delete
The performance hit to `_removeModels` of **1 more delete** in comparison to the existing **N splices** in `_removeModels` + **2N deletes** in `_removeReference` is insignificant.

All 3 of the solutions below fix the bug that the ***WRONG*** `options.index `is being sent to the 'add' event.  I don't have a strong opinion about which solution to implement, I went with the most conservative because it seemed like it would have the least downstream side-effects and the cost really is minimal.  The `delete` happens once per collection `set` IFF `options.remove:true` AND `options.silent:false` AND the collection has elements to remove. 

**Solutions**
* `delete options.index` - **consistent**  _cases 1,3_ This is the truest representation of the documentation/changelog "no longer set", and has the least side-effects to options.  You are more likely to leave options in the state that it was in when it was passed into `set`.  If we are trying to minimize side-effects, then we should add a `delete options.index` after it is used by the 'add' callbacks as well.
* `options.index = void *;`  and `setOptions` has a default `index:void 0` option -  **consistent** _cases 1,2_
* `options.index = void *;` - **inconsistent** _cases 1,2,3_

backbone.js
jrdb: If you expose something as part of an API, you shouldn't constrain how people look at what was exposed regardless of hypothetical use-cases.  I think most people would choose consistency over inconsistency if consistency comes almost for free.

In terms of cost:

N = num elements to delete
The performance hit to `_removeModels` of **1 more delete** in comparison to the existing **N splices** in `_removeModels` + **2N deletes** in `_removeReference` is insignificant.

All 3 of the solutions below fix the bug that the ***WRONG*** `options.index `is being sent to the 'add' event.  I don't have a strong opinion about which solution to implement, I went with the most conservative because it seemed like it would have the least downstream side-effects and the cost really is minimal.  The `delete` happens once per collection `set` IFF `options.remove:true` AND `options.silent:false` AND the collection has elements to remove. 

**Solutions**
* `delete options.index` - **consistent**  _cases 1,3_ This is the truest representation of the documentation/changelog "`options.index` will no longer be set ", and has the least side-effects to options.  You are more likely to leave options in the state that it was in when it was passed into `set`.  If we are trying to minimize side-effects, then we should add a `delete options.index` after it is used by the 'add' callbacks as well.
* `options.index = void *;`  and `setOptions` has a default `index:void 0` option -  **consistent** _cases 1,2_
* `options.index = void *;` - **inconsistent** _cases 1,2,3_

jridgewell: Let's just delete the prop to make it easy. But why are we guarding with `this.models > 0`?

jrdb: It's not guarding on `this.models`, it's guarding on the passed in `arguments[0] : models` (to be removed).  It's an optimization.  If there were no models to delete, then `options.index` was not set in _removeModels.  Therefore, no need to delete `options.index`.

jrdb: @akre54 - Coincidentally, came across a use case on a chained action.  A copy of the `options` with defaults for unset properties was created in the event listener.  The `options` copy was used to modify a related collection with different `set` behavior.

`options = _.extend({index:0}, options);`


The default, won't take when `options.index = void 0`;

jgonggrijp:
A bit late, but thanks for your contribution, @jrdb!

Please run these commands to ensure that your copy of `master` stays in sync with the central one:

```sh
# just once
git fetch jashkenas # assuming you made jashkenas/backbone the "jashkenas" remote
git checkout master
git reset --hard jashkenas/master
git config remote.pushDefault origin # assuming you made your own fork on GitHub the "origin" remote
git config branch.master.remote jashkenas
# the latter two commands ensure that you pull master from jashkenas but push it to origin

# every time you start working on a new contribution
git checkout master
git pull
git checkout -b feature-name # work on a separate branch
```

Issue #3965, opened 2016-02-09
Collections' modelId method should be passed the Model as a second argument


amiller-gh:
Right now `modelId(attrs)` is only passed a plain-object representation of the model's attributes. This makes it difficult for application developers to create n-degree polymorphic Collection subclasses.

Contrived Example:

``` javascript
// A collection of mixed entity types that maintains sort order based 
// on their joinDate property

var Entities = Collection.extend({
  modelId(attrs){
    // What ID attribute to use? Not determined at runtime.
  },
  comparator: 'joinDate'
});


// These are only two entity types with different idAttributes, but assume 
// There may be up to (n) different entity types, each with different idAttributes, 
// determined by the database backing this example application.

var Company = Model.extend({
  idAttribute: 'companyId'
});

var Customer = Model.extend({
  idAttribute: 'customerId'
});

// Proceeding to add and remove Customers, Companies, etc, to the Entities
// Collection will mess up the `_byId` cache because the Entities collection
// is not aware of all the different possibilities for `idAttribute` at runtime and 
// a `modelId` method cannot be written to satisfy the needs of this app.

```

If the modelId callback took the form `modelId(attrs, model)`, where the second argument is the model in question, Backbone can retain backwards compatibility but allow application developers to do something like this:

``` javascript
// Always give precedence to the provided model's idAttribute. 
// Fall back to the Collection's idAttribute, and then to the default value `id`.

Entities.modelId = function modelId(attrs, model){
  return attrs[model.idAttribute || this.model.prototype.idAttribute || 'id'];
};
```

It may even make sense to make the above `modelId` function the default implementation and support polymorphic Collections out of the box. It still maintains backwards compatibility: For any implementation that is not already polymorphic, Collections will still function normally – models will have their idAttribute set when added to the Collection by the Collection's model constructor. Those that are already polymorphic, or have some type of non-standard idAttributes, will already have this modelId method overwritten.

Happy to throw up a PR if you think this is a good idea.

jridgewell:
Please! But keep in mind, we've tried doing this [before](https://github.com/jashkenas/backbone/issues/3408#issuecomment-66223271) without much success.

epicmiller:
PR is here: https://github.com/jashkenas/backbone/pull/3966

prantlf:
[...]

Issue #2043, opened 2012-12-28
Backbone polymorphism does not support two models with same ID


gabetax:
Following the polymorphic collection instructions in #793, there seems to be an issue if you have two different models types (saving to two different URLs, and ultimately two different database tables) with the same `id` 

``` js
collection = new App.Collections.PaymentsCollection(
  {"id":6,"amount":"2000.0","class_name":"CreditCard"},
  {"id":6,"amount":"1000.0","class_name":"ECheck"}
)
collection.models.length
>>> 1
collection.models
>>> [CreditCard]
```

I can understand why a collection would want to remove duplicate models of identical objects, but I think it's a reasonable use case to have a polymorphic collection that may have models of different classes but with the same id.

The main sticking point seems to be in [`collection.get`](https://github.com/documentcloud/backbone/blob/master/backbone.js#L709), which [`collection.add`](https://github.com/documentcloud/backbone/blob/master/backbone.js#L605) uses in its duplication check.

Do you have any opinions on the best way to handle this polymorphism use case? Up until this one issue, the solution from #793 has felt easy and elegant to me.

PR #3966, opened 2016-02-09
Update `modelId()` to Allow Polymorphic Collections with Different, Unknown, Model.idAttributes

amiller-gh:
For Issue: https://github.com/jashkenas/backbone/issues/3965
## Changes
- `modelId()` now takes the idAttribute of the model in question as the second argument
- `modelId()` will use the passed `idAttribute` before falling back to `Collection.prototype.model.idAttribute` and then `'id'`
- Retains backwards compatibility
- Update tests to show default `modelId()` fallbacks
- Update tests to show default polymorphic collection behavior
## Rationale

Right now `modelId(attrs)` is only passed a plain-object representation of the model's attributes. This makes it difficult for application developers to create n-degree polymorphic Collection subclasses.

Contrived Example:

``` javascript
// A collection of mixed entity types that maintains sort order based 
// on their joinDate property

var Entities = Collection.extend({
  modelId(attrs){
    // What ID attribute to use? Not determined at runtime.
  },
  comparator: 'joinDate'
});


// These are only two entity types with different idAttributes, but assume 
// There may be up to (n) different entity types, each with different idAttributes, 
// determined by the database backing this example application.

var Company = Model.extend({
  idAttribute: 'companyId'
});

var Customer = Model.extend({
  idAttribute: 'customerId'
});

// Proceeding to add and remove Customers, Companies, etc, to the Entities
// Collection will mess up the `_byId` cache because the Entities collection
// is not aware of all the different possibilities for `idAttribute` at runtime and 
// a `modelId` method cannot be written to satisfy the needs of this app.

```

By making the modelId callback take the form `modelId(attrs, model)`, where the second argument is the model in question, Backbone can retain backwards compatibility but enable the default implementation to be:

``` javascript
// Always give precedence to the provided model's idAttribute. 
// Fall back to the Collection's idAttribute, and then to the default value `id`.

Entities.modelId = function modelId(attrs, idAttribute){
  return attrs[(idAttribute || this.model.prototype.idAttribute || 'id'];
};
```

It still maintains backwards compatibility: For any implementation that is not already polymorphic, Collections will still function normally – models will have their idAttribute set when added to the Collection by the Collection's model constructor. Those that are already polymorphic, or have some type of non-standard idAttributes, will already have this modelId method overwritten.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -984,7 +984,7 @@
     get: function(obj) {
       if (obj == null) return void 0;
       return this._byId[obj] ||
-        this._byId[this.modelId(obj.attributes || obj)] ||
+        this._byId[this.modelId(obj.attributes || obj, obj)] ||
         obj.cid && this._byId[obj.cid];
     },
 
@@ -1088,8 +1088,8 @@
     },
 
     // Define how to uniquely identify models in the collection.
-    modelId: function(attrs) {
-      return attrs[this.model.prototype.idAttribute || 'id'];
+    modelId: function(attrs, model) {
+      return attrs[(model && model.idAttribute) || this.model.prototype.idAttribute || 'id'];
     },
 
     // Private method to reset all internal state. Called when the collection
@@ -1129,7 +1129,7 @@
         // Remove references before triggering 'remove' event to prevent an
         // infinite loop. #3693
         delete this._byId[model.cid];
-        var id = this.modelId(model.attributes);
+        var id = this.modelId(model.attributes, model);
         if (id != null) delete this._byId[id];
 
         if (!options.silent) {
@@ -1152,7 +1152,7 @@
     // Internal method to create a model's ties to a collection.
     _addReference: function(model, options) {
       this._byId[model.cid] = model;
-      var id = this.modelId(model.attributes);
+      var id = this.modelId(model.attributes, model);
       if (id != null) this._byId[id] = model;
       model.on('all', this._onModelEvent, this);
     },
@@ -1160,7 +1160,7 @@
     // Internal method to sever a model's ties to a collection.
     _removeReference: function(model, options) {
       delete this._byId[model.cid];
-      var id = this.modelId(model.attributes);
+      var id = this.modelId(model.attributes, model);
       if (id != null) delete this._byId[id];
       if (this === model.collection) delete model.collection;
       model.off('all', this._onModelEvent, this);
@@ -1175,8 +1175,8 @@
         if ((event === 'add' || event === 'remove') && collection !== this) return;
         if (event === 'destroy') this.remove(model, options);
         if (event === 'change') {
-          var prevId = this.modelId(model.previousAttributes());
-          var id = this.modelId(model.attributes);
+          var prevId = this.modelId(model.previousAttributes(), model);
+          var id = this.modelId(model.attributes, model);
           if (prevId !== id) {
             if (prevId != null) delete this._byId[prevId];
             if (id != null) this._byId[id] = model;
--- a/backbone.js
+++ b/backbone.js
@@ -984,7 +984,7 @@
     get: function(obj) {
       if (obj == null) return void 0;
       return this._byId[obj] ||
-        this._byId[this.modelId(obj.attributes || obj, obj)] ||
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -1088,8 +1088,8 @@
     },
 
     // Define how to uniquely identify models in the collection.
-    modelId: function(attrs) {
-      return attrs[this.model.prototype.idAttribute || 'id'];
+    modelId: function(attrs, model) {
+      return attrs[(model && model.idAttribute) || this.model.prototype.idAttribute || 'id'];
```
craigmichaelmartin: Could this be shortened to just `attrs[(model && model.idAttribute) || 'id'];`?

craigmichaelmartin: Nevermind, just saw the note about maintaining backward compatibility, sorry :)

craigmichaelmartin:
:+1: If liked, would it be fitting to remove the [couple](http://backbonejs.org/#Collection-modelId) [warnings](https://github.com/jashkenas/backbone/blob/master/test/collection.js#L1717-L1733) about collections with polymorphic models using `idAttribute` needing to override modelId?

backbone.js
jrdb: Instead of passing the entire model, you could pass the idAttribute of the model.

Then you could more easily implement a get that would work with your example.


```
collection.get({companyId:1234}, 'companyId');
collection.get({customerId:1234}, 'customerId');
```

```
 get: function(obj, idAttribute) {
      if (obj == null) return void 0;
      var id = this.modelId(this._isModel(obj) ? obj.attributes : obj, idAttribute);
      return this._byId[obj] || this._byId[id] || this._byId[obj.cid];
    },
```

```
modelId: function (attrs, idAttribute) {
    return attrs[idAttribute || this.model.prototype.idAttribute || 'id'];
},
```

epicmiller: I found a major hole in this PR because of this comment, thank you! 

The _byId cache stores models in the collection by both `cid` and `id`, so by modifying `modelId` to return the id of polymorphic elements, we run the risk of cache conflicts. Ex: A mixed entities collection containing both companies and users where there is a company with `companyId: 1` and a user with `userId: 1`. This could cause issues.

To fix this, the _byId cache would need to track models internally by the unique combination of idAttribute and id:  `_byId[model.idAttribute + this.modelId(model.attributes, model.idAttribute)] = model`.

However, then `Collection.get` will no longer be able to find models by `id` at all any more.

We'd need to tweak `Collection.get` to do something like this:
```
    // Get a model from the set by cid, id, model object with id or cid
    // properties, or an attributes object that is transformed through modelId.
    // When fetching by id, use the optional idAttribute if provided, otherwise the 
    // provided model's idAttribute, or the collection's template model's idAttribute
     get: function(obj, idAttribute) {
      if (obj == null) return void 0;
      idAttribute = idAttribute || obj.idAttribute  || this.model.prototype.idAttribute;
      return this._byId[obj] ||
        this._byId(idAttribute + obj) ||
        this._byId[idAttribute + this.modelId(obj.attributes || obj, idAttribute)] ||
        obj.cid && this._byId[obj.cid];
    },
```
This just got a little more complicated, but what do you think? Messing with how models are stored in `_byId` is a major change...

epicmiller: Although, that conflicting id concern is assuming polymorphic models can have conflicting ids, an assumption that Backbone already seems to ignore: http://backbonejs.org/#Collection-model

epicmiller: Yeah, Backbone as it is assumes polymorphic collections won't contain models of different types with the same id: http://jsfiddle.net/53oo2bo9/ 

That doesn't feel intuitive to me, but as it is now, this PR will still work fine :)

I like the idea of just passing the idAttribute, will update the PR

epicmiller:
Another benefit of this change: Allows collections using component generators to use the appropriate idAttribute of its possible models.

http://jsfiddle.net/53oo2bo9/4/

epicmiller:
@caseywebdev, pulling you in because of your original work on `modelId`: https://github.com/jashkenas/backbone/pull/3133

Also, wondering if it may be time to talk about this again in a separate thread: https://github.com/jashkenas/backbone/issues/2043 (see minor freak out on above line comment for context)

PR #3967, opened 2016-02-10
Refactored Router to make routeToRegExp a static method

beatfactor:
I'd like to be able to use `routeToRegExp` in node.js without having to create an instance of the Backbone.Route in order to be able to share the route definitions with the Express router (exported as regex).

This PR is mostly just refactoring to support this.


akre54:
Pro tip: the method is exported as `Router.prototype.routeToRegExp`. No need to create a new instance.

It may at one point have made sense to put it on the class instead of the instance, but for backwards-compatibility reasons we shouldn't move this around.

Issue #882, opened 2012-01-22
Take another run through with Ember.js


jashkenas:
Before we do a 0.9.0 release, it's been a while since the last time we took a dive into what SproutCore/Amber/Ember/Data.js was up to, and if there's anything worth borrowing. Let's be sure to take another run through ... and if you have a personal favorite feature which you think would benefit Backbone, this thread would be the ideal place to mention it.

PR #3968, opened 2016-02-10
Computed properties

milkshakeuk:
Hi I am opening this pull request as I think it could be very useful to include the ability to have computed properties on models in backbone.

These could/should be accessed by `get('field')` and the result of the function would be returned;
The computed result would also be serialised into the json object when `toJSON` gets invoked.
When `toJSON` is invoked as a result of calling `save()` on the model these computed properties should not be included in the serialised json object that gets sent to the server.

this would be incredibly helpful in conjunction with marionette views and or when using a template system such as handlebars.

here is an example of how this might get used:

```javascript
var Model = Backbone.Model.extend({
  computed: {
    fullName: function () {
      return this.get('firstName') + ' ' + this.get('lastName');
    }
  }
});
var model = new Model({
  firstName: 'Tim',
  lastName: 'Burton'
});

model.get('fullName'); // returns 'Tim Burton'
````
and when serialising...
```javascript
model.toJSON();
/**
 *  returns
 *  {
 *    firstName: 'Tim',
 *    lastName: 'Burton',
 *    fullName: 'Tim Burton'
 *  }
 */
```
but when saving the model data would serialise to...
```javascript
model.save();
/**
 *  serialised:
 *  {
 *    firstName: 'Tim',
 *    lastName: 'Burton',
 *  }
 */
```

let me know what you think, I have included tests in the pull request.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -427,8 +427,23 @@
     initialize: function(){},
 
     // Return a copy of the model's `attributes` object.
+    // When saving we dont want to pollute the attributes with computed
+    // properties but when serialising for use with the templates we do
     toJSON: function(options) {
-      return _.clone(this.attributes);
+      var isSaving = _.has(options, 'emulateHTTP');
+      if (isSaving) return _.clone(this.attributes);
+
+      var hasComputed = !!this.computed;
+      var computed    = {};
+
+      if (hasComputed) {
+        _.each(this.computed, function(value, key) {
+          computed[key] = _.isFunction(value) ? value.call(this) : value;
+        }, this);
+      }
+
+      // ensure non-computed property wins in property clash
+      return _.extend(computed, this.attributes);
     },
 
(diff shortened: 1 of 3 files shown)


akre54:
Duplicate of https://github.com/jashkenas/backbone/issues/882#issuecomment-3638000. This kind of thing is much better handled and with less magic with a method.

milkshakeuk:
I have read the referenced comment... 

The problems I see with having it in a method is that it's never serialised into what gets passed to the template (e.g. Marionette) and if you have many views which all take the same model and use the same computed property,  then that's a lot of boiler plater you have to duplicate in each view which might want to use that computed property.

akre54:
You can always override `toJSON` in your base model or individual model to add this behavior. It's not something we should be adding to Backbone core, though.

milkshakeuk:
This is true, thanks for looking at the pull request.

PR #3973, opened 2016-03-03
Remove unused name parameter from Router.execute

bryanjenningz:
The name parameter isn't ever used. Removing it reduces confusion.


megawac:
`name` can be used by `Router.execute` overrides to provide custom functionality based on route (although we have no unit test coverage for it right now). I've seen it used in some Router extensions

See http://backbonejs.org/#Router-execute

akre54:
Should we add a comment because this keeps popping up? We also should put in a test for this to catch anyone ahead of time.

PR #3974, opened 2016-03-04
Test for 3 parameters in Router.prototype.execute

bryanjenningz:
This test saves time by automatically failing pull requests that remove the third parameter (the "name" parameter) from Router.prototype.execute. This test also serves as documentation for why Router.prototype.execute has a third parameter, despite the third parameter not being used by default.

See https://github.com/jashkenas/backbone/pull/3973

### DIFF
--- a/test/router.js
+++ b/test/router.js
@@ -1058,5 +1058,14 @@
     var myRouter = new MyRouter;
     Backbone.history.start({root: '®ooτ', pushState: true});
   });
-
+  
+  QUnit.test('Router.prototype.execute function definition should have 3 parameters (callback, args, name)', function(assert) {
+    // The name parameter isn't used in the default Router.prototype.execute function, but having the name parameter
+    // in the Router.prototype.execute function is useful because it serves as documentation for people who want to 
+    // extend Router.prototype.execute to use the name parameter.
+    assert.expect(1);
+    assert.ok(/function\s*\(\s*callback\s*,\s*args\s*,\s*name\s*\)/.test(Backbone.Router.prototype.execute.toString()),
+      'Router.prototype.execute should have 3 parameters: callback, args, name');
+  });
+  
 })();
--- a/test/router.js
+++ b/test/router.js
@@ -1058,7 +1058,7 @@
     var myRouter = new MyRouter;
     Backbone.history.start({root: '®ooτ', pushState: true});
   });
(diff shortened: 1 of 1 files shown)


test/router.js
```diff
@@ -1059,4 +1059,12 @@
     Backbone.history.start({root: '®ooτ', pushState: true});
   });
 
+  QUnit.test('Router.prototype.execute function definition should have 3 parameters (callback, args, name)', function(assert) {
+    // The name parameter isn't used in the default Router.prototype.execute function, but having the name parameter
+    // in the Router.prototype.execute function is useful because it serves as documentation for people who want to
+    // extend Router.prototype.execute to use the name parameter.
+    assert.expect(1);
+    assert.ok(/function\s*\(\s*callback\s*,\s*args\s*,\s*name\s*\)/.test(Backbone.Router.prototype.execute.toString()),
```
akre54: How about testing that `name` gets passed? This is pretty hacky.

megawac:
I'll be for this if its implemented as a test that `name` is passed after a route change

bryanjenningz:
I'm not sure how you would test that `name` gets passed into `Backbone.prototype.execute` since `Backbone.prototype.execute` doesn't use the `name` argument.
This is where `execute` is defined: https://github.com/jashkenas/backbone/blob/master/backbone.js#L1516

megawac:
@bryanjenningz you can just hijack this suite https://github.com/jashkenas/backbone/blob/82d8ec840c88ef5740b3ff6a8a0757a3c15da733/test/router.js#L365 or copy paste and make a new suite with something testing name

bryanjenningz:
I found a test that already checks to see if a non-default `execute` function gets passed the `name` argument, but it doesn't test if the default `execute` function has a `name` parameter. I guess I'm not sure how to test if the default `execute` function has a `name` parameter other than by using a regex to check if it's in the default `execute` function.

See https://github.com/jashkenas/backbone/blob/82d8ec840c88ef5740b3ff6a8a0757a3c15da733/test/router.js#L917

megawac:
The existing test is sufficient, thanks though @bryanjenningz.

We leave unused params on several functions just to make it obvious for anyone extending core functionality that they will receive that extra data. I don't think we need to unit test it though, just our convention.

PR #3977, opened 2016-03-05
Removed unused parameter options in toJSON method

hugohenrique:


akre54:
This is left in place to show users that the method will be called with 'options', we don't actually use it anywhere. (See #2201, #1535, #1616, #3228)

Issue #3975, opened 2016-03-04
Backbone 1.3.2


jridgewell:
I'm making a mess of this release process.

Checklist for v1.3.2 (which is really v1.3.0):

- Update `Backbone.VERSION`
- Update version comment at the top of `backbone.js`
- `npm run build`
- Bump `package.json`
- Create changelog entries telling everyone I screwed up and these versions are really the same
- Update `gh-pages` branch
- Push to npm after someone confirms I'm not an idiot.

Anyone know of anything else I missed? @paulfalgout, @megawac, @akre54?

akre54:
🙈

Minify / source map main backbone.js file, rebuild docco / docs

(Also the changelog release date on 1.3.0 is out of date, might wanna fix
that)

paulfalgout:
There's also a comment at the top of the library containing the version

jridgewell:
> Also the changelog release date on 1.3.0 is out of date, might wanna fix that

Yah, I'm just going to update the changelog to say `1.3.2` instead, so I'll update the date too.

> There's also a comment at the top of the library containing the version

Added.

paulfalgout:
Nice!  Well now that you'll have done 3 releases you'll be an expert.

jridgewell:
:stuck_out_tongue_winking_eye:

megawac:
Also create git tag, either by `npm version patch` or `git tag`

JSteunou:
FYI `npm version` + `npm scripts` hooks (pre|post version, pre|post publish) could help you a lot with that.

jridgewell:
> FYI npm version + npm scripts hooks (pre|post version, pre|post publish) could help you a lot with that.

Yup, I'm going to work on that after I do this one.

elgubenis:
the feature release broke something in our development environment, we are using marionette though and that might be an issue, seems like the backbone collections are not correctly triggering events to the collectionveiw after fetching, but need to investigate more to be sure

paulfalgout:
Marionette has a fixed version dependency on backbone. Since Bb isn't using semver, it requires we manually update Marionette with every release.  So if you've updated your Bb dependency locally, and you're using a package manager, you're actually using a different copy of Backbone than Marionette is.

elgubenis:
we are omitting the fixed version for marionette, in our environment marionette always uses the latest backbone version, that might be the issue

megawac:
[...]

PR #3980, opened 2016-03-07
Fix 1.3.1 regression with once and no context

megawac:
ping #3975

/cc @jridgewell  @elgubenis

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -308,6 +308,7 @@
   Events.once = function(name, callback, context) {
     // Map the event into a `{event: once}` object.
     var events = eventsApi(onceMap, {}, name, callback, _.bind(this.off, this));
+    if (typeof name === 'string' && context == null) callback = void 0;
     return this.on(events, callback, context);
   };
 
--- a/test/events.js
+++ b/test/events.js
@@ -417,6 +417,18 @@
     assert.equal(obj.counterB, 1, 'counterB should have only been incremented once.');
   });
 
+  QUnit.test('bind a callback with a default context when none supplied', function(assert) {
+    assert.expect(1);
+    var obj = _.extend({
+      assertTrue: function() {
+        assert.equal(this, obj, '`this` was bound to the callback');
+      }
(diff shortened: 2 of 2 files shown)


backbone.js
```diff
@@ -308,6 +308,7 @@
   Events.once = function(name, callback, context) {
     // Map the event into a `{event: once}` object.
     var events = eventsApi(onceMap, {}, name, callback, _.bind(this.off, this));
+    if (arguments.length === 2 && typeof name === 'string') callback = void 0;
```
akre54: Is there any reason we couldn't use `!context`? Doesn't checking arguments length cause a de-opt?

jridgewell: > Doesn't checking arguments length cause a de-opt?

It will (in sloppy mode (Backbone)), because we're changing a named parameter.

megawac: Good point, forgot that it causes a deopt if you modify a param. I'll change it to `context != null`

backbone.js
```diff
@@ -308,6 +308,7 @@
   Events.once = function(name, callback, context) {
     // Map the event into a `{event: once}` object.
     var events = eventsApi(onceMap, {}, name, callback, _.bind(this.off, this));
+    if (typeof name === 'string' && context == null) callback = void 0;
```
jridgewell: Will this work?

```js
return this.on(events, typeof name === 'string' ? void 0 : callback, context);
```

I believe the issue is because we're converting from the string API into the object API. When using the object API to begin with, it should handle correctly.

jridgewell:
Damn, you merged before I commented. @megawac: Can you check my comment?

backbone.js
megawac: Yeah, that would work as well justin

jridgewell: Can we do that then? `null` should be a valid `context`.

megawac: `null` is a valid context in this patch. Feel free to tweak this implementation though @jridgewell

PR #3988, opened 2016-03-15
Handle Dot Notation in Model Get & Set

alexgurrola:


megawac:
I would prefer to add a `_.get` to underscore which can be overwritten with deep support (and for free with lodash [_.get](https://lodash.com/docs#get).

`Model.prototype.get = (attr) => _.get(this.attributes, attr)`

megawac:
Also -1 on this as we'd also have to support deep set and friends. I'd rather it be handled by lodash/underscore

akre54:
What's the advantage of adding this to core instead of a plugin / recipe? Seems like the kind of thing that's trivial to add yourself if you need it.

megawac:
@akre54 it's significantly less trivial for `set`

backbone.js
```diff
@@ -439,6 +439,21 @@
 
     // Get the value of an attribute.
     get: function(attr) {
+      if (attr.indexOf('.') !== -1) {
+        var ref = this.attributes,
```
akre54: You could probably golf this a bit more:

```js
return attr.split('.').reduce(function(attrs, a) { return attrs && attrs[a]}, this.attributes);
```

akre54:
I think our advice over the years is that deep properties are an anti pattern. Do you disagree?

alexgurrola:
The problem is with complex RESTful payloads that contain nested objects (which I'm using in an API for a Symfony / Doctrine ORM project), which ends up causing inconsistencies with the change sets and syncs when Backbone completely replaces and maintains only the top level keys.

megawac:
I feel like they are unavoidable in many situations as many services output JSON objects with deep properties. Therefore, I think the backbone philosophy should be updated to support modern API decisions

akre54:
That's what the `parse` method is for. I'd be interested in seeing a wholistic solution to the problem, this one is pretty half-baked at the moment.

megawac:
`parse` is annoying to work with as you have to make a custom function per payload, for both creation and for updating the server.

The implementation in this pull request is obviously half baked. I proposed a solution (delegate to `get`/`set` if they exist) which would be flexible and easy to maintain

alexgurrola:
I agree, there are better ways to build a solution, as in porting an _.get from lodash per @megawac, but this code at the moment is all that has kept my project using Backbone.js with a Modern API.

Tell me you wouldn't want dot notation with a JSON output as such:

`
{"route":{"controller":"MenuLink"},"meta":{"method":"get","sentinel":{"view":false,"create":false,"edit":false,"delete":false,"publish":false,"design":false,"dev":false,"master":false},"editUrl":"//admin.sitetheory.io/Menu/Link/Edit","relations":[{"type":"HasOne","key":"parent","relatedModel":"MenuLink","reverseRelation":{"key":"children","relatedModel":"MenuLink"}}],"allowedAction":{"view":true,"create":true,"edit":true,"delete":true,"publish":false,"dev":false},"countCurrent":8,"countTotal":8,"pageCurrent":1,"pageTotal":1,"status":[{"code":"SUCCESS","message":"Successfully executed request."}]},"payload":[{"id":177,"time":1457749536,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1457749536,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":null,"iconHtml":null,"routing":null,"name":"Untitled","url":null,"priority":99,"browserTarget":null,"accessHide":false,"linkTarget":null,"active":null,"contentInfo":null,"route":null,"level":1},{"id":176,"time":1457749532,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1457749532,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":null,"iconHtml":null,"routing":null,"name":"Untitled","url":null,"priority":99,"browserTarget":null,"accessHide":false,"linkTarget":null,"active":null,"contentInfo":null,"route":null,"level":1},{"id":164,"time":1457740297,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1457740321,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":35275,"iconHtml":null,"routing":null,"name":"Christmas","url":null,"priority":1,"browserTarget":"","accessHide":false,"linkTarget":35275,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/All-I-Want-For-Christmas","level":1},{"id":126,"time":1443046118,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1457741195,"userId":2,"editUserId":2,"menu":27,"parent":164,"view":35170,"iconHtml":null,"routing":null,"name":"Buy Now","url":null,"priority":1,"browserTarget":"","accessHide":false,"linkTarget":35170,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/Buy-Now","level":2},{"id":38,"time":1442950123,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1452026589,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":35167,"iconHtml":null,"routing":null,"name":"Home","url":null,"priority":5,"browserTarget":"","accessHide":false,"linkTarget":35167,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/","level":1},{"id":125,"time":1443046113,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1446256038,"userId":2,"editUserId":1,"menu":27,"parent":null,"view":35173,"iconHtml":null,"routing":null,"name":"Sample Stream","url":null,"priority":2,"browserTarget":"","accessHide":false,"linkTarget":35173,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/Sample-Stream","level":1},{"id":45,"time":1442954349,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1443046083,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":35168,"iconHtml":null,"routing":null,"name":"Blog","url":null,"priority":3,"browserTarget":null,"accessHide":false,"linkTarget":35168,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/Blog","level":1},{"id":44,"time":1442954330,"vendor":null,"status":1,"timeStatus":null,"timeEdit":1443046077,"userId":2,"editUserId":2,"menu":27,"parent":null,"view":35171,"iconHtml":null,"routing":null,"name":"Portfolio","url":null,"priority":4,"browserTarget":null,"accessHide":false,"linkTarget":35171,"active":null,"contentInfo":{"image":null,"icon":null,"menuLink":null},"route":"/Portfolio","level":1}]}
`

akre54:
I'm skeptical but willing to take a look at an implementation. Seems like there'd be a ton of edge case behavior questions to work out.

How would that work with deeply nested change events? If I call `model.set('a.b.c', 1)` what does the change event look like? How do we decide when to merge / overwrite? Most other libraries with support for deeply nested properties have a pretty large API surface area for dealing with these sorts of questions.

megawac:
Yeah, `change:*` events are what I'm most worried about. It'd be simple to create a change event for `model.set('a.b.c', 1)` `"a.b.c"` but less simple to create one for `"a"` and `"a.b"` if that is even desired.

It becomes especially complex when you begin considering support for other forms of properties (such as `a['b'][1].x`

alexgurrola:
The dot notation should only be used to cascade into the object, while still leaving the requested attribute string intact for the changeset, which is far simpler to maintain internally. For example, `model.set('a.b', 'c')` would make a `changed['a.b'] = 'c'` so the trigger would be the same key for the event as the attribute specified.  Even my half-baked solution works this way.  ;)

akre54:
@alexgurrola all that text is very distracting, can you please trim it down? We know what a large JSON response looks like.

It also looks like most of that information would be better handled in some form of nested objects / collections. Your payload is a single level structure, which is perfect for a collection. You're not really directly reading those properties on one model, are you? Especially for the `relation` / `meta` objects, it seems like you should set those up when you parse the response.

alexgurrola:
@akre54 Correct.  The payload is creating a Collection, but since each Model has nested ORM objects, each one uses the string notation to listen for an event the same way it gets and sets for consistency.

akre54:
> but since each Model has nested ORM objects, each one uses the string notation

What does this mean? I don't see any nesting in `payload`.

akre54:
>  It becomes especially complex when you begin considering support for other forms of properties (such as a['b'][1].x

Yeah this is why I'm thinking that `model.get('a').get('b').get('c').first()` is a better idea. You can listen to changes on only the models you want. If there's a way we can make this a little friendlier for plugins I'm all ears, but I'm not sure it's worth ripping out our core behavior to support this.

alexgurrola:
@akre54 Here's a better example. http://hastebin.com/izikapigod.js

elgubenis:
i think these plugins are pretty cool to handle your issues

https://github.com/powmedia/backbone-deep-model
https://github.com/afeld/backbone-nested

especially backbone-nested helped us a lot (we do some kind of illustrator thing where objects have deep properties like style.font.size etc. in conjuction with patch: true saving, for atomic puts/patches)

alexgurrola:
@elgubenis  I agree, those are excellent plugins for deep models.

Only downside for me is that we use [Backbone-Relational](http://backbonerelational.org/) to connect the nested Collections inside the Models, which allows tagging of other types in our API and using references to maintain the nested Models and Event bubbles consistently.

I could probably just overwrite the Backbone.Model prototype with the Backbone.DeepModel, though, since it doesn't seem to use any Backbone.Model.prototype functions anywhere in the source.

elgubenis:
cool, we have the same combination, i just can't remember how we connected them to work together, i remember we had a problem with recursive event triggering combining both, but was fixed with doing the deep set silently, I hope you wont run into the same issues

alexgurrola:
@elgubenis I'll try it out and see where that road leads.  Much appreciated.

PR #3991, opened 2016-03-22
Update backbone.js

zgl2007dj:
1 ) improve the patch method
 2 ) _.size is not nessesary, bool will be better

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -562,12 +562,14 @@
       if (!diff) return this.hasChanged() ? _.clone(this.changed) : false;
       var old = this._changing ? this._previousAttributes : this.attributes;
       var changed = {};
+      var hasChanged;
       for (var attr in diff) {
         var val = diff[attr];
         if (_.isEqual(old[attr], val)) continue;
         changed[attr] = val;
+        hasChanged = true;
       }
-      return _.size(changed) ? changed : false;
+      return hasChanged ? changed : false;
     },
 
     // Get the previous value of an attribute, recorded at the time the last


backbone.js
```diff
@@ -644,7 +646,7 @@
       if (attrs && wait) this.attributes = _.extend({}, attributes, attrs);
 
       var method = this.isNew() ? 'create' : (options.patch ? 'patch' : 'update');
-      if (method === 'patch' && !options.attrs) options.attrs = attrs;
+      if (method === 'patch' && !options.attrs) options.attrs = this.changedAttributes(attrs);
```
akre54: That's not how this method is supposed to work but the rest of the pull is great. Back out this change, and I'll merge. Thanks!

zgl2007dj: Thanks, I fixed it.
 But,  i want the method work like that, and overrided model's set function in my app. Maybe not the best way : )

akre54: You can override `Model#save` or `Model#sync` to add this in your own app if you need it.

akre54:
Can you squash down into a single commit and I'll merge. Thanks

backbone.js
```diff
@@ -562,12 +562,14 @@
       if (!diff) return this.hasChanged() ? _.clone(this.changed) : false;
       var old = this._changing ? this._previousAttributes : this.attributes;
       var changed = {};
+      var hasChanged;
       for (var attr in diff) {
         var val = diff[attr];
         if (_.isEqual(old[attr], val)) continue;
-        changed[attr] = val;
+        changed[attr] = val || '';
```
jridgewell: Why are we forcing this into a string?

zgl2007dj: if changed[a] = undefined
and i post changed to server,  will not work
how to deal with it ? 
` changed[attr] = val || null;`
better?

jridgewell: You should normalize in your own code, not in the library.

akre54: Agreed.

zgl2007dj: @jridgewell 
yeah, I fixed it . 
@akre54 
I had squashed down into a new commit

akre54:
This is still showing up as 7 commits. Try running `git rebase -i HEAD~7`, squash each commit into its previous so you only have one remaining, and then force-push up to github.

rafde:
Doesn't merge also have the option to squash merge now?

akre54:
Yup! They just added it. Gonna test it out now I guess :)

PR #3997, opened 2016-03-28
fix a typo

zotherstupidguy:
fix a typo

### DIFF
--- a/index.html
+++ b/index.html
@@ -3264,7 +3264,7 @@ inbox.messages.fetch({reset: true});
 
     <ul>
       <li>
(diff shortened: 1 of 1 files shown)


captbaritone:
Probably not worth reverting, but I'm not convinced this is actually a typo. "a"/"an" is determined by the initial sound of the following word. The [Wikipdia page](https://en.wikipedia.org/wiki/SQL) lists both /ˈɛs kjuː ˈɛl/ and /ˈsiːkwəl/ as valid pronunciations.

akre54:
Yeah totally. Depends on which one you say (I personally pronounce it "sequel"), but you're right both are valid. Reverting.

PR #4000, opened 2016-03-29
Fix failing global tests

akre54:
This should fix two of the failures in the test cases.

The other failure seems to be that in IE8 `onhashchange` is added to `window`. I'm not familiar enough with `attachEvent` (does it automatically create a global `onhashchange`?), but it seems that us calling `Backbone.history.stop()` isn't enough to remove the global. Anyone with IE experience care to shed some light?

There were a few other failures in https://travis-ci.org/jashkenas/backbone/builds/119075098 around default model attributes in IE8 too.

### DIFF
--- a/test/view.js
+++ b/test/view.js
@@ -14,6 +14,11 @@
         className: 'test-view',
         other: 'non-special-option'
       });
+    },
+
+    afterEach: function() {
+      $('#testElement').remove();
+      $('#test-view').remove();
     }
 
   });


akre54:
There were a few other failures in https://travis-ci.org/jashkenas/backbone/builds/119075098 around default model attributes, but they seem to be passing in this test.

megawac:
@akre54 when you're trying to debug sauce failing in some browsers, create a branch on jashkenas/backbone so travis runs sauce

megawac:
Still 3 sauce failures post merge:
https://travis-ci.org/jashkenas/backbone/builds/120952966

akre54:
> create a branch on jashkenas/backbone so travis runs sauce

Good call. 

> Still 3 sauce failures

Right. One is the global `onhashchange` (does calling `attachEvent('hashchange', fn)` create the `onhashchange` for you? Should we delete that prop manually after the test?

The second two are IE8 defaults / prototype bugs. @jridgewell any thoughts?

jridgewell:
This is issue with IE8 is overriding enumerability of properties that are non-enumerable (like `#hasOwnProperty`. The [for-in loop](https://github.com/jashkenas/backbone/blob/master/backbone.js#L498-L507) won't traverse it, since it's "non-enumerable" in IE8. :unamused:

akre54:
Time to drop IE8 support? (no immediate changes, just no longer testing / edge case for it)

jridgewell:
That's fine with me.

akre54:
#4008

PR #4006, opened 2016-04-04
Clarification on the use of model.id

jammon:
The docs should state explicitly that `model.id` should not be manipulated directly. (See #4004)


index.html
```diff
@@ -1243,6 +1243,7 @@ <h2 id="Model">Backbone.Model</h2>
       A special property of models, the <b>id</b> is an arbitrary string
       (integer id or UUID). If you set the <b>id</b> in the
       attributes hash, it will be copied onto the model as a direct property.
+      <code>model.id</code> should not be manipulated directly.
```
jridgewell: Please direct them to using `#set`.

PR #4008, opened 2016-04-06
Stop testing IE8 in sauce

akre54:
In the future we can talk about removing some of the kludges for oldIE. For now we just won't be testing for <= IE8, with the goal of eventually refactoring some of the older hacks out.


jridgewell:
:+1:. Let's wait for another contributor, though?

victoryboss:
I think somebody has to checkout a branch to support IE8 .

mr47:
@victoryboss i think there is no point for it. IE died.

PR #4009, opened 2016-04-07
model defaults deep copy for array params

luki2678511:
if model defaults have array or [] params, then new Model() for array will be error data.
example:

    var Defaulted = Backbone.Model.extend({
      defaults: {
          array: []
      }
    });
    model = new Defaulted();
    assert.equal(model.get('array').length, 0);
    model.get('array').push("obj1");
    model = new Defaulted();
    assert.equal(model.get('array').length, 0);

last equal will fail, array.length is 1.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -399,7 +399,22 @@
     if (options.collection) this.collection = options.collection;
     if (options.parse) attrs = this.parse(attrs, options) || {};
     var defaults = _.result(this, 'defaults');
-    attrs = _.defaults(_.extend({}, defaults, attrs), defaults);
+    // do deep copy (model defaults hava array or [] value will generating dirty data)
+    var new_default = {};
+    if (defaults) {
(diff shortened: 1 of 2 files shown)


akre54:
This is how prototypes in javascript work. Please do a search through the issues for some background.

luki2678511:
Thanks so much, I will learn it. but how can i use array for model prototypes?

akre54:
Turn `defaults` into a method, returning your defaults object.

On Thu, Apr 7, 2016, 21:13 luki2678511 <notifications@github.com> wrote:

> Thanks so much, I will learn it. but how can i use array for model
> prototypes?
>
> —
> You are receiving this because you modified the open/close state.
> <https://github.com/jashkenas/backbone/pull/4009#issuecomment-207159290>
>
-- 

Adam K (mobile)

PR #4011, opened 2016-04-10
Issue 3996 - Self/Global null fix

### LINKED ISSUE #3996: self (or global) may be null and self.self will throw a exception
`var root = (typeof self == 'object' && self.self === self && self) ||
            (typeof global == 'object' && global.global === global && global);`

so is it `typeof self == 'object' && self && self.self === self` right?

melissawimberly:
This should fix the problem.
Other contributors: @kaidez @wuworkshop @jrmils89 

fixes #3996

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -9,8 +9,8 @@
 
   // Establish the root object, `window` (`self`) in the browser, or `global` on the server.
   // We use `self` instead of `window` for `WebWorker` support.
-  var root = (typeof self == 'object' && self.self === self && self) ||
-            (typeof global == 'object' && global.global === global && global);
+  var root = (typeof self == 'object' && self !== null && self.self === self && self) ||
+            (typeof global == 'object' && global !== null && global.global === global && global) || {};
 
   // Set up Backbone appropriately for the environment. Start with AMD.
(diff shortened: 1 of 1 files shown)


akre54:
Same as #3996 -- is this a real-world issue, or just a hypothetical?

siliconforks:
Wouldn't this be an actual real-world issue if you had a variable `self` with value `null`?  For example, consider the following node.js code:
```node
global.self = null;
const Backbone = require('backbone');
console.log(Backbone.VERSION);
```
If I try to run this with node.js I get:
```
  var root = (typeof self == 'object' && self.self === self && self) ||
                                             ^

TypeError: Cannot read property 'self' of null
```
Note that, in order to fix this, it would also be necessary to fix underscore.js (because it has the same issue).

ogonkov:
It is better to have build tool insert this wrapper

jgonggrijp:
Agreed @ogonkov. Planning to use Rollup for this, as I have done with Underscore.

Also, this is a non-issue, like the linked ticket. Not sure why I put this in the fridge instead of closing it immediately. Perhaps I wanted to reply to the OP, but I guess that isn't necessary.

Issue #3999, opened 2016-03-29
Why I can't name my event as 'constructor'?


unclechu:
``` javascript
_.extend({}, Backbone.Events).on('constructor', function () {});
```

Throws exception:

``` text
TypeError: handlers.push is not a function
```

Is it a bug?
I can reproduce it in developer console on backbonejs.org site.

Florian-R:
FWIW, this is not the same bug but somehow related to #1495, which was closed as wontfix.

megawac:
This can be fixed by checking if the [`handler` is an array](https://github.com/jashkenas/backbone/blob/master/backbone.js#L206). 

```js
var handlers = _.isArray(events[name]) ? events[name] : (events[name] = []);
```

Thoughts @jridgewell @akre54

akre54:
Eh. I think our short circuit pattern is fairly common in JS. The advice around edge cases in #1495 applies here too.

jridgewell:
Can we just drop IE8 now? Hell, even IE8 can do a prototype-less object with [black magic](https://github.com/es-shims/es5-shim/blob/master/es5-sham.js#L225-L260).

akre54:
What you got in mind? Object.create everywhere?

On Tue, Mar 29, 2016, 13:08 Justin Ridgewell <notifications@github.com>
wrote:

> Can we just drop IE8 now? Hell, even IE8 can do a prototype-less object
> with black magic
> <https://github.com/es-shims/es5-shim/blob/master/es5-sham.js#L225-L260>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
> <https://github.com/jashkenas/backbone/issues/3999#issuecomment-203002167>
>
-- 

Adam K (mobile)

mr47:
@jridgewell my god.

megawac:
> Hell, even IE8 can do a prototype-less object with black magic.

You're not kidding, I'm pretty sure that `ActiveXObject` is literally a gate to hell

akre54:
That's insane

On Tue, Mar 29, 2016, 13:30 Graeme Yeates <notifications@github.com> wrote:

> Hell, even IE8 can do a prototype-less object with black magic.
>
> You're not kidding, I'm pretty sure that ActiveXObject is literally a
> gate to hell
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly or view it on GitHub
[...]

Issue #3874, opened 2015-11-30
Object.prototype collisions of model values in Collection._byId lookup


msiconolfi:
This issue seems related to [#3842](https://github.com/jashkenas/backbone/issues/3842) but this also happens for values as well as keys.

In FireFox if you create a collection that uses the `idAttribute` field to do a id lookup. The lookup will return the wrong value if the value of a models idAttribute is also a value on Object.prototype. For example, `watch`.

In the referenced issue, people have mentioned not using prototype names as default properties, which I can understand, but attribute values should not have a limitation as we aren't always in control of what those values may be.

Here is an example:
```javascript
var M = Backbone.Model.extend({
    idAttribute: 'type',
    defaults: {
        'type': '',
        'name': ''
    }
});

var model1 = new M({'type': 'watch', name: 'Watch'});
var model2 = new M({'type': 'other', name: 'Other'});

var C = Backbone.Collection.extend({
    model: M
});

var collection = new C();

// add the models
collection.add(model1);
collection.add(model2);

// try and get the model with type: 'watch'
console.log(collection.get('watch'));
console.log(collection.length);

// LOG: function watch()
// LOG: 1
```

The above example only happens in firefox but in Chrome you can see the same behavior if you add the following:

```javascript
var model3 = new M({'type': 'constructor', name: 'Constructor'});

collection.add(model3);
console.log(collection.get('constructor'));

// LOG: function Object() { [native code] }
```

msiconolfi:
I think the problem is in the `Collection.get()` [method](https://github.com/jashkenas/backbone/blob/master/backbone.js#L970), it simply returns the first truthy object that indexes into `_byId` lookup. In Firefox, `Object.prototype` has the `watch` property so this will always return a truthy result, and because `set()` tests for [existence](https://github.com/jashkenas/backbone/blob/master/backbone.js#L854) using `.get()` the `watch` model will never be added.
```javascript
return this._byId[obj] || this._byId[id] || this._byId[obj.cid];
```

It's a little ugly but we could do something like the following:

```javascript
   if (this._byId.hasOwnProperty(obj)){
      return this._byId[obj];
   } else if (this._byId.hasOwnProperty(id)){
      return this._byId[id];
   } else {
     return this._byId[obj.cid];
  }
```

msiconolfi:
I will try to add a PR with this change as well as some tests later today.

msiconolfi:
[...]

Issue #3782, opened 2015-09-01
Model.prototype.get() and .has() counts own properties


winhillsen:
Hi all! I stumbled on the following behaviour:

```javascript
var model = new Backbone.Model();
console.log( model.get( 'constructor' ) );
```
returns the constructor function of the `model.attributes` object.

I was rather expecting the `get()` method to filter own properties and return `undefined`. The same should also hold true with `has()` (in line with Underscore's `_.has()`, which does filter own properties). Without this, a model's `attributes` object would not be a [true hash](http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/).

Thoughts?

Florian-R:
See #1495

akre54:
Something to address for v2? [Object.create is IE9+](http://kangax.github.io/compat-table/es5/#Object.create), might be worth looking into if we drop IE8.

jridgewell:
It can be polyfilled with black magic in IE7 and 8. I think we should do it.

akre54:
There's [prior](https://github.com/paulmillr/exoskeleton/commit/550265b9f41e844506467681bcf7ab04d7a2d2e5) [art](https://github.com/paulmillr/exoskeleton/commit/ffe51bba68a01b11cf3642ac54787b379801190b) in Exoskeleton for `model.attributes` and `collection._byId` from `Object.create(null)`. (see #3411, #1783, etc.)

In general, I'd say that #1783 is right in that there generally isn't a real-world use case for naming your keys "constructor" or "hasOwnProperty" (keep in mind [an object is not a hash](http://www.devthought.com/2012/01/18/an-object-is-not-a-hash/)) but if this change is easy and comes with minimal side effects, let's bite the bullet.

jashkenas:
There may be good reasons to eventually drop IE7 and IE8, but using properties like `constructor` isn't one of them. Let's leave it for now.

komaval:
> There may be good reasons to eventually drop IE7 and IE8, but using properties like constructor isn't one of them. Let's leave it for now.

this trouble is actual on my current project
you made me sad 

in my case treatment is to overwrite `get` for collection 

```
get: function(obj) {
	if (obj == null) return void 0;
	var id = this.modelId(this._isModel(obj) ? obj.attributes : obj);
	return this._byId[obj] || this._byId.hasOwnProperty(id) && this._byId[id] || this._byId[obj.cid];
},
```

PR #4012, opened 2016-04-10
Use prototypeless objects

jrmils89:
@jridgewell @wuworkshop

This makes the switch to use prototypeless objects instead of object literals. This fixes issues referenced in (at least) the following (and probably others):
#3999
#1495
#3842
#1783
#3411
#3782
#3874

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -984,7 +984,7 @@
     // properties, or an attributes object that is transformed through modelId.
     get: function(obj) {
       if (obj == null) return void 0;
-      if (typeof obj === "object") {
+      if (typeof obj === 'object') {
         return this._byId[this.modelId(obj.attributes || obj)] ||
         obj.cid && this._byId[obj.cid];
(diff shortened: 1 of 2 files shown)


akre54:
We did this in Exoskeleton. https://github.com/paulmillr/exoskeleton/issues/84 and a few others popped up, so let's be extra careful if we merge this.

mr47:
@akre54 I think `map` is solution for backbone. It's already 2016 and we still stuck with es5. But what about create a map wrapper for backbone so we can use it as is.  Create a small check for map support. For older browsers - polyfill. Later mb we can create proof of concept and test it.

mr47:
Ps: For v2 I just don't see why not

akre54:
Sure, it'd be great to see a pull that tackles this behavior. Go for it!

akre54:
Closing due to lack of activity.

PR #4024, opened 2016-04-29
Collection & Model Mixin

megawac:
Adds Collection & Model mixins for adding custom implementations of iterator/check methods to base collection and model. The main motivation of this is to add some `lodash` methods that were missing on the prototype.

Ideally one could do

``` js
Backbone.Collection.mixin({
  sum: lodash.sum,
  take: lodash.take
});
```

However, I don't have a good idea of how to support iteratees, so things like `collection.sum('value')`

Thoughts? /cc @akre54 @jridgewell

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -68,54 +68,6 @@
   // form param named `model`.
   Backbone.emulateJSON = false;
 
-  // Proxy Backbone class methods to Underscore functions, wrapping the model's
-  // `attributes` object or collection's `models` array behind the scenes.
-  //
-  // collection.filter(function(model) { return model.get('age') > 10 });
-  // collection.each(this.addView);
-  //
-  // `Function#apply` can be slow so we use the method's arg count, if we know it.
-  var addMethod = function(length, method, attribute) {
-    switch (length) {
-      case 1: return function() {
-        return _[method](this[attribute]);
-      };
-      case 2: return function(value) {
-        return _[method](this[attribute], value);
-      };
-      case 3: return function(iteratee, context) {
-        return _[method](this[attribute], cb(iteratee, this), context);
-      };
-      case 4: return function(iteratee, defaultVal, context) {
-        return _[method](this[attribute], cb(iteratee, this), defaultVal, context);
-      };
-      default: return function() {
-        var args = slice.call(arguments);
-        args.unshift(this[attribute]);
-        return _[method].apply(_, args);
-      };
-    }
-  };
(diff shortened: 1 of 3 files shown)


jridgewell:
> However, I don't have a good idea of how to support iteratees, so things like collection.sum('value')

Hm? You seem to have it already.

megawac:
Does it? I just assumed it wouldn't get transformed to a `(attr) => model.get(attr)` as it doesn't [hit this case](https://github.com/jashkenas/backbone/blob/master/backbone.js#L109) as length != 3 or 4

backbone.js
```diff
@@ -1369,6 +1298,94 @@
...
+
+  // Mix in each Underscore method as a proxy to `Collection#models`.
+
+  _.each([
+    [Collection, collectionMethods, 'models'],
+    [Model, modelMethods, 'attributes']
+  ], function(config) {
+    var Base = config[0],
+        methods = config[1],
+        attribute = config[2];
```
Florian-R: Curious to know why you've used an array of array here instead of an array of object here, e.g: 

```js
_.each([
   {base:Collection, methods: collectionMethods, attribute: 'models'},
   {base:Model,  methods: modelMethods, attribute: 'attributes'}
], function(config) {
```

jridgewell:
> I just assumed it wouldn't get transformed to a (attr) => model.get(attr) as it doesn't hit this case as length != 3 or 4

Oh, didn't read that case. What fails if we use `cb` for every case?

backbone.js
megawac: No real reason, I don't see a major advantage for either strategy at the moment. Though current impl it lends itself nicely to [`_.spread`](https://lodash.com/docs#spread) if underscore ever decides to follow suit

megawac:
I think `_.invoke` would fail in addition to `omit`, `difference`, `union`  and friends, and perhaps the reduce implementations

rafde:
how has this not been merged yet?

elgubenis:
@rafde maybe you can take a look at backbone-metal, i think its something similar

rafde:
this PR is suppose to allow you to define your own functions that would proxy against `model.attributes` && `collection.models`

`backbone-metal` doesn't look like it offers the same feature.

megawac:
Can I get another set of 👀 on this? Want to get it into #4028

akre54:
Fine by me 👍

PR #4027, opened 2016-05-03
Change to splice method so it is observable

robinchesterman:
The existing splice method uses the indexer:

`array[i + at] = insert[i]`

This is not a natively observable change, e.g. using Array.observe.

This PR changes the method to achieve the same effect but using native Array.splice.

This may be useful when using Backbone with other frameworks e.g. it may be desirable to use Backbone modelling for the excellent collection methods and REST APIs, but Aurelia or Angular for UI data-binding.

The latter frameworks can't observe changes on collection.models array if they're done with the indexer. Splice is observable.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -770,12 +770,8 @@
   // Splices `insert` into `array` at index `at`.
   var splice = function(array, insert, at) {
     at = Math.min(Math.max(at, 0), array.length);
-    var tail = Array(array.length - at);
-    var length = insert.length;
-    var i;
-    for (i = 0; i < tail.length; i++) tail[i] = array[i + at];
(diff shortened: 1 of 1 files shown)


megawac:
Spread was added in https://github.com/jashkenas/backbone/pull/3663. I'll defer to @jridgewell on this one

jridgewell:
I removed splice because of its horrible performance in loops (adding an array of models at a particular index).

How is `#splice` observable bot indexed sets aren't?

jgonggrijp:
I'm in favor of this change, but the performance claim needs quantification before the change can be made. By "quantification", I mean something a little more serious than a nanobenchmark on jsperf. Not to mention that jsperf seems to be broken.

PR #4037, opened 2016-05-28
Upgrade ESLint to latest and replace removed rules

platinumazure:
Of note:
* `no-empty-label` was replaced with `no-labels`. If we want to loosen behavior to allow sensible labels, we can make that change, but there were no errors on tightening.
* `space-after-keywords` and `space-return-throw-case` were consolidated into `keyword-spacing`. To preserve behavior

I also removed an unnecessary/redundant set of array brackets on a rule that had no extra options.

### DIFF
--- a/.eslintrc
+++ b/.eslintrc
@@ -5,11 +5,11 @@
     "amd": true
   },
   "globals": {
-    "attachEvent": true,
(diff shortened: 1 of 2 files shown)


jridgewell:
Thanks!

PR #4040, opened 2016-06-08
Adds onFirstListener and onNoListener callbacks

DomBlack:
These callbacks allow anything extending `Backbones.Events` to be
notified when something starts listening to it, and when nothing
is left listening to it.

Because of `internalOn` this is hard to implement in user land.

### Use Case

The use case for this is if you had a Model presenting the output of
an expensive continuous calculation (say real time Stock Prices or
a timer).

The `onFirstListener` function can be used to start your expensive
calculation, as you only care about it while the something is listening
to changes in it.

When that item stops listening to changes in it, `onNoListeners` will
trigger, allowing you to clean up your code (unsubscribe the stock price,
or stop the timer or even stop listening to another Event).

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,10 @@
+# EditorConfig: http://EditorConfig.org
+root = true
+
+[*]
+end_of_line = lf
+insert_final_newline = true
+trim_trailing_whitespace = true
+charset = utf-8
+indent_style = space
+indent_size = 2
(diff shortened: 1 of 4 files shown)


akre54:
Thanks but no thanks. Feel free to do this in your own project if you need it

DomBlack:
NP. I have but I ended up having to do this in Backbone because of #3615 once that PR is merged, this can be done in user land.

Issue #3183, opened 2014-06-06
Backbone.history / Router.navigate {replace:true} not working (related to issue #3182)


egpierro:
Following the same setup on issue #3182, we noticed that the {replace:true} flag is not working as expected; the Backbone documentation states that using this flag will prevent a navigate request to insert a history record in the browser. In our case, we see `module1/dummy` when we hit the browser's back button, after successfully navigating to `module1/home`.

Flow:

```
/ --> module1/home --> back button --> module1/dummy
```

Desired Flow:

```
/ --> module1/home --> back button --> /
```

One colleague suggested adding {replace:true} to the master router that is doing the handoff to the modular router as well, which worked well, but we're not sure why is it that we need to add this flag in two places, since the `module1/dummy` navigate command is being executed after the fact.

Obviously, having a permanent fix to #3182 will null this issue, but it the workaround is unavoidable, we need to find a solution to this unnecessary history addition.

pajtai:
I believe the docs are not quite right. replace:true just calls window.history.replaceState, and this method REPLACES the current state with the one called.... so if you don't want a "page" in your history then you would use replace:true when navigating way from the page, not when navigating to it.

soroushhakami:
I can confirm the behaviour @pajtai mentions. Could be clarified in the docs.

PR #4041, opened 2016-06-09
Clarify Router.navigate docs

scott-martin:
#3183 Clarify the use `{ replace: true }` with Router.navigate - that is if you do not want the current url in history that you should use this option when navigating away from that route. The current documentation led me to believe that it was for navigating to a route that I did not want in history.


akre54:
Thanks but I think it's fine as-is. This is more confusing.

It might be better to link to [`replaceState`](https://developer.mozilla.org/en-US/docs/Web/API/History_API) to compare with `popState`

PR #4042, opened 2016-06-09
Fixing gitattributes to ensure .js and .svg files are treated as text

platinumazure:
Hopefully normalize .js and .svg files as LF only going forward. DO NOT MERGE without testing in a couple of different OS's, engines, and build systems-- there may be other issues I can't see!


akre54:
Thanks @platinumazure! Anyone else want to give this a spin?

platinumazure:
@akre54 Have you ever considered using AppVeyor for CI? (Not instead of Travis, but in addition to Travis) ESLint uses it to help catch Windows-specific issues sometimes.

akre54:
If you feel like setting that up (or pointing us in the right direction), might be a great addition.

PR #4046, opened 2016-06-14
added error when jquery is not installed

doloreel:
calling new View() without jQuery will throw this error: 

`backbone.js:1287 Uncaught TypeError: Right-hand side of 'instanceof' is not an object_setElement @ backbone.js:1287setElement @ backbone.js:1275_ensureElement @ backbone.js:1362Backbone.View @ backbone.js:1223child @ backbone.js:1888(anonymous function) @ main.js:237`

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1301,8 +1301,12 @@
     // alternative DOM manipulation API and are only required to set the
     // `this.el` property.
     _setElement: function(el) {
-      this.$el = el instanceof Backbone.$ ? el : Backbone.$(el);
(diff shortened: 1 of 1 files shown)


Florian-R:
Nope. Using Backbone without jQuery is totally fine, see https://github.com/jashkenas/backbone/wiki/Using-Backbone-without-jQuery

jridgewell:
Closing per https://github.com/jashkenas/backbone/pull/4047#issuecomment-225928137.

PR #4047, opened 2016-06-14
Fix  jQuery dependancy on lines 1286 and 1306

more-like-a-puma:


jaapz:
Now when you call `setElement`, it doesn't actually set the element at all. That does not seem like a good solution to me. What is the use case for this?

jridgewell:
jQuery is a dependency if you want to use Views. If you don't want jQuery, you can [override](https://github.com/jashkenas/backbone/wiki/Using-Backbone-without-jQuery) the few methods that depend on it.

PR #4026, opened 2016-04-30
Fixed updating URL with uri-decoded value

Flamefork:
Fixes #4025

PR #4050, opened 2016-06-15
location.hash broken chracter(Kroean) issue on Safari browser.

stegano:
HI :D,
I has fixed below issue.
location.hash broken chracter(Kroean) issue on Safari browser.

e.g.
console.log("전체") // 전체 character is Korean..
-> #%04%B4 // it's invalid character. I look forward to 전체.

Thank you :)


akre54:
We've had many issues with Unicode characters in URLs over the years due to incompatibilities between Firefox, Safari, and Chrome. Have you tried #4026?

PR #4051, opened 2016-06-15
Add Korean character test spec

stegano:
#4050 location.hash broken character(Kroean) issue on Safari.
THX :D


Florian-R:
Why don't add the test in #4050? This one should also probably need a run on SauceLabs also.

PR #4060, opened 2016-07-19
separate view's dom target from its constructed element

craigmichaelmartin:
This PR allows for a separation of a view's dom target from the view's constructed element, so that a view with a dom target can have access to the constructed element before it is in the dom.

Currently, if a view is assigned an `el`, when the template html is appended to it (with whatever custom rendering process), it is reflected in the dom. You could use a temporary element, but then you lose the niceties of this.el's integration in the view. It seems like this.el should be precisely the element the view builds, and separate from the target element in the dom. This PR is backwards compatible with current behavior, but allows for this separation. A non-theoretical use case for the separation is for plugins - to be able to pass actual elements for transformation before they are added to the dom. In order to do this now, an `el` must not be passed to the view, and something must sit over the view to manage this. This proposal would allow for the separation of concerns by the view itself.

I could very well be missing something in backbone that would already foster this; or maybe I am too entrenched in how I see views and need to re-imagine their construction. Part of the difficulty in writing this is that there are so many possible scenarios for views - from simple wrappers for already rendered dom elements, to deeply nested hierarchies in single page apps.

If there is positive feedback I'll add tests and update the docs. Thanks for your feedback!

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1242,7 +1242,7 @@
   var delegateEventSplitter = /^(\S+)\s*(.*)$/;
 
   // List of view options to be set as properties.
-  var viewOptions = ['model', 'collection', 'el', 'id', 'attributes', 'className', 'tagName', 'events'];
+  var viewOptions = ['model', 'collection', 'el', 'target', 'id', 'attributes', 'className', 'tagName', 'events'];
 
   // Set up all inheritable **Backbone.View** properties and methods.
   _.extend(View.prototype, Events, {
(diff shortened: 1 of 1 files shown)


rafde:
Usually the tests would provide insight on the intentions for this feature. Without the test, I can't imagine a scenario where you need this.

jridgewell:
I'm with @rafde here, can you provide sample code that this would help solve?

craigmichaelmartin:
It solves the issue of `el` being the dom target (if specified), and the de facto element to build your view's html into. Not having this separated forfeits the view's own ability to have fine grained control over the element that it is building, and managing its insertion into the dom.

Here is an example use case with this code and a jQuery plugin (slick carousels). With the new code I can pass in a target and so can separate building the view's element, and inserting into the dom. In this example case, it is beneficial because I only want to insert the transformed element into the dom (I don't want it to hit the dom, and then be queried out of the dom, and then put back in). Here is a basic fiddle for this case. https://jsfiddle.net/owwfhqzb/

craigmichaelmartin:
A slight difference is that `remove` operates on `el` and not `target`. IMO, this is nice. I want my view to clean up everything encapsulated in it, but I don't want the view to reach outside of itself and remove the dom element it happened to be placed into. A vast majority of the time (for me, and again maybe this is just how I've envisioned and use views) the hook a view uses is used by sibling views and should remain.

akre54:
That example isn't particularly convincing. You're just wrapping the `$el`.

Absent a better use case, I think this looks like the job for a Backbone plugin or a design pattern.

craigmichaelmartin:
I find the ability for a view to both have a dom reference, and separately have its own constructed element, pretty fundamental. This allows for a view to control when/how it's own constructed element is associated with the dom. Right now a view's `el` seems conflated, but it must be how I'm approaching views. Thanks for you feedback!

PR #4061, opened 2016-07-20
Implements the JavaScript Iterable protocol.

### LINKED ISSUE #3954: Collection for-of support
Would be nice if we could write 

`for (let model of collection) ...`

Maybe all it would take is adding:

```
if (areSymbolsAndGeneratorsSupported) {
  Collection.prototype[Symbol.iterator] = function* () {
    for (let model of this.models) yield model
  }
}
```
Browsers without symbols and generators (probably) don't have for-of anyway and if you transpile you have all of them. 

What do you think?

leebyron:
This introduces new methods related to Iterators on Backbone.Collection to mirror those found on Array: `values`, `keys`, `entries`, and `@@iterator`. Each of these methods will return a JavaScript Iterator, which has a `next` method, yielding the models or ids of models contained in the Collection.

The CollectionIterator is careful to use the `at()` and `modelId()` methods on the host collection rather than direct access to the `models` property, which should ensure it is resilient to creative subclassing of Backbone.Collection and future feature addition.

CollectionIterator's `next` method follows the same [algorithm defined by ES6 for ArrayIterator's `next`](http://www.ecma-international.org/ecma-262/6.0/#sec-%arrayiteratorprototype%.next).

The [`@@iterator`](http://www.ecma-international.org/ecma-262/6.0/#sec-well-known-symbols) method is defined using `Symbol.iterator` if it exists in the JavaScript runtime (modern browsers/node.js) and falls back to the string `"@@iterator"` which was popularized by older versions of Firefox and has become the standard fallback behavior for other third-party libraries. This ensures that Backbone can still be used across all browsers, even with use of these new methods.

Supporting Iterable allows better integration between Backbone and the most recent additions to the JavaScript language, including `for of` loops and data-collection constructor functions, as well as better integration with other third-party libraries that accept Iterables instead of only Arrays.

Fixes #3954

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1106,6 +1106,21 @@
       return attrs[this.model.prototype.idAttribute || 'id'];
     },
 
+    // Get an iterator of all models in this collection.
+    values: function() {
+      return new CollectionIterator(this, ITERATOR_VALUES);
+    },
+
+    // Get an iterator of all model IDs in this collection.
+    keys: function() {
+      return new CollectionIterator(this, ITERATOR_KEYS);
+    },
+
+    // Get an iterator of all [ID, model] tuples in this collection.
+    entries: function() {
+      return new CollectionIterator(this, ITERATOR_KEYSVALUES);
+    },
+
     // Private method to reset all internal state. Called when the collection
     // is first initialized or reset.
     _reset: function() {
@@ -1201,6 +1216,72 @@
     }
 
   });
+
+  // Defining an @@iterator method implements JavaScript's Iterable protocol.
+  // In modern ES2015 browsers, this value is found at Symbol.iterator.
+  /* global Symbol */
+  var $$iterator = typeof Symbol === 'function' && Symbol.iterator;
+  if ($$iterator) {
+    Collection.prototype[$$iterator] = Collection.prototype.values;
+  }
+
+  // CollectionIterator
+  // ------------------
+
+  // A CollectionIterator implements JavaScript's Iterator protocol, allowing the
+  // use of `for of` loops in modern browsers and interoperation between
+  // Backbone.Collection and other JavaScript functions and third-party libraries
+  // which can operate on Iterables.
+  var CollectionIterator = function(collection, kind) {
+    this._collection = collection;
+    this._kind = kind;
+    this._index = 0;
+  };
+
+  // This "enum" defines the three possible kinds of values which can be emitted
+  // by a CollectionIterator that correspond to the values(), keys() and entries()
+  // methods on Collection, respectively.
+  var ITERATOR_VALUES = 1;
+  var ITERATOR_KEYS = 2;
+  var ITERATOR_KEYSVALUES = 3;
+
+  // All Iterators should themselves be Iterable.
+  if ($$iterator) {
+    CollectionIterator.prototype[$$iterator] = function() {
+      return this;
+    };
+  }
+
+  CollectionIterator.prototype.next = function() {
+    if (this._collection) {
+
+      // Only continue iterating if the iterated collection is long enough.
+      if (this._index < this._collection.length) {
+        var model = this._collection.at(this._index);
+        this._index++;
+
+        // Construct a value depending on what kind of values should be iterated.
+        var value;
+        if (this._kind === ITERATOR_VALUES) {
+          value = model;
+        } else {
+          var id = this._collection.modelId(model.attributes);
+          if (this._kind === ITERATOR_KEYS) {
+            value = id;
+          } else { // ITERATOR_KEYSVALUES
+            value = [id, model];
+          }
+        }
+        return {value: value, done: false};
+      }
+
+      // Once exhausted, remove the reference to the collection so future
+      // calls to the next method always return done.
+      this._collection = void 0;
+    }
+
+    return {value: void 0, done: true};
+  };
 
   // Underscore methods that we want to implement on the Collection.
   // 90% of the core usefulness of Backbone Collections is actually implemented
--- a/test/collection.js
+++ b/test/collection.js
@@ -1758,6 +1758,60 @@
     assert.equal(c2.modelId(m.attributes), void 0);
   });
 
+  QUnit.test('Collection implements Iterable, values is default iterator function', function(assert) {
+    /* global Symbol */
+    var $$iterator = typeof Symbol === 'function' && Symbol.iterator;
+    // This test only applies to environments which define Symbol.iterator.
+    if (!$$iterator) {
+      assert.expect(0);
+      return;
+    }
+    assert.expect(2);
+    var collection = new Backbone.Collection([]);
+    assert.strictEqual(collection[$$iterator], collection.values);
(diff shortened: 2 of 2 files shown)


jdalton:
What I don't like about this is the fallback use of `@@iterator` since it introduces an inconsistency in older and newer environments in that older environments will have an enumerable string key, "@@iterator", while newer ones will have a symbol key (which isn't included in things like `Object.keys`). I know this has caused some bugs in Ramda (a lib implementing this style) too.

jdalton:
I'm fine with it being a modern only enhancement.

leebyron:
That's definitely a concern to weigh. I'm curious to know more about the issues Ramda encountered. I would be surprised if `Object.keys()` presented an issue, since that's a pretty uncommon operation in general on a Backbone.Collection but even if it was performed, it should not include the properties of the Collection.prototype (unless of course you're calling `Object.keys()` directly on Collection.prototype). If this is a serious concern, we could use `Object.defineProperty` instead to ensure it is non-enumerable.

Another potential approach to avoid inconsistency is to *always* define `Collection.prototype["@@iterator"]` while only defining `Collection.prototype[Symbol.iterator]` should `Symbol.iterator` be defined.

An unfortunate issue of only implementing Iterable when Symbol is defined is the potential for inconsistent behavior between browsers when in use with libraries that support the fallback (Ramda being a fine example).

jdalton:
> I'm curious to know more about the issues Ramda encountered. 

The issues [here](https://github.com/ramda/ramda/issues/1306) and [here](https://github.com/TheLudd/identity-transduce-hack/pull/2).

> An unfortunate issue of only implementing Iterable when Symbol is defined is the potential for inconsistent behavior between browsers when in use with libraries that support the fallback (Ramda being a fine example).

Naw, there wouldn't be an issue because Backbone wouldn't have the fallback so it wouldn't be detected in those libs. If folks really want support there's `Symbol` shims in es6-shim and core-js.

megawac:
I agree with JDD, only export iterators if `Symbol.iterator` is present in the environment or shimmed -- I'm not a fan of the `@@iterator` approach

backbone.js
```diff
@@ -1202,6 +1217,61 @@
...
+    if (this._collection) {
+
+      // Only continue iterating if the iterated collection is long enough.
+      if (this._index < this._collection.length) {
+        var model = this._collection.at(this._index);
+        this._index++;
+
+        // Construct a value depending on what kind of values should be iterated.
+        var value;
+        if (this._kind === 'value') {
```
megawac: I think I'd rather define an ENUM of kinds rather than use strings but not strongly opposed to this

leebyron:
Thanks for the issues, @jdalton that's informative. I don't think those apply here - those issues are about *consuming* iterables rather than implementing them. The issue there is that ramda was detecting iterables by looking for Symbol.iterator if it exists, but falling back to "@@iterator" - which meant that if an implemented iterable *only* implemented "@@iterator" that ramda wouldn't detect it was iterable in environments with Symbol defined. The proposed solution in that issue is a good one: to check for either Symbol.iterator or "@@iterator" existence to determine if something is iterable regardless of if Symbol is defined. For example, React does this kind of [double-detection](https://github.com/facebook/react/blob/master/src/shared/utils/getIteratorFn.js#L34) in its support of iterables. 

For libraries that *implement* iterable, there's very little downside to using the fallback. Libraries consuming iterables which only look for Symbol.iterator would behave identically when given an Iterable which does or doesn't use the fallback. Older versions of Firefox will suddenly work correctly (which support Iterables but use the pseudo-symbol). Libraries which look for both will benefit from having iterables define the fallback without requiring es6-shim or core-js. The downside is that there is an additional prototype method defined which, as you mentioned, appears if you Object.keys() the prototype.

Anyhow, I can remove it since you both feel strongly about it. Supporting Iterable in ES6 (or shimmed) environments is still great even without using the fallback pseudo-symbol.

backbone.js
leebyron: Do you have a preferred enum style in mind? I don't feel strongly about this, happy to take suggestions.

backbone.js
```diff
@@ -1202,6 +1217,69 @@
...
+  // which can operate on Iterables.
+  var CollectionIterator = function(collection, kind) {
+    this._collection = collection;
+    this._kind = kind;
+    this._index = 0;
+  };
+
+  var KIND_VALUE = 1;
+  var KIND_KEY = 2;
+  var KIND_KEY_VALUE = 3;
```
leebyron: @megawac let me know if you prefer this enum style over strings

jdalton:
>  I don't think those apply here - those issues are about consuming iterables rather than implementing them.

Let's avoid the potential headache.

> Older versions of Firefox will suddenly work correctly (which support Iterables but use the pseudo-symbol).

User adoption of new FF releases is pretty fast. The '@@iterator' property was used from  FF27-FF35 which is less than 0.2% in Dec 2015.

> Anyhow, I can remove it since you both feel strongly about it. Supporting Iterable in ES6 (or shimmed) environments is still great even without using the fallback pseudo-symbol.

Great!

backbone.js
megawac: Yep, I prefer this for sure. Mind renaming these `ITERATOR_VALUEs`, `ITERATOR_KEYS`, `ITERATOR_KEYVALUES`

leebyron:
> Great! As a result of this PR I've updated Lodash's checks to allow shims too.

🎉

megawac:
LGTM, I'll defer to another contrib to merge /cc @jridgewell @akre54

backbone.js
```diff
@@ -1202,6 +1217,69 @@
...
+        var model = this._collection.at(this._index);
+        this._index++;
+
+        // Construct a value depending on what kind of values should be iterated.
+        var value;
+        if (this._kind === ITERATOR_VALUES) {
+          value = model;
+        } else {
+          var id = this._collection.modelId(model.attributes);
+          if (this._kind === ITERATOR_KEYS) {
```
megawac: another option to using these enums to indicate what to place as value would be to have the `values`, `keys`, and `entries` pass a `function(model, collection)` to define the value. I'm fine with this either way.

backbone.js
```diff
@@ -1202,6 +1217,69 @@
...
+  // use of `for of` loops in modern browsers and interoperation between
+  // Backbone.Collection and other JavaScript functions and third-party libraries
+  // which can operate on Iterables.
+  var CollectionIterator = function(collection, kind) {
+    this._collection = collection;
+    this._kind = kind;
+    this._index = 0;
+  };
+
+  var ITERATOR_VALUES = 1;
```
megawac: mind leaving an inline comment on what these variables are for

akre54:
Looks great. Thanks @leebyron! I think we're at the point where if you need this you're probably targeting evergreen browsers anyways. We can reassess if anyone feels particularly strongly.

leebyron:
Comments added. I agree, @akre54, targeting evergreen browsers is good practice. I'm happy with the state of this PR.

jridgewell:
I dig it. 👍

PR #4075, opened 2016-09-16
Add filtering to documentation

msrose:
This is a direct port of @captbaritone's changes in jashkenas/underscore#2514, as suggested by @jashkenas. It allows you to quickly filter through the documentation by the listings in sidebar, and jump to the first match by pressing Enter. The functionality can be tried out here: https://rawgit.com/msrose/backbone/docs-search/index.html.

I've made every list item in the sidebar searchable based on the inner text of the contained `<a>` tag. There is a greater variety of items listed in the backbone docs compared to the underscore ones, but I think it still makes sense to have everything searchable based on the current implementation of the filtering.

The code includes includes the ability to filter by "alias" (i.e an alternate name for a function), which is useful in underscore but has no parallel in backbone as far as I can tell. The code to support aliases is minimal and may come in handy in the future, so I've left it in.

### DIFF
--- a/docs/search.js
+++ b/docs/search.js
@@ -1,0 +1,58 @@
+(function() {
+  var functions = document.querySelectorAll('[data-name]');
+  var sections = document.querySelectorAll('.searchable_section');
+  var searchInput = document.getElementById('function_filter');
+
+  function strIn(a, b) {
+    a = a.toLowerCase();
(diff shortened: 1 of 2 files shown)


jashkenas:
Fantastic! Thanks for including the live demo.

Issue #4034, opened 2016-05-24
Feature Suggestion: Backbone.Model.prototype.update


benmccormick:
A fairly common pattern I find myself typing when working with Backbone models involves code like this:

```
//iterating
model.set('totalCount', model.get('totalCount') + 1);
model.set('team1Count', model.get('team1Count') + 1);

//updating a max
model.set('maximum', _.max([model.get('maximum'), newValue]));

//updating according to an algorithm

let cost = model.get('cost');
if (!_.isFinite(cost)) {
  //if we haven't calculated the cost yet, do so now
  cost = calculateCost(model);  
} 
model.set('cost', applyCouponCode(cost, couponCode));
```

There's nothing wrong with these examples, but I find myself writing these "get-modify-set" chunks of code a lot.  I think this experience could pretty easily be made nicer.  I'm picturing an `update` method on model that would make the above examples look like this.

```
//iterating
model.update(['totalCount', "team1Count"], count => count + 1);

//updating a max
model.update('maximum', max => _.max([max, newValue]));

//updating according to an algorithm

model.update('cost', function (cost, key, _model) {
  if (!_.isFinite(cost)) {
    //if we haven't calculated the cost yet, do so now
    cost = calculateCost(_model);  
  } 
  return applyCouponCode(cost, couponCode)
});
```

The update function would likely take 3 options: `attrs`, `transform`, `options`. 

`attrs`- an attribute name or array of attribute names to apply the transform to
`transform` - a function (defaults to _.identity) that will be passed 3 arguments `value`, `key`, `model` and will return the new value for each attribute it is run against
`options` - these would be options that would be passed through to the underlying set call that would result from the update

I'm opening this issue basically to gauge interest in having a function like this as part of Backbone core and to receive feedback.  If there is interest I'd be happy to open a PR to explore further.

Florian-R:
Like this idea. Another approach could be to allow Model#set to accept this signature, but maybe it's overcomplicating its API too much.

erosb:
+1 for this, and probably better to imlement in a separate method instead of modifying set()

akre54:
> Another approach could be to allow Model#set to accept this signature

What'd you have in mind?

akre54:
👍  on this idea. Reminds me of ImmutableJS's [`update`](https://facebook.github.io/immutable-js/docs/#/Map/update).

Florian-R:
> What'd you have in mind?

Add a new signature for set like

```js
[...]

PR #4083, opened 2016-10-05
Implementing Backbone.Model.prototype.update() (requested in #4034)

erosb:
Adding the `Backbone.Model.prototype.update()` as it was proposed in issue #4034 . I added some unittests too.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -549,9 +549,9 @@
       var attrNames = _.isArray(attrName) ? attrName : [attrName];
       var model = this;
       var newValues = _.reduce(attrNames, function(valueCollector, attr) {
-        valueCollector[attr] = transformer.call(model, this.get(attr), model);
+        valueCollector[attr] = transformer.call(model, model.get(attr), model, attr);
         return valueCollector;
-      }, {}, this);
+      }, {});
       this.set(newValues, setOptions);
       return this;
     },
--- a/backbone.js
+++ b/backbone.js
@@ -547,8 +547,9 @@
     // function which maps the current attribute value(s) to new attribute value(s)
     update: function(attrName, transformer, setOptions) {
       var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var model = this;
       var newValues = _.reduce(attrNames, function(valueCollector, attr) {
-        valueCollector[attr] = transformer.call(this, this.get(attr), this);
+        valueCollector[attr] = transformer.call(model, this.get(attr), model);
         return valueCollector;
       }, {}, this);
       this.set(newValues, setOptions);
--- a/backbone.js
+++ b/backbone.js
@@ -548,7 +548,7 @@
     update: function(attrName, transformer, setOptions) {
       var attrNames = _.isArray(attrName) ? attrName : [attrName];
       var newValues = _.reduce(attrNames, function(valueCollector, attr) {
-        valueCollector[attr] = transformer.call(this, this.get(attr));
+        valueCollector[attr] = transformer.call(this, this.get(attr), this);
         return valueCollector;
       }, {}, this);
       this.set(newValues, setOptions);
--- a/backbone.js
+++ b/backbone.js
@@ -543,15 +543,17 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
-    update: function(attrs, mapper, setOptions) {
-      if (_.isArray(attrs)) {
-        var mapperArgs = _.map(attrs, function(attr) {
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      if (_.isArray(attrName)) {
+        var mapperArgs = _.map(attrName, function(attr) {
           return this.get(attr);
         }, this);
-        this.set(mapper.apply(this, mapperArgs), setOptions);
+        this.set(transformer.apply(this, mapperArgs), setOptions);
       } else {
-        var newValue = mapper.call(this, this.get(attrs));
-        this.set(attrs, newValue, setOptions);
+        var newValue = transformer.call(this, this.get(attrName));
+        this.set(attrName, newValue, setOptions);
       }
       return this;
     },
--- a/backbone.js
+++ b/backbone.js
@@ -543,6 +543,19 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
+    update: function(attrs, mapper, setOptions) {
+      if (_.isArray(attrs)) {
+        var mapperArgs = _.map(attrs, function(attr) {
+          return this.get(attr);
+        }, this);
+        this.set(mapper.apply(this, mapperArgs), setOptions);
+      } else {
+        var newValue = mapper.call(this, this.get(attrs));
+        this.set(attrs, newValue, setOptions);
+      }
+      return this;
+    },
+
     // Clear all attributes on the model, firing `"change"`.
     clear: function(options) {
       var attrs = {};
--- a/backbone.js
+++ b/backbone.js
@@ -546,15 +546,12 @@
     // accepts an attribute name (or array of attribute names) and a transformer
     // function which maps the current attribute value(s) to new attribute value(s)
     update: function(attrName, transformer, setOptions) {
-      if (_.isArray(attrName)) {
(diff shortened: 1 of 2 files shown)


megawac:
Looks good to me

test/model.js
```diff
@@ -1445,4 +1445,40 @@
...
+        prop1: prop1 + '-changed'
+      };
+    });
+    assert.equal(model.get('prop0'), 'value0-changed');
+    assert.equal(model.get('prop1'), 'value1-changed');
+  });
+
+  QUnit.test('#4034 - Model.prototype.update() passes its options to set()', function(assert) {
+    var onChange = function() {
+      onChange.called = true;
```
megawac: can you change this to just

```js
var onChange = function() {
   assert.ok(false)
}
```

test/model.js
```diff
@@ -1445,4 +1445,40 @@
...
+    };
+    var model = new Backbone.Model({prop0: 'value0'});
+    model.on('change', onChange);
+    model.update('prop0', function(prop0) {
+      return prop0 + '-changed';
+    }, {silent: true});
+    assert.equal(onChange.called, undefined);
```
megawac: drop this assertion, perhaps change to just checking the value through a .get?

test/model.js
```diff
@@ -1445,4 +1445,40 @@
...
+    });
+    assert.strictEqual(model.get('prop1'), 'value1-changed');
+  });
+
+  QUnit.test('#4034 - Model.prototype.update() with multiple attributes', function(assert) {
+    var model = new Backbone.Model({
+      prop0: 'value0',
+      prop1: 'value1'
+    });
+    model.update(['prop0', 'prop1'], function(prop0, prop1) {
```
Florian-R: Personal taste maybe, but I find @benmccormick proposition for multiple args nicer, e.g:

```js
model.update(['totalCount', "team1Count"], count => count + 1);
```

erosb: IMO from this call it is not straightfoward to figure out which attribute will be updated.

Florian-R: Yep but I like the fact it makes transformer reusable, dumb exemple : 

```js
increment = count => count + 1

model.update('team1NumberOfFaults', increment)
model.update(['player1Score', 'team1Count'], increment)
```

erosb:
@megawac @Florian-R @benmccormick I changed the `update()` method in f635b54ae30d8ec858d87eeee5dcadab4fb0d4c5 to conform to the original proposal. It seems I slightly misunderstood the purpose of multi-attribute invocations. Thanks for explaining it.

backbone.js
```diff
@@ -543,6 +543,18 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var newValues = _.reduce(attrNames, function(valueCollector, attr) {
+        valueCollector[attr] = transformer.call(this, this.get(attr));
```
benmccormick: I think it would be a good idea to pass the model into the transformer as a 3rd argument.  That allows you to do transforms that rely on other attributes.  That probably won't be the greatest code, but it is something that people will probably want to do.

erosb: @benmccormick the model is passed as `this` , see the assertion at `test/model.js` , line 1451

benmccormick: @erosb ok.  That does limit the ability to use es6 arrow functions though, since that won't work with them (this will be in the callers scope and it will ignore the `call` binding.  Passing it as an argument is more flexible and obvious.  You could also `this` bind the model and just pass the attributes hash as the 3rd argument, similar to [Backbone.Validation](https://github.com/thedersen/backbone.validation)

erosb: Okey, I changed it in 31ef8af332b7a4082ca7691ccc3237ec973ce27a

backbone.js
```diff
@@ -543,6 +543,18 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var newValues = _.reduce(attrNames, function(valueCollector, attr) {
```
rafde: `_.reduce(attrNames, _.bind(function(valueCollector, attr) {` for lodash 4 compat?

Florian-R: Lodash compat should be achieved by the user IMHO

megawac:
Nope, we're compatible throughout the rest of backbone

backbone.js
rafde: Your expectation from lodash users it to overwrite this function just to change it to 
`var newValues = _.reduce(attrNames, _.bind(function(valueCollector, attr) {`?
What will probably happen is someone will open a PR to change it to use bind. 
I figure it'd make sense to do it now.

I think https://github.com/jashkenas/backbone/pull/4083#issuecomment-252236917 is in agreement with my suggestion.

Florian-R: I was more thinking of monkey patching the lodash reduce to mimic underscore API. I thought Lodash compat wasn't guaranteed (it's untested and the Lodash API has widely diverged), but per @megawac's comment you must be right, so nevermind.

akre54: `Function.prototype.bind` is slow, let's use the `var model = this;` pattern.

erosb: @akre54 I don't understand what do you mean by "let's use the `var model = this;` pattern"

rafde:
```
  var model = this;
  var newValues = _.reduce(attrNames, function(valueCollector, attr) {
    valueCollector[attr] = transformer.call(model, model.get(attr), model);
  });
```

rafde: @erosb ☝️

erosb: thanks @rafde , changed in 14561bf5b6660883466ad28583025eba6fc78a93

backbone.js
```diff
@@ -543,6 +543,19 @@
...
 
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var model = this;
+      var newValues = _.reduce(attrNames, function(valueCollector, attr) {
+        valueCollector[attr] = transformer.call(model, this.get(attr), model);
+        return valueCollector;
+      }, {}, this);
```
jridgewell: This `this` is unnecessary.

backbone.js
```diff
@@ -543,6 +543,19 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var model = this;
+      var newValues = _.reduce(attrNames, function(valueCollector, attr) {
+        valueCollector[attr] = transformer.call(model, this.get(attr), model);
```
jridgewell: Use `model.get(attr)`, so we can eliminate the `this` used below.

jridgewell: Let's also pass `attr` as one of the attributes.

backbone.js
erosb: Okey, all done in 820e5cc28746b39d27b0530af2c796b9d2fcccc8

erosb:
@jashkenas could you please merge this PR? Thanks

backbone.js
```diff
@@ -543,6 +543,19 @@
       return this.set(attr, void 0, _.extend({}, options, {unset: true}));
     },
 
+    // accepts an attribute name (or array of attribute names) and a transformer
+    // function which maps the current attribute value(s) to new attribute value(s)
+    update: function(attrName, transformer, setOptions) {
+      var attrNames = _.isArray(attrName) ? attrName : [attrName];
+      var model = this;
+      var newValues = _.reduce(attrNames, function(valueCollector, attr) {
+        valueCollector[attr] = transformer.call(model, model.get(attr), model, attr);
```
jridgewell: Do we think this is the right method signature?

jashkenas:
Closing, per #4034...

PR #4088, opened 2016-10-12
Add comment about combined _.defaults and _.extend

unclechu:


backbone.js
```diff
@@ -400,7 +400,11 @@
     if (options.collection) this.collection = options.collection;
     if (options.parse) attrs = this.parse(attrs, options) || {};
     var defaults = _.result(this, 'defaults');
+
+    // Additional `_.defaults()` wrap after `_.extend()` makes sense
+    // when `attrs` has a property explicitly set to `undefined`.
```
jridgewell: Let's link to https://github.com/jashkenas/backbone/issues/3842.

unclechu:
@jridgewell it's done. Is it okay that I made this revert commit? I mean if you do squash merge at the end it won't matter.

jridgewell:
No, we can squash merge it. Still need the link to #3842. And someone else should look over the wording. @captbaritone? 😉

unclechu:
@jridgewell 
```text
Still need the link to #3842
```
Do you mean I should add `#3842` to the commit message?

backbone.js
captbaritone: Rather than "makes sense" can you explain the behavior that it actually avoids?

jridgewell:
To the comment, not the commit message.

backbone.js
```diff
@@ -400,7 +400,12 @@
...
+    // Discussion: https://github.com/jashkenas/backbone/issues/3842
```
captbaritone: You can just use GitHub's issue shorthand `#1234`. You could use this as a model: https://github.com/jashkenas/backbone/blob/master/backbone.js#L1158-L1159

unclechu:
@jrburke how about now?

captbaritone:
Rather than "makes sense" can you explain the behavior that it actually avoids?

unclechu:
@captbaritone I added line about it helps to avoid conflicts with `Object.prototype` properties.

captbaritone:
I'm not familiar with this issue, and those who are reading this comment probably won't be either. In what way does it "make sense"? Does it preserve attributes which have been explicitly set to `undefined`? Does it overwrite them?

jridgewell:
It has two steps.

1. We have to create an object that contains all properties, `attrs` taking precedence over `defaults`. We clobber prototype inheritance here on purpose.
  - `_.extend({}, defaults, attrs)`
2. Anything in Step 1 that is `undefined` should be overridden with its `defaults` value. Since we clobbered prototype inheritance, we know for certain that if something is `undefined`, it was set specifically to `undefined` in `attrs`. (Or, if it was set in `defaults`, it will still be `undefined` after this step)
  - `_.defaults(_, defaults)`

jgonggrijp:
Superseded by #4267.

PR #4092, opened 2016-10-28
fix remove

sstrigler:
In particular this bug prevented shift() to work more than once, but in
general left the collection in a wrong state after a remove.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1151,7 +1151,7 @@
         var model = this.get(models[i]);
         if (!model) continue;
 
-        var index = this.indexOf(model);
+        var index = this.models.indexOf(model);
         this.models.splice(index, 1);
         this.length--;
 
--- a/test/collection.js
+++ b/test/collection.js
@@ -402,10 +402,15 @@
   });
 
   QUnit.test('shift and pop', function(assert) {
-    assert.expect(2);
+    assert.expect(7);
     var collection = new Backbone.Collection([{a: 'a'}, {b: 'b'}, {c: 'c'}]);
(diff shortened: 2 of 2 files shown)


jridgewell:
Do you have a test case? `Colleciton#indexOf` does exactly this through underscore's method.

sstrigler:
Yes, you're right. Just found out myself. The deeper problem is in fact that Collection#indexOf doesn't seem to work correctly.

```javascript
myCollection.indexOf(myCollection.at(0));
-1
myCollection.models.indexOf(myCollection.at(0));
0
```

I'll add a testcase. Then I'll check what the problem with _.indexOf is.

sstrigler:
*sigh* - only seeing this problem in firefox, chrome seems fine for instance. 

You do have tests in place that should cover that problem already, I've found. That's why I tried a different browser.

sstrigler:
So, I added tests, but they are pointless since they already work on master. I'll try to come up with a minimalistic example that can be tested from within a browser. Or is there a way to run those tests from within different environments already?

jridgewell:
What's your code is causes the error? I just tested in Firefox, and it worked perfectly.

sstrigler:
Thanks for your efforts! I've also come to the same conclusion now. The comment about Chrome working fine was a cache issue, it's not. If I just use the code as in the tests it's fine. It seems the problem is that I'm using HumanModel (which inherits from Backbone.Model) and this somehow makes 'indexOf' not work correctly. Still on it though. Thanks again. I'll get back to this once I really understood what's going on and then either suggest a solution or (more likely) just close the issue.

PR #4094, opened 2016-11-07
 Backbone examples

chadhindsight:
I thought it would be good for folks to view apps that use backbone.


akre54:
We already have an examples section that links to our TodoMVC (one of the originals IIRC) ;)

https://backbonejs.org/#examples

Issue #4098, opened 2016-11-15
model.fetch


shawnemullen:
This is not new but I thought I would re-raise it.  There needs to be a way to use model.fetch to fetch only what is on the server.  The fact it merges with what the client already has (actually this isn't really true), is a big problem.  When I do a fetch, I expect the model returned to be the model fetched.

If anyone has a work around I'd like to hear it.

elgubenis:
but thats pretty useful to keep local state moving throughout your app, at least in some of my apps thats useful, i've already seen other examples using it this way, as i understand the idea is to move models or collections of models, forwards and backwards through your views and routers (with Backbone.Radio or window(global)) you could implement a fetch method yourself which clears the model before fetching
```js
clearFetch() {
this.clear();
return this.fetch();
}
```

```md
*clearmodel.clear([options])*
Removes all attributes from the model, including the id attribute. Fires a "change" event unless silent is passed as an option.
```

keep this in mind, the idAttribute (default: id) will be removed also, so if you would like to keep that "link" just preserve it before the .clear and .set it after the .clear

alundiak:
@shawnemullen this can be closed right?

cc/ @jashkenas

jgonggrijp:
I'm assuming that the question has been answered. @shawnemullen feel free to reopen if you feel otherwise.

PR #4101, opened 2016-11-20
Implement reset option for model.set

### LINKED ISSUE #3253: Add a "reset" method to Backbone.Model
I needed to update some of a Model's data that came in from the server.  Currently, there are two options: call `Model.set`, or set `Model.attributes` directly.  I didn't want changes to be recorded, but I also couldn't use `silent` because I needed the respective views to update.  So, I wrote a monkey patch:

``` coffeescript
Backbone.Model.prototype.reset = (attributes, options) ->
    attrs = attributes || {};
    if options.parse
      attrs = this.parse(attrs, options) || {}

    @set(attrs, options);
    @changed = {};
```

Wondered why `Backbone.Model` doesn't have a reset method like `Backbone.Collection`?

blikblum:
Implements the reset option for `model.set`. Fixes #3253 #4098 

When reset flag is passed as a truthy value to model.set, the properties that exist in model.attributes but not in the attributes being set will be deleted triggering the appropriate change event, effectively resetting the model state. If reset flag is passed to `model.fetch` and the request succeeds the model state will be reset with the values returned by the ajax call.

This  feature is necessary because the primitives provided by Backbone now does not allow to get such behavior without shortcomings. 

`model.clear().set(newattrs)` will fire change events twice

`model.clear({silent:true}).set(newattrs)` will not fire change event for the unset properties

`model.clear().fetch()` suffers from the above issues plus if request fails, will end with an empty model

Also, in all the situations above, passing validate flag and validation failing, will end with an empty model

Important to note that the changes introduced by this PR are narrow and backward compatible.

Glad to iterate the changes if deems appropriate.


akre54:
This has been brought up and turned down a few times in the past. What's wrong with `model.clear(); model.set(model.defaults());`?

jridgewell:
I'm on board with it, and it's been requested multiple times. We'll probably have a weird problem with `Collection#fetch`'s `reset` option, though.

blikblum:
AFAIK is the first PR, all other were issues.

Clarification: it does not introduce a new method `reset`. It adds an option. It could be used for arbritary values other than defaults. This is how is supposed to be used: `model.set({a: 1}, {reset: true})`

It does not change behavior of current code 

> What's wrong with model.clear(); model.set(model.defaults());?

1. It will fire change event twice for attributes that are updated and will fire change once for attributes that keeps the same value

In a model with attributes {a: 1, b: 2, c: 3} 

`model.clear(); model.set({a: 1, b: 3});` 'change:a' is fired once, 'change:b' is fired twice, 'change:c' is fired once

`model.set({a: 1, b: 3}, {reset: true});` 'change:a' is not fired, 'change:b' is fired once, 'change:c' is fired once
  
2. If model.set fails due to validation we would end with an empty model

blikblum:
> We'll probably have a weird problem with Collection#fetch's reset option, though.

Can you elaborate the problem?

It does not change behavior of existing code. The change is restricted to set method and does not touch in fetch code at all. 

It will work for fetch also because the reset option is propagated to model.set that is only called inside fetch after a sucessfull request

emileber:
I like the `reset` option as well but I don't think using `delete` on each key is the way to go. 

Would just replacing the `attributes` hash with a new object be more efficient?

prantlf:
How about using the name `replace` instead of `reset`?

`reset` has a different meaning when fetching a collection. If you wanted to re-fetch a collection with resetting the models using your options, `fetch({merge: true, reset: true})` would reset the collection instead of merging. `fetch({merge: true, replace: true})` would work as you expected.

ksurakka:
I think this PR is still valid.

The suggested workaround `model.clear(); model.set(model.defaults());`, by @akre54, doesn't work correctly because of either duplicate or non-existence change events. Those cases are described here: https://github.com/jashkenas/backbone/issues/3253#issuecomment-135461345

And if this really is a duplicate of some other item, I think that other item should be referenced.

PR #4106, opened 2016-12-03
Fix typo in index.html

piperchester:


captbaritone:
http://english.stackexchange.com/questions/56435/an-other-vs-another

reustle:
I agree with the PR change. I can't tell if @captbaritone posted the link in support of your change, or against it 😄

captbaritone:
Sorry about that. I originally posted it in favor of the change, but on closer inspection, it looks like the choice of "an __other__" was probably made so that "other", the name of the variable used in the method signature, could be bolded.

We should either close this pull request or change the name of the variable in the signature to "another" (which sounds a little awkward to me).

akre54:
Yup, let's leave as is. Technically you're telling the object you want to listen on *the* other, so we could maybe change the article from indefinite to definite.

PR #4107, opened 2016-12-05
 Fixed Korean character crash bug on Safari Browser

stegano:
Fixed Korean character crash bug on Safari Browser
Add Korean character test spec
THX :)

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1720,7 +1720,10 @@
     // decoded for comparison. `%25` should not be decoded since it may be part
     // of an encoded parameter.
     decodeFragment: function(fragment) {
-      return decodeURI(fragment.replace(/%25/g, '%2525'));
+	    fragment = decodeURI(fragment.replace(/%25/g, '%2525'));
+	    // If contains Korean then encode hashString - Korean character crash bug on Safari Browser :(
+	    fragment = /[ㄱ-힣]/.test(fragment) ? encodeURIComponent(fragment) : fragment;
+	    return fragment;
     },
 
     // In IE6, the hash fragment and search params are incorrect if the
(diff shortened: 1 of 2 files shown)


backbone.js
```diff
@@ -1720,7 +1720,10 @@
     // decoded for comparison. `%25` should not be decoded since it may be part
     // of an encoded parameter.
     decodeFragment: function(fragment) {
-      return decodeURI(fragment.replace(/%25/g, '%2525'));
+      fragment = decodeURI(fragment.replace(/%25/g, '%2525'));
+      // If contains Korean then encode hashString - Korean character crash bug on Safari Browser :(
+      fragment = /[ㄱ-힣]/.test(fragment) ? encodeURIComponent(fragment) : fragment;
```
jdalton: Has this been [reported to Safari](https://bugs.webkit.org/query.cgi?format=specific&product=WebKit)?

stegano: Hi @jdalton
Thank for your reply :)
I saw your replay and reported it
https://bugs.webkit.org/show_bug.cgi?id=165597

stegano: This issue appears to have been fixed in a future version of Safari, but it will not work with older (and current) Safari browsers :(

stegano: @jdalton
I hope that this feature will approved because almost all users are still uses the old browser version :)

ogonkov:
Harcoding some language-specific characters is looks weird.

stegano:
@ogonkov  Thank you for your good feedback. Let's find another way :D

ogonkov:
I've got similar problem with broken encoding of Russian, and ended with patching `decodeFragment` on my side

PR #4131, opened 2017-04-04
Use the public `on` method when listening

jbboehr:
I've merged master into #3615. Any reason it was never merged before?


jbboehr:
I'm gluing together a sockjs client and a cross-tab communication library, but without this patch, `listenTo` won't unregister the events because it mutates the object instead of calling the `off` method which I've hooked into, see https://static.vid.me/js/bus.js

PR #4136, opened 2017-04-25
Allow multiple attributes to be unset at once.

jrparish:


akre54:
Let's keep the API simple. You can always add this as a plugin.

PR #4137, opened 2017-04-28
Clarifying improvements to router documentation

captbaritone:
Given that route wildcards have names, it is easy to expect they might be passed to the action function as key/value pairs.

Additionally, the definition of an "action" is quite vague, and could easily be confused with an event or some other construct. I've attempted a minor improvement here.


captbaritone:
@akre54 I'm guessing this needs to get cherry-picked. Would you like me to open a second PR against `gh-pages`?

akre54:
Nah I'll grab it.

PR #4140, opened 2017-05-26
fix Collection.model breaks w/ shorthand function

wawjr3d:
({ bar() {} }).bar.prototype === undefined in Node, Firefox, Chrome

while

({ bar: function() {} }).bar.prototype is an object

In the shorthand function case, Collection.prototype.modelId throws


wawjr3d:
looks like the travis build runs on node 10 so i won't be able to add tests for this case. let me know what to do for testing

PR #4148, opened 2017-06-23
Collections: Add => Check id against the default specified instead of just null

RobinHerbots:
Hi, 

Say you have a collection and add some new models to it.  If you had specified some default for empty id other then null it will fail.  So it would be better to check against the default value instead of only considering null.

```
 var myFancyModel = Backbone.Model.extend({
       defaults: {
           id: -1,
           roleId: "",
           claimType: "",
           claimValue: ""
       }
   });
var myFancyCollection = Backbone.Collection.extend({
    model: myFancyModel ,
    initialize: function (models, options) {
      ...
    }
});

var test = new myFancyCollection();
//add 2 new models to the collection
test.add(new  myFancyModel ());  
test.add(new  myFancyModel ()); //does not insert a new model


```
Validating against the default solves this, see PR.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1178,7 +1178,7 @@
     _addReference: function(model, options) {
       this._byId[model.cid] = model;
       var id = this.modelId(model.attributes),
-      did = this.modelId(model.defaults) || null;
+      did = this.modelId(model.defaults);
       if (id != did) this._byId[id] = model;
       model.on('all', this._onModelEvent, this);
     },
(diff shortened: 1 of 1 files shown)


paulfalgout:
You can't have multiple of the same ids within a collection.  I don't think this has to do with the defaults.  It you added a 3rd `add` to your example it wouldn't add anything either.  I'd also question the value of any model needing a default id.

RobinHerbots:
@paulfalgout , 

The default id represents the "new" value on the backend.  So adding models with the default id are treated as all new ones.

paulfalgout:
maybe

jgonggrijp:
I agree with Paul. The `idAttribute` is intended to be unique and therefore shouldn't have a default. If the backend needs it, adding it in `toJSON` is a good solution.

Closing, but comments still welcome.

PR #4162, opened 2017-09-27
Add CDN links

LukasDrgon:
I added a [jsDelivr CDN links](https://www.jsdelivr.com/package/npm/backbone) to your site so that people can use files directly without downloading them. jsDelivr is also the [fastest opensource CDN](https://www.cdnperf.com/) available and built specifically for production usage.


jashkenas:
We don't encourage third-party CDN usage. 

You're certainly welcome to use 'em if you want to, but the main docs shouldn't link to them.

PR #4167, opened 2017-10-29
Eliminate eslint warnings

ambisoft:
Removed unnecessary quotes and brackets to eliminate 12 eslint warnings (reported when running npm test)


jashkenas:
No thanks -- those parentheses are there for clarity, and the code (for example, in the first example), reads better with them in.

PR #4174, opened 2017-12-13
Fixes #4097, add parse option to docs section for Backbone.Collection

### LINKED ISSUE #4097: Collection Docs: Missing note about parse option
On the [Model docs](http://backbonejs.org/#Model), there is this note:

> If `{parse: true}` is passed as an option, the attributes will first be converted by parse before being set on the model.

This appears to be true on [Collections](http://backbonejs.org/#Collection) as well, but is not documented.

avdeev:


jgonggrijp:
A bit late, but thanks for your contribution!

Issue #4085, opened 2016-10-06
Backbone navigate - url encode in different language with diacritics


adyz:
I have a search form that redirects the user to `search/thetermusersearched` where `thetermusersearched` is the exact value he entered. All this with the navigate function provided by Backbone.

This is all fine when I use strings in English (masa de calcat), but when I add diacritics in the input (masă de călcat) I get the route function fired twice.

The problem I have is in Firefox and Safari (the later in Mac and iOS)

I tried using `encodeURI` and `encodeURIComponent` when I use the `navigate`, but no success.

---
HTML

```
<div id="view-goes-here">
    <a href="#" data-string="masa de calcat">One alert</a>
    <a href="#" data-string="masă de călcat">Two alerts</a>
</div>
```

---
JS

```
    var R = Backbone.Router.extend({
        routes: {
            'results/:query': 'results'
        },
        results: function(query) {
            alert('Route triggered: ' + decodeURIComponent(query));
        }
    });
    var myR = new R;
    Backbone.history.start();

    $(function(){
      $('a').on('click', function(e){
       e.preventDefault();
       var href = $(this).data('string');
       href = 'results/' + encodeURIComponent(href);
       console.log(href);
       myR.navigate(href, {trigger: true});
      })
    })
```

---
See fiddle here: http://jsfiddle.net/adyz/qcged76e/4/

Any thoughts on this?

astronautas:
On Ubuntu Chrome, whenever I click on either English string or the one with diacritics, I still get single alerts for both. Maybe the problem is not with backbone at all but your local stack?

gwynjudd:
Hello,

I have some reports that we are also seeing this issue. I did some investigation into this. I could not replicate it using the jsfiddle in chrome/windows. On Safari on the Mac, I did see the issue.

At the root, the problem seems to be that in webkit (replicates issue), the location hash is URL encoded, but in chrome (does not replicate), the URL hash is not encoded. I will attach screenshots showing the behaviour and how it causes the route to fire twice.

Firstly, the route function fires once due to the jsfiddle code calling route.navigate, which calls loadURL. This is expected behaviour and happens on all platforms so I won't mention it again:

![screen shot 2016-11-28 at 9 37 33 am](https://cloud.githubusercontent.com/assets/314188/20651548/78f0e8be-b54c-11e6-9f86-3a958e48d910.png)

Following the handling of this event, a hash change event fires. This causes the checkUrl method to fire:

[...]

PR #4175, opened 2017-12-14
Fix navigate with replace: true and diacriticals/unicode characters in the hash

gwynjudd:
It seems there is a change in chrome 63 (also similarly
introduced a while ago in some earlier version of iOS Safari) where the
hash portion of the location.href is encoded, where it used to be
decoded

Fixes https://github.com/jashkenas/backbone/issues/4085

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1763,7 +1763,7 @@
     // Checks the current URL to see if it has changed, and if it has,
     // calls `loadUrl`, normalizing across the hidden iframe.
     checkUrl: function(e) {
-      var current = this.getFragment();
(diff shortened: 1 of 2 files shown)


gwynjudd:
Anyone reading this, I couldn't actually get the added unit tests to "successfully fail". The reason being, it seems that running in karma-chrome-launcher, the 'hashchange' event never actually fires so the checkUrl method isn't called. Despite this, I added the tests in the hope that somebody could figure that out. I can verify that the change does actually work by manually testing it in various contexts.

Issue #1019, opened 2012-02-17
Model.parse needs a compliment: Model.serialize


BenLambert:
So Backbone has a great function for parsing data from preexisting APIs into something more suitable for the Backbone model.

But it is only half of the equation. We also need to serialize the data back into the format required by the server.

Right now if you are making use of Model.parse, you basically also have to write your own custom `sync` method. That's not entirely bad, but it could be much simpler if Backbone included a `serialize` method on the model that was called by the default Backbone.sync method.

Basically it would just mean changing this line in Backbone.sync
 `params.data = JSON.stringify(model.toJSON());`
to this:
 `params.data = JSON.stringify(model.serialize.call(model, model.toJSON()));`
and then updating the Backbone.Model to include a new no-op `serialize` method just like `parse`:

```
parse: function(resp, xhr) {
  return resp;
},
serialize: function(attrs) {
  return attrs;
},
```

PR #4179, opened 2017-12-28
Call model.toJSON always when sync model

AriasBros:
When you save your model with patch option to true, toJSON isn't called. This behavior goes against the issue #1019 where it tells that:

> toJSON is already the complement of parse

With this PR the default behavior is the same as the current one, but it gives us the opportunity to manipulate the attributes before send them to the persistent layer, either using a CREATE, PATCH or PUT method.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -437,7 +437,8 @@
 
     // Return a copy of the model's `attributes` object.
     toJSON: function(options) {
-      return _.clone(this.attributes);
+        options || (options = {});
+        return options.attrs ? options.attrs : _.clone(this.attributes);
     },
 
     // Proxy `Backbone.sync` by default -- but override this if you need
@@ -1573,7 +1574,7 @@
     // Ensure that we have the appropriate request data.
     if (options.data == null && model && (method === 'create' || method === 'update' || method === 'patch')) {
       params.contentType = 'application/json';
-      params.data = JSON.stringify(options.attrs || model.toJSON(options));
+      params.data = JSON.stringify(model.toJSON(options));
     }
 
     // For older servers, emulate JSON by encoding the request into an HTML-form.
--- a/test/model.js
+++ b/test/model.js
@@ -1447,6 +1447,19 @@
     model.save({x: 1}, {wait: true});
   });
 
+  QUnit.test('toJSON is called during save(..., {patch: true})', function(assert) {
+    assert.expect(1);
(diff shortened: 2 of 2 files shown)


ogonkov:
If you have to manipulate attributes before sending, why not pass already manipulated object in `attrs`?

AriasBros:
> If you have to manipulate attributes before sending, why not pass already manipulated object in `attrs`?

Because the object `attrs` should store data in the format that is consumed by your application, not in the format of your persistence layer.

And, if `toJSON` is the complement of `parse`, it should called always before send data to the persistence layer, in the same way that `parse` is always called when data is received from the persistence layer.

For example, a model with the next `parse` implementation:

    parse(response, options) {
        response.rating = response.rating * 2;
        return response;
    }

In the persistence layer, rating is stored as an integer from 0 to 5. But in our app we use rating as a number from 0 to 10.

Now, we want update the rating in our model:

    model.save({ rating: 7 }, { patch: true });

Currently, with this line, our model will be updated to rating 7. And the persistence layer also will receive a 7 as rating, which is wrong. This is failing also in the premise that `toJSON` is the complement of `parse`, because it is not being called.

You are suggesting to do:

    model.save({ rating: 7 / 2 }, { patch: true });

Which is also wrong, because yes, your persistence layer will receive a rating of 3.5, that is the expected value. But your model will have a rating of 3.5 too, and that value is not what you want.

With this PR, in the same way that we have a `parse` method in our model, we will have the `toJSON` method too:

    toJSON(options) {
        let attributes = Backbone.Model.prototype.toJSON.apply(this, arguments);
        attributes.rating = attributes.rating / 2;

        return attributes;
    }

ogonkov:
Very neat. Now it make a sense for me. Thank you for a great explanation.

ogonkov:
It is great PR, but i think we should add some flag, to switch for new behaviour, because it is pretty much could break someones apps

```
Backbone.useToJSON = false;
```

jgonggrijp:
@ogonkov I agree with you, but I think I want this behavior to be the default in a future Backbone 2.0. How about we name the flag `Backbone.future` instead of `Backbone.useToJSON`, and then potentially also use that flag for other behaviors that we want to change in the future?

ogonkov:
One flag for few features? @jgonggrijp

jgonggrijp:
Well, *changed* features. That we intend to become standard.

Issue #4185, opened 2018-02-21
Model 'change' event is not triggered after Exception.


vinaymor002:
**Steps to Reproduce**
Add a 'change' event handler for a Backbone Model instance.
Make a change to model attribute, change event handler is called and some exception occurs in the handler.
Again, make a change to model attribute. No 'change' event handlers are called.

Code to demonstrate this.
```javascript
var model = new Backbone.Model();

var changeHandler = function() {
    console.log("name changed to: " + model.get('name'));
};
Backbone.Events.listenTo(model, 'change', changeHandler);

model.set('name', 'sachin');
// prints "name changed to: sachin" on console
model.set('name', 'virat');
// prints "name changed to: virat" on console

var model2 = new Backbone.Model();

var changeHandler = function() {
    console.log("name changed to: " + model2.get('name'));
    a.log("model changed"); // will throw an exception. a is not defined.
};
Backbone.Events.listenTo(model2, 'change', changeHandler);

model2.set('name', 'Sachin');
// prints "name changed to: Sachin" on console. Also exception occurs inside change handler

model2.set('name', 'Virat');
// nothing is printed. model change handler is not called.
model2.set('name', 'Rohit');
// nothing is printed. model change handler is not called.
```

**Analysis:**
Inside Model set function, this.trigger is called and if Exception is thrown in handlers, this._changing is never set to false and on subsequent model changes, 'change' event will not be triggered. 
```javascript
if (changing) return this;
            if (!silent) {
                while (this._pending) {
                    this._pending = false;
                    this.trigger('change', this, options);
                }
            }
            this._pending = false;
            this._changing = false;
            return this;
```
A developer's code should never break behaviour of a framwork. Smilarly subsequent changes to model should trigger 'change' event on model. 

**Solution**
Should wrap trigger logic around try/finally block ensuring this._changing and this._pending are set to false.

jgonggrijp:
Closing this as event handling is not exception-safe anywhere in Backbone. Addressing that in a consistent way would likely incur a significant performance penalty. See #4186.

PR #4186, opened 2018-02-21
#4185 Model 'change' event is not triggered after Exception.

vinaymor002:
PR for #4185


jgonggrijp:
Good point! I'm now strongly leaning towards slapping a "wontfix" on this, but I'll leave it open for a day or so in case somebody else wants to reply.

GammaGames:
I agree. Exceptions in the handler should ideally be handled and prevented by the dev, and consistent handler behavior is more useful overall than trying to take into account the edge cases this would have + applying them to the other handler functions in the library.

PR #4196, opened 2018-05-16
Es2015plus

unquietwiki:
After my messy attempt to contribute some fixes ~2 weeks ago, I went back over what I had on my end. Since I'm trying to use Backbone with an ES2015+ project; and the assorted online guides about trying such date back 3+ years with mixed results; I figured I'd try to make a conversion effort. ESLint & Deep Code were particularly helpful. I figured you all could do something with this on your own time; hence the separate branch. Thanks.


noobiek:
Hi,
Is there a reason you decided to go this road and not take (and possibly improve) this https://github.com/typhonjs-backbone/backbone-es6 ?

unquietwiki:
I didn't even see that come up in my searches on the topic before. Thanks
for bringing it to my attention.

That particular effort looks moribund ATM, and I was able to spin out this
effort based on reference materials & heavy code linting. I also wanted it
to be as close to the original as possible: I don't have good unit testing
of my own, but was able to try fixing up the testing in this batch.

typhonrt:
Moribund.... Really? ;P Uh.. I don't do anything w/ it until this repo actually decides to pull the trigger on a release, so `1.3.3` compatible. If `1.4` is actually released I'd update `backbone-es6`. ;P Please release! ;P

unquietwiki:
@typhonrt @noobiek I've got some personal stuff going on the next few days, so I won't have time to compare my yeoman effort here; against what you guys came up with. I'd be willing to put in some effort to help out whomever wants the help: I just want https://github.com/unquietwiki/floorplan to work right; I'm having trouble passing a Backbone.Collection from the browser into a series of ES2015+ classes. You might also want to look at what linters I set up to get this far. Thanks!

noobiek:
So far I did not have any problems using BB in ES6 webpack builds.
You could've taken something like this - https://github.com/sabarasaba/modern-backbone-starterkit
Then you just - import Backbone from 'backbone' - into your components, whatever.
Or did you miss that one too?

If your problems were with something else - can you be more specific where you had problems and what these problems were?

noobiek:
@typhonrt why not move forward, branching off BB to something else? Give BB a second life perhaps, and integrate bb.stickit into it (they are looking for maintainers, btw). I like the way bb.stickit handles views updates so much!

amenadiel:
Hey I've heard you were discussing BB as es6 modules so I came to throw my version into the mix


https://github.com/HuasoFoundries/backbone_es6/

It's just old method extract refactoring technique, along with es6 named imports and finally rollup.js to bundle it out. I make sure my build passes the official repo tests.

blikblum:
@amenadiel Do you tested to see if tree shaking worked?

We did the same with Marionette but tree shaking nor uglify dead code elimination worked even using ES modules. Tested with webpack 4x

I found that extending the prototype after the class constructor avoided tree shaking or dead code elimination

The only way to remove unused class from the bundle would be using ES6 class or wrapping all definition in a iif with a  /*#__PURE__*/ annotation. 
Related: 
https://github.com/babel/babel/issues/5632
https://github.com/mishoo/UglifyJS2/issues/1261

amenadiel:
@blikblum tree shaking works for some components, like View for example. (module is 11Kb vs 73Kb of the full Backbone module)
In other cases, like Collection, it depends on Model, so you can't really import collection by itself.

unquietwiki:
Hey everyone. Sorry for the belated followup: got married, and also been recovering from a cold.

(ponders)

@typhonrt Your attempt at this is a lot in line with what I was trying to do. However, I'm unclear how to actually try to use it; let alone help clean that up. I have attached the linting rules I used for mine; as for building, the process needs refinement.
[eslintrc.txt](https://github.com/jashkenas/backbone/files/2073970/eslintrc.txt)

@amenadiel I like that yours is on NPM, and has a bit more current documentation on it. However, when you say you can't use classes with it; that would seem to defeat what Typhon and I want to do?

@noobiek Aside from what we've posted here thus far, I also found https://github.com/maratfakhreev/backbone-es6-promise from @maratfakhreev , and https://github.com/lukasolson/backbone-super from @lukasolson . It would seem from what I wanted to do; and what everyone else has tried to do already; it seems like there has to be some kind of solution here? The stumbling blocks I am aware of overall include proper super/this inheritance, proper use of arguments, and seeing if any of this works in the latest browsers without being Babelized (which would help determine how to make that work).

Some related reading material I found trying to understand this stuff.
- https://zendev.com/2018/05/09/understanding-spread-operator-in-javascript.html
- https://medium.freecodecamp.org/here-are-examples-of-everything-new-in-ecmascript-2016-2017-and-2018-d52fa3b5a70e

unquietwiki:
https://github.com/unquietwiki/floorplan I have my latest code updated for the project I'm working on; did set it to use **backbone_es6** from @amenadiel . My Backbone classes in the browser still can't properly consume collections for some reason. :-/

amenadiel:
@unquietwiki currently the common way is to do

```js
var messageModel = Backbone.Model.extend({
   ...
});
```

So... you'd like to be able to do

```js
class MyModel extends Backbone.Model {
  ...
}
```

It doesn't sound like a big deal. If we rewrite Model, View, Controller to become classes, then extending them would be transparent.

However, to keep the compatibility with current projects, `extend` should be kept as a static method with takes me to the bottomline: I have yet to learn to implement a static method capable to perform prototypical inheritance from a class to a plain object.

unquietwiki:
https://github.com/unquietwiki/floorplan/blob/master/references.md Check
there: I dug up a number of updated JS references that might help.

typhonrt:
How the classical extend is handled in my effort which is separated out in: 
https://github.com/typhonjs-backbone/backbone-es6/blob/master/src/extend.js

and the module runtime entry point: 
https://github.com/typhonjs-backbone/backbone-es6/blob/master/src/ModuleRuntime.js#L36-L37

PR #4199, opened 2018-06-29
Adds support for non-delegated events

shishirkulkarni:
This pull request adds support for specifying bindings for non-delegatable events in the `events` hash. This removes the burden from the developer to manually maintain event bindings for non-delegatable events like `scroll` which is a common usecase.
The way you can specify a non-delegatable event is by prefixing the event name with a `*`. For example:
```
events: {
        "*scroll .my-list": "listScrolled"
}
```
This pull request addresses #4200

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1301,7 +1301,7 @@
   };
 
   // Cached regex to split keys for `delegate`.
-  var delegateEventSplitter = /^(\S+)\s*(.*)$/;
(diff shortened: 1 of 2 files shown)


ogonkov:
Same could be achieved via subview, no need to add "magic" where it's not required.

jgonggrijp:
I agree with the above comment and will close this for the same reason as #4200.

PR #4202, opened 2018-08-11
Removed checks for $

w0298i:
Removed checks for 'root.$', which is shorthand for selecting elements in newer browsers.


w0298i:
Closed this because nothing came back, and it's filling up my feed.

Issue #4224, opened 2019-10-08
TypeError: existing.set is not a function


PeterW570:
```
const collection = new Backbone.Collection();
const model = new Backbone.Model({
    id: 'hasOwnProperty'
});
collection.set([model]); // TypeError: existing.set is not a function
```

In `Collection#set`, it does: `var existing = this.get(model);` which in this case is setting `existing` to `Object#hasOwnProperty` rather than `undefined` because `Collection#get` is returning the result of`this._byId['hasOwnProperty']`

Have reproduced this bug in a [codepen here](https://codepen.io/PeterW570/full/mddyaOx).

paulfalgout:
This seems like an odd thing to fix.  This to me seems unlikely in general, but fixing it would add at least some cost to most actions across backbone where this issue is never likely to arise.  If anything the best fix to me seems to be something like adding to the docs "`id`s cannot be members of `Array.prototype.String`"

PeterW570:
Can also cause the same bug by setting it to a different attribute and then setting idAttribute of the model to that, e.g:

```
  const keyedModel = Backbone.Model.extend({ idAttribute: 'key' });
  const keyedModelCollection = Backbone.Collection.extend({ model: keyedModel });
  const collection = new keyedModelCollection();
  const model = new keyedModel({ key: 'hasOwnProperty' });
  collection.set([model]); // TypeError: existing.set is not a function
```

Agree it's an edge case, but I think it's a valid use case if the string is a valid unique identifier in the database

jgonggrijp:
As far as I'm concerned, this isn't specific to Backbone; it's on the list of things you should do nowhere in JavaScript. Of course, it's fine if you learn by experimentation that doing stuff like this breaks things, but this doesn't need documentation in Backbone, let alone a fix.

As for unique identifiers in databases, I expect those to be serial numbers or something more advanced like a UUID. Certainly not names of JavaScript's global object methods.

Issue #3623, opened 2015-05-22
Why remove the root's trailing slash when wantPustState on IE8?


houkanshan:
Example:

``` javascript
Backbone.history.start({ pushState: true, root: '/root/' })
```
1. The browser is IE8 (no pushState)
2. Our server will force a trailing slash, a request of `/root` will receive a redirection to `/root/`
3. Visit `/root/aa/bb`

Due to this commit d550b8b , an replaced url will be `/root#aa/bb`, then the server returns a redirection to `/root/`, which lead to an additional request, and at the same time the hash will be lost. The `/root/aa/bb` can't be reached.

I've read the #2656 and #3443. I'm thinking that should the tailing slash here also be optional ? Which means:

``` javascript
Backbone.history.start({ pushState: true, root: '/root/' }) // /root/#aa/bb
Backbone.history.start({ pushState: true, root: '/root' }) // /root#aa/bb
```

blzzz:
+1

paulfalgout:
ping @akre54  maybe this can be closed due to https://github.com/jashkenas/backbone/pull/4008

PR #4225, opened 2019-10-09
avoid Object.prototype inheritance for col._byId

BROOSWAJNE:
fixes issue #4224

`collection._byId` is an internal object used solely for mapping ids to models, and doesn't need any of the standard object prototype methods for its internal usage.  
Unlikely to be a breaking change for any external uses, unless someone is really hacking into the core backbone functionality, and they can simply do things like `Object.prototype.hasOwnProperty.apply(collection._byId)`.

Alternative solution could be to use a `Map()` object for `collection._byId`, but that would require many more code changes, would not be supported in old browsers, and is more likely to break external uses.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1122,7 +1122,7 @@
     _reset: function() {
       this.length = 0;
       this.models = [];
-      this._byId  = {};
+      this._byId  = Object.create(null);
     },
 
     // Prepare a hash of attributes (or other model) to be added to this
--- a/test/collection.js
+++ b/test/collection.js
@@ -101,6 +101,26 @@
     assert.equal(collection2.get(model.clone()), collection2.first());
   });
 
+  QUnit.test('#4224 - internal map this._byId does not inherit from Object.prototype', function(assert) {
+    assert.expect(4);
+
+    var col1 = new Backbone.Collection();
+    assert.equal(col1._byId.constructor, undefined, 'this._byId has a constructor');
+    assert.equal(col1._byId.hasOwnProperty, undefined, 'object prototype methods are present on this._byId');
+
+    var model = new Backbone.Model({id: 'hasOwnProperty'});
+    // will throw if regression
+    col1.set([model]);
+    assert.deepEqual(col1.models, [model], 'models not correctly added');
+
+    var KeyedModel = Backbone.Model.extend({idAttribute: 'key'});
+    var KeyedCollection = Backbone.Collection.extend({model: KeyedModel});
+    var col2 = new KeyedCollection();
+    // will throw if regression
+    col2.add({key: 'hasOwnProperty'});
+    assert.deepEqual(col2.models[0].attributes, {key: 'hasOwnProperty'}, 'models not correctly added');
+  });
+
   QUnit.test('has', function(assert) {
     assert.expect(15);
     assert.ok(col.has(a));


jgonggrijp:
I just closed #4224 as invalid, but I'll give this PR some thought. I don't think supporting names of JavaScript's `Object` prototype properties as identifiers is important, but *maybe* this could make Backbone less sensitive to prototype pollution. *Maybe*. This has to be weighed against the obvious cost of requiring `Object.create`, which is almost certainly a breaking change.

ogonkov:
I think #4224 was valid. Backbone should check for hasOwnProperty, when accessing model by id that is same as some builtin method. It could be possible attack vector aswell, if we write to object without checks. Like model with id `__proto__`.

jgonggrijp:
I closed #4224 because it was motivated by a use case that I believe shouldn't be supported. However, I left the current PR open because I think it might improve security. So we don't really disagree on that count.

You really have to ask yourself, though: how does the id of a model end up being something like `__proto__` or `hasOwnProperty`?

ogonkov:
I think we should ask "do we handle it correctly _when_ it happen"

ogonkov:
Lets imagine that user app is somehow wanted `__proto__` as model id. What would be correct behavior in that case? I think it would be correct to return model with such id.

jgonggrijp:
Backbone is not obliged to support whatever the user wants to do. We can set preconditions, such as "choose sensible values as model ids". `__proto__` is not a sensible value for a model id in my book.

I'm more interested in scenarios where the model id ends up being `__proto__` or something like that *unintentionally*, i.e., when an attacker it trying to hack a Backbone application. I can think of two ways this might happen:

1. The app developer might set the model id based on user input. This is always a bad idea, so the app developer just shouldn't do that.
2. The database backend has already been compromised by an attacker and they inserted `__proto__` as a model id somewhere. While exotic, I currently cannot rule out that this might grant the attacker additional capabilities. This hypothetical scenario is the reason I've left this PR open. My question is basically this: *would* this grant the attacker any new capabilities, or would it only break things?

It *is* a breaking change because it requires `Object.create`, which is not backwards compatible with all environments currently supported by Backbone (such as Internet Explorer).

ogonkov:
One of my project was once pwned with `__proto__` poisoning
I had "wonderful" 4 hours of debugging trying to figure out how this works

ogonkov:
`Object.create` supported since ie 9. I see that it is a minimal version in tests.

Where is IE support declared?

ogonkov:
Seems like IE8 support was dropped since #4008

jgonggrijp:
> One of my project was once pwned with `__proto__` poisoning
> I had "wonderful" 4 hours of debugging trying to figure out how this works

Yes, that's the potential danger I'm interested in. I'm not sure it applies here, though, because prototype poisoning depends on assigning properties of `__proto__` rather than `__proto__` itself. This most commonly happens with recursive assignment functions like Lodash's `merge` and `defaultsDeep`. If somebody can provide a proof-of-concept (PoC) to show that it can actually be exploited in this case, that would be really helpful. That would also give us something to base a useful regression test on.

> `Object.create` supported since ie 9. I see that it is a minimal version in tests.
> 
> Where is IE support declared?

Oops, I stand corrected. I was looking at https://caniuse.com/?search=object.create, not realizing that the top of the page is showing me `Object.entries` instead of `Object.create`. Strange order of search results...

Although #4008, which you linked to, also mentions workarounds for old IE that haven't been removed yet. Which means that IE8 probably still works, although it is untested. The situation is the same for Underscore, currently. That means it's still a breaking change.

jgonggrijp:
Of course support for IE8 is eventually going to be dropped entirely, because it's extremely outdated. Still, that's something you do on a major version upgrade, not within the 1.x series. Note that @akre54 wrote "future" and "eventually", not "whenever we feel like it".

I'm fine with not fixing #3623, which happened accidentally, and with dropping IE8 support in tests, as has happened in #4008 (I did that in Underscore, too, simply because SauceLabs didn't support IE8 anymore), but technically it's still a breaking change.

Again, if we can establish that there is *actually* a security risk here, *then* I might be willing to make this breaking change already in the 1.x series. After all, then it is a bugfix, and bugfixes are always at least somewhat breaking changes by necessity.

paulfalgout:
Personally I think IE8 should be dropped as a concern without worry about it affecting a major release.  Backbone has not historically been as concerned with this, but I strongly support introducing breaking changes in major versions.  However I think in this case dropping IE8 is forgivable.  It's a unique situation for a breaking change and any cost of supporting IE8 is not worth the effort at this point.  Particularly for a smaller community with no real reasonable way of actually testing any breaking changes on the platform.

jgonggrijp:
I think we mostly agree. I would like to argue, however, that as long as this change is not provably urgent (I'm sceptical that it actually solves a problem), there is no real cost in postponing the merge until the next major release. In other words, saving a questionable bugfix for later is more forgivable than making a certain breaking change now.

Issue #4159, opened 2017-09-22
Collection does not purge internal `_byId` reference to a removed model


cueedee:
Please find below about the tersest reproduction I could fathom of a sequence of steps in a Backbone application that leads to a `Collection` getting into a state of internal confusion. 

A working copy can be found in this [JSFiddle](https://jsfiddle.net/cueedee/ph9uo43w/) .

What it demonstrates is a `collection` keeping an internal (`_byId`) reference to a `remove()`-ed `Model`.  That `Model` is no longer part of `collection.models`, however, it can still be retrieved through its (former) `id`. 

```js
var
    MyView = Backbone.View.extend( {

        initialize: function () {

            this.model = new Backbone.Model( {

                id:  'foo'
            ,   foo: 'bar'
            } );

            this.collection = new Backbone.Collection( this.model );

            this.listenTo( this.model, 'change:id', function ( model, id ) {

                if ( id == null ) {
                    this.collection.remove( model );
                }

            } );

            this.model.unset( 'id' );

            $('#models' ).val( JSON.stringify( this.collection.models      ));
            $('#_byId'  ).val( JSON.stringify( this.collection._byId       ));
            $('#get-foo').val( JSON.stringify( this.collection.get( 'foo' )));
        }

    })

, foo = new MyView()
;
```

TM-Olashyn-Dmytro:
Hmm, interesting. When we add model to collection, __byId_ add two properties: model.id and model.cid, their reference to model. Issue on 1134 line, **id === undefined**, but __byId_ contains **model.id** - _'foo'_.

Collection don't has some special logic when change model.id to **value == null**, check line 1183.

Also you can change **unset** to **set** _null_ or _undefined_. **this.model.set('id', null);**

P.S.: I think you never need set your model id to null or undefined, it is anti pattern.

![developer tools - httplocalhost9696](https://user-images.githubusercontent.com/30149027/30748465-0f00c462-9fb9-11e7-8af5-d4447fa46e5e.jpg)

![backbone_model_event](https://user-images.githubusercontent.com/30149027/30748519-3c2bde36-9fb9-11e7-932c-34babac994ea.jpg)

jashkenas:
Yes — you should never do this, but it's still a bug.

We should fix this by making sure that if an `id` is changing from a value to `null`, we `delete this._byId[id]` first.

rgroothuijsen:
[...]

PR #4227, opened 2019-11-05
Remove model from collection when model ID is unset

rgroothuijsen:
This PR fixes issue #4159, where a model was not properly removed from a collection after its ID was unset. The cause for this is that `_onModelEvent` handles potential model deletions upon detecting an event called "change", when the actual event name being sent was "change:_(attribute)_". An extra check was added so that the latter case is handled as well.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -769,6 +769,9 @@
   var setOptions = {add: true, remove: true, merge: true};
   var addOptions = {add: true, remove: false};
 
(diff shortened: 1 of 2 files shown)


ogonkov:
This is a breaking change, and logic 'no id = no place in collection' is seems broken to me generally.

jgonggrijp:
Obsolete because of #4249.

Issue #4203, opened 2018-08-21
Error: "Cannot read property 'idAttribute' of undefined" in collection


avasuro:
### Description

If I create new collection class, that extends Backbone.Collection, and define it's "model" property as a function which returns instantiated model - then on collection constructed error "Cannot read property 'idAttribute' of undefined" is thrown. That's because class methods has no property "prototype", which is used in Backbone.Collection.prototype.modelId method.

### Expected behavior

No error thrown if collection has property "model" defined as class method.

### Actual behavior

Error "Cannot read property 'idAttribute' of undefined" thrown.
Here is an example: https://jsfiddle.net/e1ybjmw3/

### Environment

1. Backbone version: 1.3.3
2. Additional build tools, etc: no build tools (that's important, because if babel used all works fine)

avasuro:
Yes, but it looks unpleasant, especially if this is the only situation in application where such approach (extending classes directly through prototype) is used, plus if the collection is extended with many methods then you need to scroll down several screens to see the "model" property.
It kills all the benefits of syntaxic sugar of es6 classes.

dreyks:
yeah that's ugly. I'm writing in coffeescript so it doesn't look that bad for me, but still a hack

syntruth:
This is my patched solution:

https://gist.github.com/syntruth/a76762aab0e442833bfd88bd937aa766

...it's not pretty, but it's working for now.

If I wasn't using Backbone Stickit, I'd moved to the Backbone-ES6 version.

dreyks:
Backbone-ES6 is literally the same code but written in ES6. I has all the same issues

ogonkov:
Another way, to make `model` a function, w/out using prototype is use `get`, i found it a bit more good-looking

```js
class C extends Backbone.Collection {
  get model() {
    return function(attrs, opts) {
      return new M(attrs, opts);
    };
  }
}
```

jgonggrijp:
If you set `model` to a function that returns an instance (and doesn't have a `prototype.idAttribute`), you need to define `modelId` as well. Documented [over here](https://backbonejs.org/#Collection-modelId):

> If your collection uses a [model factory](https://backbonejs.org/#Collection-model) and those models have an `idAttribute` other than `id` you must override this method.

PR #4234, opened 2020-03-03
Add check for `this.model.prototype`

### LINKED ISSUE #4233: Couldn't construct model, with polymorphic models
Following code is throwing `TypeError`

```js
class MyModel extends Backbone.Model {
  validate(attributes) {
  	return null;
  }
}

class MyCollection extends Backbone.Collection {
  modelId(attr) {
    return attr.x;
  } 

  model(attributes, options) {
    return new MyModel(attributes, options);
  }
}

try {
const inst = new MyCollection([{a: 2}]);
} catch(ex) {
  alert(ex)
}
```

```
TypeError: this.model is not a constructor
    at MyCollection._prepareModel (VM16 backbone.js:1137)
    at MyCollection.set (VM16 backbone.js:889)
    at MyCollection.add (VM16 backbone.js:814)
    at MyCollection.reset (VM16 backbone.js:960)
    at new Backbone.Collection (VM16 backbone.js:765)
    at new MyCollection ((index):40)
    at window.onload ((index):51)
```

https://jsfiddle.net/ogonkov/ng19ae6d/

I think it have same root with #4203

ogonkov:
Simple check that `this.model` is constructor. We could have more complicated check, but looks like it doesn't required.

Fixes #4233

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1134,7 +1134,15 @@
       }
       options = options ? _.clone(options) : {};
       options.collection = this;
-      var model = new this.model(attrs, options);
+
+      var model;
+      if (this.model.prototype) {
+        model = new this.model(attrs, options);
+      } else {
+        // ES class methods didn't have prototype
+        model = this.model(attrs, options);
+      }
+
       if (!model.validationError) return model;
       this.trigger('invalid', this, model.validationError, options);
       return false;
--- a/test/collection.js
+++ b/test/collection.js
@@ -2118,7 +2118,7 @@
           '        var MyModel = Backbone.Model.extend({});\n' +
           '        return new MyModel(attrs, options);\n' +
           '    }\n' +
-          '}).model');
+          '}).model')();
     } catch (error) {
       model = error;
     }
--- a/test/collection.js
+++ b/test/collection.js
@@ -2138,14 +2138,8 @@
       model: model
     });
 
-    var result, instance;
-    try {
-      instance = new MyCollection([{a: 2}]);
-      result = true;
-    } catch (error) {
-      result = false;
-    }
+    var instance = new MyCollection([{a: 2}]);
 
-    assert.ok(result && instance, 'Should instantiate collection with model');
+    assert.ok(instance, 'Should instantiate collection with model');
   });
 })(QUnit);
--- a/test/collection.js
+++ b/test/collection.js
@@ -2113,12 +2113,12 @@
   QUnit.test('#4233 - can instantiate new model in ES class Collection', function(assert) {
     var model;
     try {
-      model = new Function('return ' + '({\n' +
+      model = new Function('return ({\n' +
           '    model(attrs, options) {\n' +
           '        var MyModel = Backbone.Model.extend({});\n' +
           '        return new MyModel(attrs, options);\n' +
-          '    }).model;\n' +
-          '}');
+          '    }\n' +
+          '}).model');
     } catch (error) {
       model = error;
     }
--- a/test/collection.js
+++ b/test/collection.js
@@ -2109,4 +2109,43 @@
     var collection = new Backbone.Collection([model]);
     assert.ok(collection.get(model));
   });
+
+  QUnit.test('#4233 - can instantiate new model in ES class Collection', function(assert) {
+    var model;
+    try {
+      model = new Function('return ' + '({\n' +
+          '    model(attrs, options) {\n' +
+          '        var MyModel = Backbone.Model.extend({});\n' +
+          '        return new MyModel(attrs, options);\n' +
+          '    }).model;\n' +
+          '}');
+    } catch (error) {
+      model = error;
+    }
+
+    if (model instanceof SyntaxError) {
+      assert.expect(0);
+      return;
+    }
+
+    assert.expect(1);
+
+    var MyCollection = Backbone.Collection.extend({
+      modelId: function(attr) {
+        return attr.x;
+      },
+
+      model: model
+    });
+
+    var result, instance;
+    try {
+      instance = new MyCollection([{a: 2}]);
+      result = true;
+    } catch (error) {
+      result = false;
+    }
+
+    assert.ok(result && instance, 'Should instantiate collection with model');
+  });
 })(QUnit);


ogonkov:
@jashkenas can you look at this PR? Seems minor change

jaapz:
Would be nice to have this merged, because the documentation on backbonejs.org currently is inaccurate (https://backbonejs.org/#Collection-model), and it's impossible to implement polymorphic models based on attrs as of now

backbone.js
```diff
@@ -1134,7 +1134,15 @@
...
       options.collection = this;
-      var model = new this.model(attrs, options);
+
+      var model;
+      if (this.model.prototype) {
+        model = new this.model(attrs, options);
+      } else {
+        // ES class methods didn't have prototype
+        model = this.model(attrs, options);
+      }
```
jgonggrijp, line 1144: This change doesn't make any difference. Try this:

```js
function returnPlainObject() {
    return {};
}

var x = new returnPlainObject();

console.log(x, x instanceof returnPlainObject);
```

jgonggrijp:
Already fixed, see #3966 and https://github.com/jashkenas/backbone/issues/4203#issuecomment-1003768061.

backbone.js
ogonkov: I'm not quite follow you. Did example from attached issue works?

jgonggrijp: Regarding #4233, I realize I might have jumped to conclusions too early, so I have reopened it. My apologies on that one.

Regarding this review comment, I just meant that your change will not make a difference. If `this.model` is a function that returns an instance of a `Backbone.Model` subclass, then `this.model(...)` does exactly the same thing as `new this.model(...)`. Therefore your change will also not fix #4233. You would need to bring a solid test to convince me otherwise. :-)

jgonggrijp: Ah, I understand now. The "class methods are nonconstructible" problem. Man, ES classes are such a pain.

I would prefer that you add a standalone `construct` function with the same semantics as `new` (but avoiding the keyword), so we can write `var model = construct(this.model, attrs, options)` over here. That function goes like this (Underscore and ES3 syntax for backwards compatibility):

```js
var construct = _.restArguments(function(ctor, args) {
    // _.create works even if the argument is undefined
    var instance = _.create(ctor.prototype);
    var result = ctor.apply(instance, args);
    return _.isObject(result) ? result : instance;
};
```

However, this is not a hard condition for acceptance. The only hard condition is that you add a test, so we can detect it if this breaks again.

ogonkov: I'm not sure that i can add test here, as long it involves es6 syntax, even with babel es class methods will likely would be transpiled to `function`

jgonggrijp: You can write a standalone function and then manually set its `prototype` to `undefined`. That should give the same problem with the old code. Probably good to first check that your test fails with the old logic, though.

jgonggrijp: I learned from a quick test that this approach probably won't work:

```js
function test() {}
test.prototype = undefined;
new test; // no error, returns empty object
```

I'm still looking whether I can hack this in with ES3 syntax. Failing that, we can have a test with ES6 syntax, as long as we put it in a separate module (so we can skip it in pre-ES6 environments).

jgonggrijp: Here is a way to create a non-constructor within an ES3 module if the engine supports it, and skip the test otherwise:

```js
// ES6 notation as a string
var esMethod = '{\n' +
'    model(attrs, options) {\n' +
'        return new TestModel(attrs, options);\n' +
'    }\n' +
'}';

(globalThis || self || window).TestModel = Backbone.Model.extend(...);

try {
    var nonConstructible = Function('return ' + esMethod + '.model;')();
    var TestCollection = Backbone.Collection.extend({
        model: nonConstructible,
        modelId: function(attrs) {...}
    });
    // remainder of test here
} catch (error) {
    // syntax not supported, skip test
}
```

ogonkov: But this test will always fail, because of lack of support?

ogonkov: I see syntax error in Phantom, i'm not sure that my code even invoked

jgonggrijp: The trick is in the `Function(...)()` and the `try`/`catch`. Do you see how that works?

I just realized that you may need to put *only* the `Function(...)()` in the `try` block and return early from the `catch`. Otherwise you will not be able to distinguish the syntax being unsupported from the `TypeError` due to `collection.model` not being constructible.

ogonkov: What should test do if there a syntaxerror?

jgonggrijp: It will go into the `catch` block.

jgonggrijp: > I see syntax error in Phantom, i'm not sure that my code even invoked

Oops, missed this comment. Could you show the exact code you're feeding to PhantomJS?

ogonkov: Check out my latest commit with test

jgonggrijp (changes requested):
I wanted to make sure that the test actually fails, if executed in an environment that supports the notation and your fix hasn't been applied yet. So here's what I did:

1. I made a local checkout of your branch.
2. I reverted the commit with your fix, so I had the test but not the fix.
3. Ran the tests in PhantomJS, which doesn't support the notation. Your test was skipped as expected.
4. Ran the tests in Safari, which does support the notation. (Tip: you can run the tests in any browser of choice by simply opening `test/index.html`. You can even run the tests [online](https://backbonejs.org/test/).) Your test was skipped *again*, *not* as expected.
5. I experimentally removed the `try`/`catch` around the `Function('...')` bit, so I could see the `SyntaxError` that was apparently being thrown in Safari. (Tip: while this is the type of change you don't want to keep, you can still commit it. This makes it easy to revert the change again later. You can remove all commits you don't want to share with the world before pushing by running `git rebase -i`. See `git help rebase` for more information.)
6. I ran the test again in Safari. I turned out that the `SyntaxError` was due to a nesting mistake; you put `).model;` after the `model(...){...}` method, while it should have been after the object notation that contained it. Looking closer at what you had written, I also noticed that you had forgotten the immediate invocation of the created `Function`.
7. I added new commits in which I rectified the nesting error and added the missing invocation. The last comment below contains the changes that you need to make in order to rectify this.
8. Ran the test again in Safari. The `SyntaxError` was gone and it was now failing with `false`. This was as expected in principle, but I wanted to make very sure that it was due to the "function is not a constructor" bug we have been chasing, so ...
9. ... in yet another commit, I removed the `try`/`catch` around the `instance = new MyCollection([{a: 2}]);` line as well. I'm suggesting this change for your PR as well, see second comment below.
10. I ran the tests again in Safari, and there it was: `function is not a constructor`. Hooray.
11. I ran the tests again in PhantomJS, and it failed with a `SyntaxError` due to the lack of support for object method notation: `Unexpected token '('`. Also as expected. This would most likely have happened already after step 5, but I didn't test PhantomJS at the time.
12. I reverted my change from step 5 (i.e., I put back the `try`/`catch` around the `Function(...)()` part).
13. Tested again. Test was skipped in PhantomJS as intended. Test failed with `function is not a constructor` in Safari, as intended.
14. Undid my revert from step 2, i.e., I put back your fix.
15. Tested one final time. Test skipped in PhantomJS, test successful in Safari, all good.

In conclusion, you'll need to make a couple of changes, but after that I'll be confident that you fixed the bug.

test/collection.js
```diff
@@ -2109,4 +2109,43 @@
...
+    if (model instanceof SyntaxError) {
+      assert.expect(0);
+      return;
+    }
+
+    assert.expect(1);
+
+    var MyCollection = Backbone.Collection.extend({
+      modelId: function(attr) {
+        return attr.x;
```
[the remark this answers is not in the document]
ogonkov: I didn't remember why i have added this

jgonggrijp (approved):

jgonggrijp:
Thanks @ogonkov!

PR #4235, opened 2020-03-05
fixes #4217 static inheritance

### LINKED ISSUE #4217: Function extend does not properly expose static properties from base classes
If a backbone model is an ES6 class with a statc property, and one use `Model.extend` to make a subclass (instead of the ES6 `extends`), the static properties of the base model are not added to the subclass.

oliverfoster:
[#4217](https://github.com/jashkenas/backbone/issues/4217)
* Added ES6-style constructor and static property inheritance rather than just copying the enumerable static properties each time.


oliverfoster:
closed as no response

Issue #4246, opened 2022-01-08
Upgrade devDependencies, add lockfile


jgonggrijp:
Since there was no lockfile yet, I decided to pull out `yarn`. Yarn gave me lots of warnings, so it's clearly time for some upgrades.

ogonkov:
Is yarn really necessary?

jgonggrijp:
No, of course every other package manager will work as well. Would you prefer another one, and if so, which and why?

ogonkov:
I guess npm is good enough in current state. An im one of this people, that believe that lock file is not required for _library_.

jgonggrijp:
Now we seem to be discussing *two* topics.

Whether to have a lockfile or not: as far as I know, all current package managers (at the very least npm and yarn) give you a lockfile by default, creating it on the spot if it didn't exist yet. So the question is only whether to commit it or not. Given that every single person who installs Backbone's `devDependencies` is going to have a lockfile anyway, we might as well all have the same one, no? In fact, I think that is one of the main purposes of a lockfile: ensuring that everyone on the team is working on the exact same thing.

Which package manager to use: sure, npm is good enough, but so is yarn. On top of that, yarn is faster and deterministic and it deduplicates dependencies. I find it convenient. I can be convinced to use another package manager, but the argument will have to be stronger than "good enough".

ogonkov:
`package-lock.json` will not be packaged. If we commit it it would be for our own CI, user will end with different package-lock.

ogonkov:
I'm not pushing here to avoid package-lock, just opinion

jgonggrijp:
I appreciate your opinion, and your active participation in general. Please continue sharing your opinion.

Yes, the lockfile is for our team and for CI. I consider that enough reason to commit it. Let's put that part of the discussion aside, because I believe there is really no added value in keeping it out of Git.

Regarding whether to use npm or yarn, I'm still leaning towards yarn, but I'll allow some time for more people to chime in with opinions and arguments. I'm not in a hurry.

paulfalgout:
I typically prefer npm to yarn, but as @jgonggrijp is the most active contributor at this point, I'd say dealers choice.  However I do think keeping backbone and underscore in similar tooling is ideal, and it looks like underscore is using npm?  Is it switching to yarn?

jgonggrijp:
Why do you typically prefer npm over yarn, @paulfalgout? I would be really interested to learn that.

[...]

PR #4248, opened 2022-01-10
Replace Travis with GitHub Workflows

### LINKED ISSUE #4247: Replace travis with GH Workflows
Travis seems not running for few last commits. I guess GH workflows woild be more robust decision.

I can create PR.

ogonkov:
Fixes #4247


.github/workflows/tests.yml
```diff
@@ -0,0 +1,34 @@
...
+    env:
+      NPM_CONFIG_PROGRESS: "false"
+
+    runs-on: ubuntu-latest
+
+    steps:
+      - uses: actions/checkout@v2
+      - uses: actions/setup-node@v2
+        with:
+          node-version: 6
```
jgonggrijp, line 22: @ogonkov I presume this step implicitly performs `npm install`. Would it be somehow possible to pass the `--no-audit` flag? I'm asking because of https://github.com/jashkenas/backbone/issues/4246#issuecomment-1013860308.

ogonkov: I just migrate exist travis config. We can improve it in this PR, sure

jgonggrijp: Good point, I realize I'm moving the goalposts a bit. It's not a hard condition in any case, the most important thing is that this works again.

.github/workflows/tests.yml
```diff
@@ -0,0 +1,38 @@
...
```
jgonggrijp, line 25: Now that we're trying to save energy anyway, please enable [caching](https://github.com/marketplace/actions/setup-node-js-environment#caching-packages-dependencies) as well:

```suggestion
          node-version: 6
          cache: 'npm'
```

.github/workflows/tests.yml
```diff
@@ -0,0 +1,38 @@
...
+      - run: npm install --no-audit && npm install karma-cli karma-sauce-launcher
```
jgonggrijp, line 26: The second `npm install` should probably include `--no-audit` as well.

```suggestion
      - run: npm install --no-audit && npm install --no-audit karma-cli karma-sauce-launcher
```

jgonggrijp (approved):
Looks good to me overall. 👍

.github/workflows/tests.yml
```diff
@@ -0,0 +1,39 @@
...
+    branches:
+      - ${{ github.event.repository.default_branch }}
+    paths:
+      - '*.js'
+      - 'package.json'
+
+jobs:
+  tests:
+    env:
+      NPM_CONFIG_PROGRESS: "false"
```
paulfalgout, line 17: This might be a thing in node v6.. I'm not sure.. but npm for a while has defaulted progress to false if it's running in a known CI..   so I _assume_ this is no longer necessary.. but node 6 might be memory-holed for me.

paulfalgout (approved):
One small comment, but I'm not sure if it's relevant currently

.github/workflows/tests.yml
jgonggrijp: Yes, it's off by default in CI: https://docs.npmjs.com/cli/v6/using-npm/config#progress.

Though it probably doesn't hurt to turn it off explicitly. Out of curiosity, why does it need to be off, anyway?

paulfalgout: It's just a fraction slower to show the progress bar and nobody is watching the progress on CI anyway

paulfalgout: Also it's off for npm v6.. but I dunno what version of npm is loaded for node 6?

ogonkov: It is npm v3 for node 6

PR #4249, opened 2022-01-12
Add model changeId event

### LINKED ISSUE #3882: Internal _byId index of Collections wrong during model change event
During model change event, when `id` attribute changes, the internal `_byId` index of containing collection is invalid. Because of this, the model cannot be obtained with `collection.get(id)` method, even though it has ID attribute already set. This is non-intuitive as internal indexing of collections should be transparent to the end user, and when `collection.get(ID)` returns null, it should be be safe to assume that the model with such ID does not exist in collection. 

Internally, this issue is caused by delayed firing of collection's `_onModelEvent` method which maintains the `byId` index (https://github.com/jashkenas/backbone/blob/05fde9e201f7e2137796663081105cd6dad12a98/backbone.js#L1144). It is being bound to the `all` event on model level, and `all` handlers are invoked only after all other events are processed (https://github.com/jashkenas/backbone/blob/05fde9e201f7e2137796663081105cd6dad12a98/backbone.js#L357) making the collection update it's index only after change handlers are executed.

Please see the example below:

```
class Entry extends Backbone.Model
  initialize: =>
    @on 'change', (e) =>
      console.log('onChange Model ', e.id, window.collection.get(e.id), window.collection.get(e.cid), window.collection.get(e.cid).id)

class Collection extends Backbone.Collection
  model: Entry
  initialize: =>
    @on 'change', (e) =>
      console.log('onChange Collection', e.id, window.collection.get(e.id), window.collection.get(e.cid))

window.collection = new Collection()
entry = collection.add({'name':'unsaved'})
entry.set({'name':'saved', 'id':10})
```

It's console output is:

```
onChange Model 10 undefined Entry {cid: "c1", …} 10
onChange Collection 10 Entry {cid: "c1", …} Entry {cid: "c1", …}
```

Please note that in the model change handler the collection is in invalid state, as `collection.get(10)` produced `undefined` while `collection.get('c1').id` returns 10, meaning that the object with ID 10 is actually in the collection.

Also, the second line shows inconsistency with the `change` event bound at the collection level, which behaves correctly.

### LINKED ISSUE #4159: Collection does not purge internal `_byId` reference to a removed model

paulfalgout:
Fixes #3882
Fixes #4159

Backbone does a bit of extra work to determine when to update `_byId` on every model change event because `change:id` will not work if `idAttribute` has changed.  This causes issues as the `change` event happens after every `change:` event which means during a change the `_byId` hasn't updated.  Rather than adding complexity to collection the solution is to have the model notify with the id changes.

If adding a public event isn't desired, for an internal solution the model is aware of it's collection and could modify model.collection._byId directly within the set.

Either of these solutions seem preferrable to handling `change:[idAttribute]`

Replaces the need for:
https://github.com/jashkenas/backbone/pull/4227/files#diff-c773bb9be277f0f3f2baa308b6e0f3a486790fe99fea81ddd0ba409846250571R1205

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -516,7 +516,11 @@
       }
 
       // Update the `id`.
-      if (this.idAttribute in attrs) this.id = this.get(this.idAttribute);
+      if (this.idAttribute in attrs) {
+        var prevId = this.id;
+        this.id = this.get(this.idAttribute);
+        this.trigger('changeId', this, prevId, options);
+      }
 
       // Trigger all relevant attribute changes.
       if (!silent) {
@@ -1207,13 +1211,11 @@
       if (model) {
         if ((event === 'add' || event === 'remove') && collection !== this) return;
         if (event === 'destroy') this.remove(model, options);
-        if (event === 'change') {
+        if (event === 'changeId') {
           var prevId = this.modelId(model.previousAttributes(), model.idAttribute);
           var id = this.modelId(model.attributes, model.idAttribute);
-          if (prevId !== id) {
-            if (prevId != null) delete this._byId[prevId];
-            if (id != null) this._byId[id] = model;
-          }
+          if (prevId != null) delete this._byId[prevId];
+          if (id != null) this._byId[id] = model;
         }
       }
       this.trigger.apply(this, arguments);
--- a/index.html
+++ b/index.html
@@ -1032,6 +1032,7 @@ view.stopListening(model);
       <li><b>"reset"</b> (collection, options) &mdash; when the collection's entire contents have been <a href="#Collection-reset">reset</a>.</li>
       <li><b>"sort"</b> (collection, options) &mdash; when the collection has been re-sorted.</li>
       <li><b>"change"</b> (model, options) &mdash; when a model's attributes have changed.</li>
+      <li><b>"changeId"</b> (model, previousId, options) &mdash; when the model's id has been updated.</li>
       <li><b>"change:[attribute]"</b> (model, value, options) &mdash; when a specific attribute has been updated.</li>
       <li><b>"destroy"</b> (model, collection, options) &mdash; when a model is <a href="#Model-destroy">destroyed</a>.</li>
       <li><b>"request"</b> (model_or_collection, xhr, options) &mdash; when a model or collection has started a request to the server.</li>
--- a/test/collection.js
+++ b/test/collection.js
@@ -122,7 +122,7 @@
   });
 
   QUnit.test('update index when id changes', function(assert) {
-    assert.expect(4);
+    assert.expect(6);
     var collection = new Backbone.Collection();
     collection.add([
       {id: 0, name: 'one'},
@@ -130,7 +130,11 @@
     ]);
     var one = collection.get(0);
     assert.equal(one.get('name'), 'one');
-    collection.on('change:name', function(model) { assert.ok(this.get(model)); });
+    collection.on('change:name', function(model) {
+      assert.ok(this.get(model));
+      assert.equal(model, this.get(101));
+      assert.equal(this.get(0), null);
+    });
     one.set({name: 'dalmatians', id: 101});
     assert.equal(collection.get(0), null);
     assert.equal(collection.get(101).get('name'), 'dalmatians');


backbone.js
```diff
@@ -1207,13 +1210,11 @@
       if (model) {
         if ((event === 'add' || event === 'remove') && collection !== this) return;
         if (event === 'destroy') this.remove(model, options);
-        if (event === 'change') {
+        if (event === 'changeId') {
           var prevId = this.modelId(model.previousAttributes(), model.idAttribute);
```
jgonggrijp, line 1214: This line computes the previous id based on `model`'s *current* `idAttribute`.

jgonggrijp (changes requested):
Thanks for this insightful fix @paulfalgout. Great to have you still on board.

> Backbone does a bit of extra work to determine when to update `_byId` on every model change event because `change:id` will not work if `idAttribute` has changed.

Just a nuance: if I'm understanding the situation correctly, the purpose of the extra work is not to support changes in the `idAttribute` of individual model classes. The line that I have commented on below illustrates that Backbone does *not* support such a thing, and supporting it would in fact not be straightforward. Fortunately, the documentation also does not suggest that this is supported.

The real purpose of the extra work is to support polymorphic collections of multiple model types with different idAttributes. Your changes seem just as reasonable under these assumptions.

> This causes issues as the `change` event happens after every `change:` event which means during a change the `_byId` hasn't updated. Rather than adding complexity to collection the solution is to have the model notify with the id changes.

Brilliant. I think this is the only correct way to solve the problem. I also like the event name you've chosen; `changeId` fits in comfortably with `change:id` and `modelId`.

> If adding a public event isn't desired, for an internal solution the model is aware of it's collection and could modify model.collection._byId directly within the set.

No, this would be an incomplete solution. Each model has only one `collection` reference, but it might be a member of other collections at the same time. Those other collections would not be notified in this case.

> Either of these solutions seem preferrable to handling `change:[idAttribute]`

`change:${idAttribute}` does not handle the polymorphic scenario, but it does handle the scenario where a model is a member of multiple collections. So I would say this solution and the previous one are about equally bad.

> Replaces the need for: https://github.com/jashkenas/backbone/pull/4227/files#diff-c773bb9be277f0f3f2baa308b6e0f3a486790fe99fea81ddd0ba409846250571R1205

Yes, that's just a combination of two bad solutions.

I would like to request two refinements:

- [ ] Include the previous `id` in the event payload of `changeId`. I suggest placing it between `this` and `options`.
- [ ] Add `changeId` to the [Catalog of Built-in Events](https://backbonejs.org/#Events-catalog) in the documentation.

Otherwise I'm completely convinced by your test and implementation. 🚀

paulfalgout:
`the purpose of the extra work is not to support changes in the idAttribute of individual model classes`
Ah yes, I was being lazy and using `idAttribute` to refer to whatever attribute is being considered the id rather than the value of `idAttribute` itself.  I think we're on the same page there.

And yep, polymorphic does invalidate the other suggestion.  Good catch.  I'll update the PR and help in general when I can with changes.  I've been lurking a bit.  Great to see some movement here.

jgonggrijp:
> so regarding:
> 
> > Include the previous id in the event payload of changeId. I suggest placing it between this and options.
> 
> I agree I missed the opportunity to align this with `change:id` but in that case wouldn't it be the current id that it is changed to?

Yes that would be more consistent, but then there would be no short way to obtain the previous `idAttribute` (compare `model.id` to `model.previous(model.idAttribute)`). For previous attributes we have `previousAttributes()`, but there is no corresponding `previousProperties()` (which would make it *entirely* consistent), and I don't think adding such a method is warranted.

For what it's worth, I didn't make the remark because of consistency, but because I think event listeners might want to inspect the previous id. Initially, I even thought it might be used in `Collection.prototype._onModelEvent`, but then I realized that would interfere with the `modelId` semantics.

jgonggrijp:
@paulfalgout I intend to make a new release. Would you consider the new `changeId` event a new feature or mostly just a bugfix? That's going to make the difference between version 1.4.1 and 1.5.0.

paulfalgout:
Good question.. it's feature-y to me, but not one anyone was specifically asking for.  If this is the only feature-like thing, maybe this is a patch release?

PR #4251, opened 2022-01-21
Fix workflows

### LINKED ISSUE #4247: Replace travis with GH Workflows

ogonkov:
Fixes #4247

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,6 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 6
-          cache: 'npm'
       - run: npm install --no-audit && npm install --no-audit karma-cli karma-sauce-launcher
       - name: Setup Sauce Labs
         run: sh -e /etc/init.d/xvfb start
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,6 @@ jobs:
         env:
           SAUCE_USERNAME: ${{ secrets.SauceUsername }}
           SAUCE_ACCESS_KEY: ${{ secrets.SauceAccessKey }}
-          BUILD_NUMBER: ${{ GITHUB_RUN_NUMBER }}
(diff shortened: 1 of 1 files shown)


.github/workflows/tests.yml
```diff
@@ -23,17 +24,17 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 6
-          cache: 'npm'
```
jgonggrijp, line 26: Would it be possible to put back this line, or is it somehow preventing the whole thing from working?

jgonggrijp (commented):
Thanks for your hard work @ogonkov!

I seem to recall that secrets are only available when the person triggering the CI (i.e., you in this case) has the right to access them. That might explain why Sauce is being skipped in this PR; we will see after merging whether this is indeed the (main) reason.

I did notice one error message in advance of the PhantomJS test run, about installing the `karma-sauce-launcher`. This might still be worth investigating, if you haven't already done so. https://github.com/jashkenas/backbone/pull/4251/checks#step:5:10

.github/workflows/tests.yml
ogonkov: It actually looking for lock file, so it would be possible after package-lock commit

ogonkov:
All other vars, that i use is also empty. 

```
Run echo "Value: $GITHUB_RUN_NUMBER"
  echo "Value: $GITHUB_RUN_NUMBER"
  printenv
  shell: /usr/bin/bash -e {0}
  env:
    NPM_CONFIG_PROGRESS: false
    SAUCE_USERNAME: 
    SAUCE_ACCESS_KEY: 
    BUILD_NUMBER: $GITHUB_RUN_NUMBER-foobar
    BUILD_ID: 
    JOB_NUMBER: 
Value: 38
```

BUILD_ID and JOB_NUMBER is not secret, but also missing.

PR #4252, opened 2022-01-22
Fix test workflow, bis

jgonggrijp:
This is a continuation of @ogonkov's work at #4251, which I'm personally pushing directly to the main  repo in order to rule out any effects from permissions. I also wanted to try a couple of things. If I end up merging this branch, #4251 will automatically be recognized as merged as well.

### DIFF
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,6 +25,7 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 6
+          cache: 'npm'
       - name: Install dependencies
         run: |
           npm install --no-audit
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -28,7 +28,7 @@ jobs:
           cache: 'npm'
       - name: Install dependencies
         run: |
-          npm install --no-audit
+          npm ci
           npm install --no-audit karma-cli karma-sauce-launcher
       - name: Test
         run: npm test
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -24,7 +24,6 @@ jobs:
       - uses: actions/setup-node@v2
         with:
           node-version: 6
-          cache: 'npm'
       - run: npm install --no-audit && npm install --no-audit karma-cli karma-sauce-launcher
       - name: Setup Sauce Labs
         run: sh -e /etc/init.d/xvfb start
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -35,6 +35,6 @@ jobs:
         env:
           SAUCE_USERNAME: ${{ secrets.SauceUsername }}
           SAUCE_ACCESS_KEY: ${{ secrets.SauceAccessKey }}
-          BUILD_NUMBER: ${{ GITHUB_RUN_NUMBER }}
-          BUILD_ID: ${{ GITHUB_RUN_ID }}
-          JOB_NUMBER: ${{ GITHUB_JOB }}
+          BUILD_NUMBER: ${{ env.GITHUB_RUN_NUMBER }}
+          BUILD_ID: ${{ env.GITHUB_RUN_ID }}
+          JOB_NUMBER: ${{ env.GITHUB_JOB }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -25,10 +25,6 @@ jobs:
         with:
           node-version: 6
       - run: npm install --no-audit && npm install --no-audit karma-cli karma-sauce-launcher
-      - name: Setup Sauce Labs
-        run: sh -e /etc/init.d/xvfb start
-        env:
-          DISPLAY: ':99.0'
       - run: npm test
       - run: ./node_modules/.bin/karma start karma.conf-sauce.js
         env:
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -36,6 +36,3 @@ jobs:
         env:
           SAUCE_USERNAME: ${{ secrets.SauceUsername }}
           SAUCE_ACCESS_KEY: ${{ secrets.SauceAccessKey }}
-          BUILD_NUMBER: ${{ env.GITHUB_RUN_NUMBER }}
-          BUILD_ID: ${{ env.GITHUB_RUN_ID }}
-          JOB_NUMBER: ${{ env.GITHUB_JOB }}
--- a/.github/workflows/tests.yml
+++ b/.github/workflows/tests.yml
@@ -3,10 +3,10 @@ name: Test
 on:
   push:
     branches:
-      - ${{ github.event.repository.default_branch }}
+      - master
   pull_request:
     branches:
-      - ${{ github.event.repository.default_branch }}
+      - master
     paths:
       - '*.js'
       - 'package.json'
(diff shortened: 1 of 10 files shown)


jgonggrijp:
@ogonkov @paulfalgout I got the Sauce Labs tests to run, mostly reliably, with the exception of Firefox 11. See the [details](https://github.com/jashkenas/backbone/runs/4904639466?check_suite_focus=true) of the above (failed) workflow run.

If you agree with the logic changes you see here, we can merge this. However, we need to make a decision about the lowest Firefox version.

The tests do pass in Firefox 11, as you can see on Sauce Labs: https://app.eu-central-1.saucelabs.com/tests/08d1010d4bea46dda25b29f2a6a9a698. However, something is going wrong in the connection with Karma. I tried this on a couple of OS versions and I had a single attempt on Windows 10 where the connection suddenly worked ([run 53](https://github.com/jashkenas/backbone/runs/4904532385?check_suite_focus=true)). So far, I haven't been able to reproduce this, but this is the reason why the `karma.conf-sauce.js` currently lists Firefox 11 with Windows 10.

I also tried Firefox 18, 25 and 32, but they all behaved similarly flaky, although versions 18 and 32 eventually succeeded on second attempts. See [run 55](https://github.com/jashkenas/backbone/runs/4904599184?check_suite_focus=true).

I see three possible courses of action:

- Keep Firefox 11 and accept that it will fail on most runs.
- Drop Firefox 11 without replacement, making 40 the oldest version. This would mean that we are only testing relatively recent versions of Firefox.
- Find a version between 11 and 40 with a better tradeoff between reliability and age. This can be done with a "quadrisect" (cf `git bisect`) where 18-25-32 was the first recursive step.

Please let me know what you think!

karma.conf-sauce.js
```diff
@@ -72,17 +68,18 @@ module.exports = function(config) {
...
+    concurrency: 2,
 
     // test results reporter to use
     reporters: ['dots', 'saucelabs'],
     port: 9876,
     colors: true,
     logLevel: config.LOG_INFO,
     sauceLabs: {
-      build: 'GH #' + process.env.BUILD_NUMBER + ' (' + process.env.BUILD_ID + ')',
+      build: 'GH #' + process.env.GITHUB_RUN_NUMBER + ' (' + process.env.GITHUB_RUN_ID + ')',
```
ogonkov, line 79: I tired to set some env neutral vars...

.github/workflows/tests.yml
```diff
@@ -22,18 +24,16 @@ jobs:
...
+      - name: Test
+        run: npm test
+      - name: Test in Sauce Labs
+        run: ./node_modules/.bin/karma start karma.conf-sauce.js
         env:
           SAUCE_USERNAME: ${{ secrets.SauceUsername }}
           SAUCE_ACCESS_KEY: ${{ secrets.SauceAccessKey }}
-          BUILD_NUMBER: ${{ GITHUB_RUN_NUMBER }}
-          BUILD_ID: ${{ GITHUB_RUN_ID }}
-          JOB_NUMBER: ${{ GITHUB_JOB }}
```
ogonkov, line 39: We can workaround this by prependings vars to the command like

```
BUILD_NUMBER="$GITHUB_RUN_NUMBER" BUILD_ID="$GITHUB_RUN_ID" JOB_NUMBER="$GITHUB_JOB" ./node_modules/.bin/karma start karma.conf-sauce.js
```

This will keep sauce config env-neutral

karma.conf-sauce.js
```diff
@@ -2,37 +2,31 @@ var _ = require('underscore');
 
 // Browsers to run on Sauce Labs platforms
 var sauceBrowsers = _.reduce([
-  ['firefox', '35'],
-  ['firefox', '30'],
-  ['firefox', '21'],
-  ['firefox', '11'],
-  ['firefox', '4'],
+  ['firefox', 'latest'],
```
ogonkov, line 5: firefox have ESR (extended support release), i guess it is good idea to add it here aswell

jgonggrijp: I wanted to do this but then I found out that Firefox ESR isn't explicitly listed on the page with supported OSs and browsers. https://saucelabs.com/platform/supported-browsers-devices

ogonkov: oh, okay then

ogonkov:
Can you run sause labs config with [`config.LOG_DEBUG`](https://karma-runner.github.io/6.3/config/configuration-file.html#loglevel)? May be we get more info about what exactly the problem is

jgonggrijp:
That was a good suggestion. The visible difference between FF11 and the other browsers was that FF11 was creating additional socket connections and then immediately closing them again. I tried increasing Karma's `browserDisconnectTimeout`, but that only seems to lead to accumulation of even more sockets.

Haven't been able to find a solution for "Response has empty body" or "Request failed with status 204" yet. Still searching the web.

ogonkov:
Some requests (which one?) has been responded with empty body, that was unexpected for client. There also warn about old client. May be it is something related to old client, used by plugin?

jgonggrijp:
Yes, good suggestion. I was just considering to upgrade the `devDependencies` in case that's playing a role. Will do that now.

jgonggrijp:
@ogonkov Upgrading the dependencies doesn't seem to have improved matters much, except that there is more debug log output. I'm tired now, so I'll resume this chore at a later date.

If you have any ideas in the meanwhile, please feel free to submit a PR to the `patch-1` branch on [my fork](https://github.com/jgonggrijp/backbone). I'll merge it as soon as I see it and push it to the main fork, so it will run in here with full permissions.

jgonggrijp:
@ogonkov I'm just going to remove Firefox 11 for now, so at least we have something to work with. I will create a new issue ticket to document the open challenge.

jgonggrijp:
I'm stuck again because Sauce Labs hasn't converted my trial account to Open Sauce yet and my trial has run out of testing minutes. To be continued.

jgonggrijp:
Update: I haven't heard back from Sauce Labs since January 17. I just made a third attempt at reestablishing contact, using a different email address.

jgonggrijp:
Update again: they finally enabled Open Sauce, so we can run browser tests again. I will resume the debugging labour soon.

.github/workflows/tests.yml
```diff
@@ -22,18 +24,16 @@ jobs:
...
-      - run: npm install --no-audit && npm install --no-audit karma-cli karma-sauce-launcher
-      - name: Setup Sauce Labs
-        run: sh -e /etc/init.d/xvfb start
-        env:
-          DISPLAY: ':99.0'
-      - run: npm test
-      - run: ./node_modules/.bin/karma start karma.conf-sauce.js
+      - name: Install dependencies
+        run: |
+          npm install --no-audit
```
ogonkov, line 31: We want `npm ci` here, since we had package lock

jgonggrijp: Thanks. I will have a look at the documentation, because I'm not really familiar with the `ci` subcommand.

jgonggrijp:
Open Sauce doesn't seem to be *really* enabled. To be continued...

ogonkov:
May be we can mark sause labs tests as non blocking for workflow and merge it in master, to fix units in PRs?

jgonggrijp:
Yes that makes sense. Good idea. How do you do that?

ogonkov (approved):

jgonggrijp:
High five!

PR #4265, opened 2023-07-19
Trigger error from failing Collection.create with wait:true

### LINKED ISSUE #4262: error event not firing for Collection.create with wait true
For [Collection.create](https://backbonejs.org/#Collection-create), the `sync` and `add` events fire correctly, but if `options` are `{wait: true}` and server responds with any error, the `error` event doesn't fire. Without `{wait:true}` the error event fires correctly, but the invalid model is added to the collection.

jgonggrijp:
This fixes #4262.

The bug arose because the transaction with the backend is mediated through `model.save` and the `'error'` event is initially triggered by the model. Normally, the collection will forward all events from the model, but with `wait: true`, the model is not yet added to the collection when the event triggers, so the collection is not listening, either. I addressed this by special-casing the `'error'` event in this particular scenario.

I determined that a similar fix is not required for the `'sync'` event on successful `create`. In that case, the model is still added to the collection before the event triggers.

I foresee two possible criticisms on the solution:

1. One could argue that a similar fix is still needed for the `'request'` event. However, since the timing of this event is fully predictable (i.e., just before the end of the call to `create`), users do not need to listen for it in order to match the timing of a callback. Adding special case code for the `'request'` event as well would further inflate the code size.
2. It is theoretically possible that a user would call `collection.create` with a model that is already in the collection. In that case, with `wait: true`, I expect that the collection will forward the `'error'` event *twice*. If this scenario is considered problematic enough, I am willing to address it.

On the other hand, someone might also argue that this issue does not need fixing at all. The old behavior, while surprising, could be argued to be consistent, since the collection cannot be expected to forward events from a model it does not yet contain.

Reviews and discussion welcome!

CC @paulfalgout @ogonkov @alanhamlett @recursivk

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -1079,10 +1079,9 @@
       if (!wait) this.add(model, options);
       var collection = this;
       var success = options.success;
-      var forwardPristineError;
       options.success = function(m, resp, callbackOpts) {
         if (wait) {
-          m.off('error', forwardPristineError, this);
+          m.off('error', this._forwardPristineError, this);
           collection.add(m, callbackOpts);
         }
         if (success) success.call(callbackOpts.context, m, resp, callbackOpts);
@@ -1096,11 +1095,7 @@
       // causes the collection to listen, and then invoke the callback
       // that triggers the event.)
       if (wait) {
-        forwardPristineError = function(mod, col, opt) {
-          if (this.has(mod)) return;
-          this._onModelEvent('error', mod, col, opt);
-        };
-        model.once('error', forwardPristineError, this);
+        model.once('error', this._forwardPristineError, this);
       }
       model.save(null, options);
       return model;
@@ -1239,8 +1234,19 @@
         }
       }
       this.trigger.apply(this, arguments);
-    }
+    },
 
+    // Internal callback method used in `create`. It serves as a
+    // stand-in for the `_onModelEvent` method, which is not yet bound
+    // during the `wait` period of the `create` call. We still want to
+    // forward any `'error'` event at the end of the `wait` period,
+    // hence a customized callback.
+    _forwardPristineError: function(model, collection, options) {
+      // Prevent double forward if the model was already in the
+      // collection before the call to `create`.
+      if (this.has(model)) return;
+      this._onModelEvent('error', model, collection, options);
+    }
   });
 
   // Defining an @@iterator method implements JavaScript's Iterable protocol.
--- a/backbone.js
+++ b/backbone.js
@@ -1079,10 +1079,26 @@
       if (!wait) this.add(model, options);
       var collection = this;
       var success = options.success;
+      var forwardPristineError;
       options.success = function(m, resp, callbackOpts) {
-        if (wait) collection.add(m, callbackOpts);
+        if (wait) {
+          model.off('error', forwardPristineError);
+          collection.add(m, callbackOpts);
+        }
         if (success) success.call(callbackOpts.context, m, resp, callbackOpts);
       };
+      // In case of wait:true, our collection is not listening to any
+      // of the model's events yet, so it will not forward the error
+      // event. In this special case, we need to listen for it
+      // separately and handle the event just once.
+      // (The reason we don't need to do this for the sync event is
+      // in the success handler above: we add the model first, which
+      // causes the collection to listen, and then invoke the callback
+      // that triggers the event.)
+      if (wait) {
+        forwardPristineError = _.bind(this._onModelEvent, this, 'error');
+        model.once('error', forwardPristineError);
+      }
       model.save(null, options);
       return model;
     },
--- a/backbone.js
+++ b/backbone.js
@@ -1082,7 +1082,7 @@
       var forwardPristineError;
       options.success = function(m, resp, callbackOpts) {
         if (wait) {
-          model.off('error', forwardPristineError);
+          m.off('error', forwardPristineError, this);
           collection.add(m, callbackOpts);
         }
         if (success) success.call(callbackOpts.context, m, resp, callbackOpts);
@@ -1096,8 +1096,11 @@
       // causes the collection to listen, and then invoke the callback
       // that triggers the event.)
       if (wait) {
-        forwardPristineError = _.bind(this._onModelEvent, this, 'error');
-        model.once('error', forwardPristineError);
+        forwardPristineError = function(mod, col, opt) {
+          if (this.has(mod)) return;
+          this._onModelEvent('error', mod, col, opt);
+        };
+        model.once('error', forwardPristineError, this);
       }
       model.save(null, options);
       return model;
--- a/test/collection.js
+++ b/test/collection.js
@@ -654,6 +654,15 @@
     assert.equal(collection.length, 1);
   });
 
+  QUnit.test('failing create with wait:true triggers error event (#4262)', function(assert) {
+    assert.expect(1);
+    var collection = new Backbone.Collection;
+    collection.url = '/test';
+    collection.on('error', function() { assert.ok(true); });
+    collection.create({id: '1'}, {wait: true});
+    this.ajaxSettings.error();
+  });
+
   QUnit.test('initialize', function(assert) {
     assert.expect(1);
     var Collection = Backbone.Collection.extend({
--- a/test/collection.js
+++ b/test/collection.js
@@ -664,6 +664,16 @@
     this.ajaxSettings.error();
   });
 
+  QUnit.test('successful create with wait:true triggers success event (#4262)', function(assert) {
+    assert.expect(2);
+    var collection = new Backbone.Collection;
+    collection.url = '/test';
+    collection.on('sync', function() { assert.ok(true); });
+    var model = collection.create({id: '1'}, {wait: true});
+    model.on('sync', function() { assert.ok(true); });
+    this.ajaxSettings.success();
+  });
+
   QUnit.test('initialize', function(assert) {
     assert.expect(1);
     var Collection = Backbone.Collection.extend({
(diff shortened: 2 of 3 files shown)


alanhamlett:
> 2. It is theoretically possible that a user would call collection.create with a model that is already in the collection. In that case, with wait: true, I expect that the collection will forward the 'error' event twice. If this scenario is considered problematic enough, I am willing to address it.

I think it's better to get an error message rendered twice rather than not at all.

alanhamlett (approved):

paulfalgout (approved):
Makes sense to me

jgonggrijp:
Thanks, both!

I realized something after submitting the PR. Even in the old situation, `create` would return the new model so you could listen on that for the error event. It can even be a oneliner if you don't need a lasting handle to the model:

```js
someCollection.create({...props}, {wait: true}).on('error', ...);
```

In other words, listening for the event is already possible, just not on the collection.

With that in mind, do you still consider the change worthwhile?

alanhamlett:
Also what about `view.listenTo(collection, 'error', callback)`? Does that already work or also not firing?

alanhamlett:
> With that in mind, do you still consider the change worthwhile?

I think it's worthwhile to make sure the error event triggers as expected, because error cases are often not as tested as success cases so someone might not notice the error even never fires in their code.

jgonggrijp:
> Also what about `view.listenTo(collection, 'error', callback)`? Does that already work or also not firing?

Not firing. That depends on the collection forwarding the event again.

Rayraz:
> 2\. It is theoretically possible that a user would call `collection.create` with a model that is already in the collection. In that case, with `wait: true`, I expect that the collection will forward the `'error'` event _twice_. If this scenario is considered problematic enough, I am willing to address it.

Imho, sending the error event twice should be considered a bug. If it's not *that* much of a hassle, I think I would opt to go the extra mile to make sure it only fires once.

If it turns out it will actually be a significat hassle to make sure it only fires once, perhaps that's a sign the status quo is actually more correct than the fix.

Rephrasing the question slightly, if I would have to choose which is less wrong:

1. A collection not forwarding an event for a model that technically isn't part of the collection yet anyway
2. A collection that broadcasts the same error twice

I'd say the first option 1 seems significantly less wrong to me.

backbone.js
```diff
@@ -1079,10 +1079,26 @@
...
       };
+      // In case of wait:true, our collection is not listening to any
+      // of the model's events yet, so it will not forward the error
+      // event. In this special case, we need to listen for it
+      // separately and handle the event just once.
+      // (The reason we don't need to do this for the sync event is
+      // in the success handler above: we add the model first, which
+      // causes the collection to listen, and then invoke the callback
+      // that triggers the event.)
+      if (wait) {
```
jgonggrijp, line 1098: This condition needs to be extended to prevent duplicate listening if the model is already in the collection. The additional function call adds weight, although this can be somewhat softened by sequencing it after the `wait` check.

```suggestion
      if (wait && !this.has(model)) {
```

backbone.js
```diff
@@ -1079,10 +1079,26 @@
       if (!wait) this.add(model, options);
       var collection = this;
       var success = options.success;
+      var forwardPristineError;
       options.success = function(m, resp, callbackOpts) {
-        if (wait) collection.add(m, callbackOpts);
+        if (wait) {
+          model.off('error', forwardPristineError);
+          collection.add(m, callbackOpts);
+        }
```
jgonggrijp, line 1087: However, this introduces a subtle bug if the model is, indeed, already in the collection. In that case, `forwardPristineError` is `undefined` and these lines will remove *all* `'error'` listeners from `model`. To fix that, I could repeat the `this.has` check here, but that is expensive, *or*

backbone.js
```diff
@@ -1079,10 +1079,26 @@
       if (!wait) this.add(model, options);
       var collection = this;
       var success = options.success;
+      var forwardPristineError;
```
jgonggrijp, line 1082: ... I can simply initialize `forwardPristineError` with a bogus value, so that removing it as an `'error'` listener is never problematic.

```suggestion
      var forwardPristineError = _.noop;
```

test/collection.js
```diff
@@ -654,6 +654,26 @@
...
+  QUnit.test('successful create with wait:true triggers success event (#4262)', function(assert) {
+    assert.expect(2);
+    var collection = new Backbone.Collection;
+    collection.url = '/test';
+    collection.on('sync', function() { assert.ok(true); });
+    var model = collection.create({id: '1'}, {wait: true});
+    model.on('sync', function() { assert.ok(true); });
+    this.ajaxSettings.success();
+  });
+
```
jgonggrijp, line 676: (Before doing any of that, though, I would of course first add a test to make sure that there is actually a problem to begin with.)

jgonggrijp (commented):
@Rayraz thanks for chiming in. It's not much of a hassle to prevent double event forwarding (less than one line of code), but it is somewhat costly performancewise. I will highlight the considerations with code comments below:

Rayraz:
I can see why you would be hesitant to introduce the `this.has(model)` check, especially if it becomes significantly slower the more models are in the collection... 🤔 Does the `_.noop` version negate the `this.has(model)` check?

If de-duplicating the event is going to be expensive no matter what, perhaps that's a sign that the status quo is a sensible default behavior. I think the argument could be made that having a collection forward events for a model it doesn't actually own is a nice-to-have rather than a must.

If it's an expensive nice-to-have, would it be so terrible to hide this type of event-forwarding behind a feature flag? The 'unexpectedness' of the default behavior could simply be documented.

For someone who are already being mindful about performance due to working with large datasets, this is probably just fine:

```JavaScript
someCollection.create({...props}, {wait: true}).on('error', callback);
```

The `view.listenTo(collection, 'error', callback)` equivalent equally doesn't strike me as completely horrible for someone who's already being very particular for the sake of performance:

```JavaScript
view.listenToOnce(someCollection.create({...props}, {wait: true}), 'error', callback);
```

jgonggrijp:
> I can see why you would be hesitant to introduce the `this.has(model)` check, especially if it becomes significantly slower the more models are in the collection... 🤔

This is not the case. `has` is implemented using `get`, which in turn uses the internal `_byId` hash table. There is no loop, although it might make a constant number of nested function calls:

https://github.com/jashkenas/backbone/blob/3b37e6aaccbdb8d661487b7cef63e8c930b6f182/backbone.js#L996-L1003

> Does the `_.noop` version negate the `this.has(model)` check?

No. It only prevents worse bugs. It is cheap, though.

> 
> If de-duplicating the event is going to be expensive no matter what, perhaps that's a sign that the status quo is a sensible default behavior. I think the argument could be made that having a collection forward events for a model it doesn't actually own is a nice-to-have rather than a must.

Agreed.

> 
> If it's an expensive nice-to-have, would it be so terrible to hide this type of event-forwarding behind a feature flag? The 'unexpectedness' of the default behavior could simply be documented.

I'm open to either documenting the unexpected behavior or silently fixing it (like this PR attempts to do) (although it should be mentioned in the change log). In my opinion, adding features would be too much honor for a corner case like this.

Rayraz:
The `get` code doesn't look *that* bad to me and people probably shouldn't be adding models in bulk this way anyways if they really care about speed? If you'd really like to silently fix this, I would vote to include the `this.has(model)` fix to avoid double events as well.

If you really don't want to add the `this.has(model)` check, I'd say double events mean you'd just be replacing one unexpected behavior with another arguably more obscure one. In that case documenting the current behavior would be the better alternative in my opinion.

jgonggrijp:
I don't have a strong opinion on the `this.has(model)` check. In order to have one, I would need solid performance measurements, but getting those is a hassle. That would probably not be worth the time.

@alanhamlett have your thoughts on the matter evolved?

paulfalgout:
TBH the `wait` kind of makes it less likely that we can know at the point of adding the error if we should attach the listener.. ie: we might not know what the id _is_ to check against until the server returns it.

backbone.js
```diff
@@ -1079,10 +1079,26 @@
...
+      // In case of wait:true, our collection is not listening to any
+      // of the model's events yet, so it will not forward the error
+      // event. In this special case, we need to listen for it
+      // separately and handle the event just once.
+      // (The reason we don't need to do this for the sync event is
+      // in the success handler above: we add the model first, which
+      // causes the collection to listen, and then invoke the callback
+      // that triggers the event.)
+      if (wait) {
+        forwardPristineError = _.bind(this._onModelEvent, this, 'error');
```
paulfalgout, line 1099: this could be..  (forgive the ES6)
```js
forwardPristineError = (...args) => {
  if (this.has(model) return;
  this._onModelEvent('error', ...args);
}
```
I _think_ this prevents the double error.. it handles when the model doesn't yet have the duplicate id.. _and_ it moves the extra check into the corner-case handler... But I could be missing something?

paulfalgout: well if the model is considered new at add it'd be a different `cid` and then the server can potentially return an existing model id.. so pre-wait `has` would be false, but post `wait` it would be `true`

jgonggrijp: Wait. (No punt intended.) The `cid` is completely independent from newness and server data, no?

Also, if the model is not in the collection beforehand and the request errors, its attributes shouldn't change.

paulfalgout: Ha well you're right on the error part.. since there's an error the attributes won't change, so there's no real issue.. _but_ the `has` checks both `model.cid` and `model.id` so it's possible to have two models with different `cid`s but the same `id` and it won't add the 2nd.. but seeing as the `id` can't change before the `error` event is thrown, it's a mute point.  So I guess the only benefit is delaying the has check until the last necessary moment?  I guess that's worth it...

PR #4269, opened 2023-07-20
Clarify relationships between set, add push and unshift

### LINKED ISSUE #3496: Document Collection push() / unshift() args / return values / events
1. `Backbone.Collection` `push()` and `unshift()` are documented as only accepting a single model, but under the hood they each pass their `model` arg to `add()`, which accepts a model or array as its `models` arg. Are they intended to be restricted to accepting a single model, or can they be documented as accepting an array (that is, accepting the same values that can be passed to `add()` as `models`)?
2. There's no return value documented for those same methods, but they both return the return value from `add()`.
3. Edit: it's undocumented that these trigger an `add` event.

jgonggrijp:
Closes #3496. I moved some of the documentation for `Collection.add` to `Collection.set` and added documentation on sorting, so that the latter now gives a complete overview of the options, return value and events. I added a reference from `Collection.add` to `Collection.set` and from `Collection.{push,unshift}` to `Collection.add`, clarifying how these methods differ from each other.

The updated documentation can be previewed here: https://cdn.statically.io/gh/jgonggrijp/backbone/document-add-push-unshift/index.html.

### DIFF
--- a/index.html
+++ b/index.html
@@ -2005,16 +2005,16 @@ var randomThree = books.sample(3);
     <p id="Collection-add">
       <b class="header">add</b><code>collection.add(models, [options])</code>
       <br />
-      Add a model (or an array of models) to the collection, firing an <tt>"add"</tt>
-      event for each model, and an <tt>"update"</tt> event afterwards. If a <a href="#Collection-model">model</a> property is defined, you may also pass
-      raw attributes objects and options, and have them be vivified as instances of the model using
-      the provided options.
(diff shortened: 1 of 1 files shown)


paulfalgout (approved):
`vivified` is not a term I see a lot, but probably better than instantiated as instances?

jgonggrijp:
Thanks for the review!

No, I don't see "vivified" very often, either. I just moved it from the `add` section to the `set` section; I suppose @jashkenas wrote that in one of his poetic moments.

PR #4270, opened 2023-07-20
Document Model.cidPrefix

### LINKED ISSUE #4226: Models are skipped or merged when idAttribute values look like cid (c12, c78, etc)
Hi there!

We're having a strange issue when using a collection of models with custom `idAttribute`. Basically, if `idAttribute` attribute has a value that "looks like" `cid` (**c12**, **c78**, etc), the model is not added to collection or merged with some previously added model.

Please, check the following simplified example:

```
const MyModel = Backbone.Model.extend({
    idAttribute: 'myId'
});

const MyCollection = Backbone.Collection.extend({
    model: MyModel
});

const list = [
  { myId: 'c10', name: 'John' },
  { myId: 'c9', name: 'James' },
  { myId: 'c8', name: 'William' },
  { myId: 'c7', name: 'Henry' },
  { myId: 'c6', name: 'Michael' },
  { myId: 'c5', name: 'Fred' },
  { myId: 'c4', name: 'George' },
  { myId: 'c3', name: 'Larry' },
  { myId: 'c2', name: 'Peter' },
  { myId: 'c1', name: 'Neil' }
];

const testCollection = new MyCollection(list);

alert(testCollection.length); // expected: 10, actual 5
```

Or check this fiddle: https://jsfiddle.net/0goy4u9w/8/

I believe this happens because Backbone stores models both by `cid` and `idAttribute`, when available, here:
https://github.com/jashkenas/backbone/blob/75e6d0ce6394bd2b809823c7f7dc014ddb6ae287/backbone.js#L1178-L1183
Then, during `set()` it finds the previously stored (by `cid`) model and assumes it's found a duplicate:
https://github.com/jashkenas/backbone/blob/75e6d0ce6394bd2b809823c7f7dc014ddb6ae287/backbone.js#L870-L877

Now, I understand that this looks like someone's tried really hard to find a defect 😬and created ids that look like cids, but we actually use custom id attributes that are unique, but don't follow same pattern (they come from different sources), and sometimes they actually look like cid. 😱

Is there a reason why `cid` is still being used as key for `_byId` even when models have `idAttribute` declared?

jgonggrijp:
Closes #4226. The updated documentation can be previewed at https://cdn.statically.io/gh/jgonggrijp/backbone/document-cidPrefix-conflict/index.html.

### DIFF
--- a/index.html
+++ b/index.html
@@ -356,6 +356,7 @@
         <li data-name="id">– <a href="#Model-id">id</a></li>
         <li data-name="idAttribute">– <a href="#Model-idAttribute">idAttribute</a></li>
         <li data-name="cid">– <a href="#Model-cid">cid</a></li>
+        <li data-name="cidPrefix">– <a href="#Model-cidPrefix">cidPrefix</a></li>
         <li data-name="attributes">– <a href="#Model-attributes">attributes</a></li>
         <li data-name="changed">– <a href="#Model-changed">changed</a></li>
         <li data-name="defaults">– <a href="#Model-defaults">defaults</a></li>
--- a/index.html
+++ b/index.html
@@ -1298,6 +1298,8 @@ if (note.has("title")) {
       If you're directly communicating with a backend (CouchDB, MongoDB) that uses
       a different unique key, you may set a Model's <tt>idAttribute</tt> to
       transparently map from that key to <tt>id</tt>.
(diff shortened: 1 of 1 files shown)


jgonggrijp:
CC @jSanchoDev @GammaGames

paulfalgout (approved):

PR #4272, opened 2023-07-21
Add a .editorconfig; prepare for switching to 4-space indentation

jgonggrijp:
I've been keeping an untracked `.editorconfig` for the past couple of days, because I usually use 4-space indentation while Backbone uses 2-space indentation.

I plan to modularize Backbone and grasp the opportunity to switch to 4-space indentation, since it is a major code disruption anyway. This will be easier to manage between different branches if I commit the `.editorconfig`.

If somebody thinks I should **not** change the indentation, now is the time to tell me! (Don't worry, I will keep different types of changes in separate commits so they are easier to trace.)

There is currently a subsection for `package.json` because that will have to stick to 2-space indentation. I don't mind removing the special case for the time being. Other comments on details are also welcome.

### DIFF
--- a/.editorconfig
+++ b/.editorconfig
@@ -1,0 +1,17 @@
+# EditorConfig helps developers define and maintain consistent
+# coding styles between different editors and IDEs
+# editorconfig.org
+
+root = true
+
+
+[*]
+
+indent_style = space
+indent_size = 2
+charset = utf-8
+insert_final_newline = true
+trim_trailing_whitespace = true
+
+[**/package.json]
+indent_size = 2


paulfalgout:
I don't _love_ changing the spacing, but as long as there's an `.editorconfig` and the commits are followable, I'm ok with it

jgonggrijp:
Thanks for thinking along. Do you "not love" changing the spacing mostly because of the disruption, or do you also prefer 2-space indents over 4-space indents in general?

jgonggrijp:
Merging, but still interested in your answer.

paulfalgout:
Hmm I don't know if I have a strong preference.. most of the projects I work on are 2-space.. I think ironically because we switched to Bb and Marionette all of which were 2.  As long as theirs an editorconfig though it's NBD.  But yeah I don't love the history of the change in the blames, but it's not _that_ big of a deal.. clearly less of a big deal with a modular split.

jgonggrijp:
The reason I want to switch to 4-space is that I find it much clearer what goes with what. I'm now a bit hesitant to go through with it, though, knowing that you don't love it.

I would appreciate more opinions; everyone watching, please consider making a reply! I'll also CC @GammaGames as you indicated you are OK with that.

paulfalgout:
meh, I'll help where I can, but you're doing a lot of the heavy lifting. Whatever is good on your eyes is ok with me.. as long as there's an editorconfig it's no big deal.

GammaGames:
I'm all for it! I also like the current `insert_final_newline` setting, if you want another suggestion I'd add the equivalent to VSCode's `files.trimFinalNewlines` setting to prevent excessive blank lines at the ends of files.

jgonggrijp:
@GammaGames "all" includes 4-space indents?

GammaGames:
Yeah, I don't have strong feelings but I do prefer 4-space because it makes the left gutter feel less cramped.

PR #4276, opened 2023-07-26
Code of conduct

jgonggrijp:
I copied [the code of conduct that @jashkenas customized for Underscore](https://github.com/jashkenas/underscore/blob/master/CODE_OF_CONDUCT.md) and took the diff with the original, [version 1.4](https://www.contributor-covenant.org/version/1/4/code-of-conduct/) of the [Contributor Covenant CoC](https://www.contributor-covenant.org/). I then downloaded [the latest version, 2.1](https://www.contributor-covenant.org/version/2/1/code_of_conduct/), and merged Jeremy's customizations into that by my best judgement.

Like in Underscore, I used Jeremy's email address as the contact option. I did not discuss this with him (sorry), but my guess is that he still feels end responsible for the project and that he would prefer this. I also think that if anyone at any point feels uncomfortable or unsafe, they will likely have interacted more with me than with Jeremy (because I'm currently more active in the project), so Jeremy may be perceived as a more objective, safer arbiter.

Compared to version 1.4, version 2.1 of the CoC comes with a new section on enforcement guidelines. I left in the original text, but tweaking or even removing it is an option. Suggestions welcome.

This type of thing is supposed make people feel safe, so if it does the opposite for you, or if you are concerned that it might do so for other people, please discuss. CC @GammaGames

### DIFF
--- a/CODE_OF_CONDUCT.md
+++ b/CODE_OF_CONDUCT.md
@@ -1,0 +1,132 @@
+# Contributor Covenant Code of Conduct
+
+## Our Pledge
+
+We as members, contributors, and leaders pledge to make participation in our
+community a harassment-free experience for everyone, regardless of age, body
+size, visible or invisible disability, ethnicity, sex characteristics, gender
+identity and expression, level of experience, education, socio-economic status,
+nationality, personal appearance, race, caste, color, religion, or sexual
+identity and orientation.
+
+We pledge to act and interact in ways that contribute to an open, welcoming,
+diverse, inclusive, and healthy community.
+
+## Our Standards
+
+Examples of behavior that contributes to a positive environment for our
+community include:
+
+* Demonstrating empathy and kindness toward other people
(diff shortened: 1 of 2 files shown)


GammaGames:
LGTM, seems pretty standard 👍️

paulfalgout:
I've never really been apart of a repo that's ever really had a conduct issue, so I don't spend a lot of time thinking about this.  In my experience the document isn't necessary, but for my part, feel free.

jgonggrijp:
I have never experienced that either, and I hope I never will. However, I do think that if it ever comes up, it will be better to have one than not.

Inclusion-promoting organizations (such as the former Ada Initiative) tend to attach a lot of value to the presence and content of a code of conduct. I consider that evidence that it matters more to the less-privileged than we might be aware.

Secondary, but also important, is that tools like Tidelift and Snyk tend to emphasize the code of conduct as well.

Issue #4205, opened 2018-10-13
there is a way to add several models with "same" id to a collection


taburetkin:
default normal behavior:
```js
const coll = new Backbone.Collection();
coll.add({ id: 1 });
coll.add({ id: 1 });
coll.add({ id: 1 });
console.log(coll.lenght); //1
```

but, if we did that with model having id in defaults then all models will be added to a collection

```js
const Model = Backbone.Model.extend({
  defaults: { id: 0 }
});
const coll = new Backbone.Collection([], { model: Model });
coll.add({  });
coll.add({  });
coll.add({  });
console.log(coll.length); // 3
```
working fiddle: https://jsfiddle.net/gkpLoaqx/

maybe id field from defaults should not be applied on model initialize?

prma85:
I was looking at it. Logically, you shouldn't be able to add more than one item with the same id, since the ID is unique. When you are sending the id attribute, the Backbone verifies if the id already exists or not before trying to add it. If it exists, nothing is done. The problem is that if you don't send this attribute, the Backbone does not do this verification and then, create the models with the default values, creating the bug.

To fix it, need to do the verification when creating the model and not when it is doing the `add` function.

taburetkin:
i suppose, that attribute defined in `idAattribute` for a model can not be part of `defaults` because of the id nature .
and i think that defaults should be transformed before applying to models attribute, something like this `_.omit(defaults, this.idAttribute)`
probably this solution is not good enough in performance terms.

i can suggest this small solution
link to model constructor:
https://github.com/jashkenas/backbone/blob/27f7d41de1f64f6662a96f44531d7518953b1e07/backbone.js#L406

i think, that this part can be replaced with this few lines
```
    var defaults = this.getDefaults();
    attrs = _.defaults(_.extend({}, defaults, attrs), defaults);
    ...
},
getDefaults() {
    var defaults = _.result(this, 'defaults');
    delete defaults[this.idAttribute]; // or defaults = _.omit(defaults, this.idAttribute);
    return defaults;  
}
```

jgonggrijp:
I've been investigating this issue, and I realized that the problem is not in Backbone, but in the hypothetical user violating the preconditions of the framework.

[...]

PR #4279, opened 2023-07-27
Add warning to documentation about setting the model id in the defaults

jgonggrijp:
Please see #4205 for a rationale of this change. My thoughts that led to this PR are in the bottom comment.

I have asked myself whether this should be documented at all, given that it is a rather obscure corner case (user sets the model id to a fixed value in the defaults and also adds instances of the model to a collection). However, the fact that #4205 was reported proves that users may encounter the situation.

The edit can be previewed as web page over here: https://cdn.statically.io/gh/jgonggrijp/backbone/document-defaults-idAttribute/index.html#Model-defaults.

### DIFF
--- a/index.html
+++ b/index.html
@@ -1405,14 +1405,13 @@ alert("Dessert will be " + (new Meal).get('dessert'));
     </p>
 
     <p class="warning">
-      In nearly all cases, the defaults should <b>not</b> set a value for the
-      model&rsquo;s <a href="#Model-idAttribute"><tt>idAttribute</tt></a>. Doing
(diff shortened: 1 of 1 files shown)


paulfalgout (changes requested):
I think this may be too harsh?

With uuids it seems plausible to me that an app could set a default id.  They're so common these days.. So I think it should be flipped somewhat something like:

If you are setting the default id of a model on a collection, then it must be universally unique.

Then the warning about duplicate ids.

jgonggrijp:
@paulfalgout Better now?

paulfalgout (approved):

PR #4284, opened 2024-01-10
Fix wrong context passed in collection.create

olhapi:
Here https://github.com/jashkenas/backbone/blob/master/backbone.js#L1088
`this` has the wrong context, it refers to a window object in the browser because we didn't pass a previously saved collection context to it.


olhapi:
@jgonggrijp can you check this and confirm that it's indeed a bug?

jgonggrijp (changes requested):
I agree this is a bug and I also agree with your fix. 👍 

This bug needs a regression test, could you add it?

In any case, thanks for reporting and fixing!

olhapi:
I've tried that, but I don't know how to test that case properly. My test succeeded when it should have failed. :)

jgonggrijp:
Alright, in that case I will do it myself some time in the near future.

olhapi:
Thanks a lot

jgonggrijp:
@olhapi Please remove the changes to `backbone-min*` and then force-push the branch. We update the minified code when we cut a new release.

(I'm working on the regression test in the meanwhile.)

olhapi:
@jgonggrijp done, just pushed.

jgonggrijp:
@olhapi The commit referenced above (https://github.com/jashkenas/backbone/commit/0f9956b565ca7d08421bdb3f9b2e583b38abc659) adds a test that fails without your fix and succeeds with it. Again thanks for contributing!

PR #4287, opened 2024-02-23
Use allowed instead of whitelisted; Use disallowed instead of blacklisted;

xwxtwd:
1.Use allowed instead of whitelisted.
2.Use disallowed instead of blacklisted.


xwxtwd:
Done.😀

jgonggrijp (approved):
Thanks!

PR #4288, opened 2024-03-06
Bump GitHub workflows to latest version

deining:
This PR bump GitHub workflows to their latest versions, thus avoiding deprecation warnings as seen [here](https://github.com/jashkenas/backbone/actions/runs/8035882913).


jgonggrijp (approved):
Thanks!

PR #4292, opened 2024-09-02
Trigger the "changeId" event only if prevId is different from this.id.

Noodlex:


backbone.js
```diff
@@ -523,7 +523,9 @@
       if (this.idAttribute in attrs) {
         var prevId = this.id;
         this.id = this.get(this.idAttribute);
-        this.trigger('changeId', this, prevId, options);
+        if(this.id !== prevId) {
```
jgonggrijp, line 526: There needs to be a space here for consistency's sake.

```suggestion
        if (this.id !== prevId) {
```

jgonggrijp (changes requested):
Thank you for taking the effort to submit a fix, @Noodlex!

Could you please add a test that fails without your change, but passes with your change? Thanks in advance.

jgonggrijp (approved):

jgonggrijp:
Thanks again!

jgonggrijp:
Released in version 1.6.1. Sorry for the delay!

PR #4293, opened 2024-09-25
Trigger "changeId" only if content change

Noodlex:
Trigger "changeId" if the content is modified.

Null or undefined are identical in the process.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -523,7 +523,7 @@
       if (this.idAttribute in attrs) {
         var prevId = this.id;
         this.id = this.get(this.idAttribute);
-        if (this.id !== prevId) {
+        if (!_.isEqual(prevId, this.id) && !((_.isNull(this.id) || _.isUndefined(this.id)) && (_.isNull(prevId) || _.isUndefined(prevId)))) {
           this.trigger('changeId', this, prevId, options);
         }
       }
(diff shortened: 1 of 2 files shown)


jgonggrijp:
Could you clarify why you think `null` and `undefined` should be considered equivalent? This is different from the semantics of the regular `change` and `change:[attribute]` events.

Noodlex:
It's a proposal.

I have a case where I have a collection in a model.
I receive the complete model with the collection and the new attributes.
These are different and in some cases, I end up with a “prevId” = undefined and a this.id = null.

The behavior seems the same to me in the case of avoiding triggering the “changeId” event, but I'm well aware that undefined is different from null.

jgonggrijp:
> I have a case where I have a collection in a model. I receive the complete model with the collection and the new attributes. These are different and in some cases, I end up with a “prevId” = undefined and a this.id = null.

To be honest, to me this seems like a situation one would rather avoid in the first place. I am not convinced yet, but maybe you can show me some example code?

Noodlex:
I don't yet have time to investigate the source of the problem. Currently, this fix on my version allows me to avoid “changeId” trigger problems.

I'll try when I'm less in a rush to find the exact source and maybe add the right fix.

I'll close this request and come back later if needed.

Thanks for your time.

PR #4295, opened 2025-03-31
Prepare 1.6.1

jgonggrijp:
I want to publish 1.6.1 on April 1. I would appreciate a review, but I will publish even without one. This PR and its changes will still be visible after the fact, and a review will still be welcome.

This branch includes all changes that will be included in the 1.6.1 release, except for the autogenerated ones, i.e.:

- the minified bundle;
- the annotated sources;
- the `MARK_RELEASE` in `modules/debug-info.js`.

If you feel that these changes should be part of a review as well, please leave a comment.

CC @GammaGames @Yahasana


paulfalgout (approved):

jgonggrijp:
Thank you @paulfalgout  🙏

Issue #4271, opened 2023-07-21
Testing against outdated vendor libraries


jgonggrijp:
The project is currently testing against local hardcopies of jQuery, json2, QUnit, require.js and Underscore. These versions are all outdated by at least seven years.

There is a real advantage to this: testing against old versions of dependencies safeguards backwards compatibility.

There is also an obvious disadvantage: the more outdated these dependencies become, the more likely it is that incompatibilities creep into newer versions, without the tests detecting them.

In my own projects, I use Backbone with the latest versions of Underscore and jQuery, and I haven't noticed any problems. So far, we might have gotten away with the outdatedness of the dependencies, thanks to the fact that jQuery and Underscore are both very stable. However, at some point, this streak of luck will likely end.

I am convinced that we need to test against recent versions of jQuery and Underscore. They can simply be sourced from the `node_modules` instead of `test/vendor`. I am less sure about the old versions; should we keep *also* testing against those or not?

Thoughts appreciated. CC @GammaGames

GammaGames:
Wow, that version of jquery released almost 11 years ago! I've been using recent versions in our projects and haven't come across any issues, underscore at 1.13.2 and jquery at 3.6.3. I don't really see any value in testing against versions that old.

PR #4298, opened 2025-04-16
Upgrade underscore dependency to fix CVE-2021-23358

rboonzaijer:
Upgrade underscore dependency (from =>1.8.3 to =>1.12.1) to fix CVE-2021-23358

Vulnerability details: https://nvd.nist.gov/vuln/detail/CVE-2021-23358
Fixed in underscore 1.12.1: https://underscorejs.org/#1.12.1

**/test/vendor/underscore.js** == https://cdn.jsdelivr.net/npm/underscore@1.12.1/underscore-esm.js

**Notes:**

- I'm not sure how to run the tests locally, let's see if they succeed in the pipeline.
- I did not check if every method from underscore has a test, and if it should have one (or does backbone need a major release if it could be a breaking change? -> in that case: perhaps use the latest underscore version 1.13.7 instead of 1.12.1?)

### DIFF
--- a/index.html
+++ b/index.html
@@ -629,7 +629,7 @@
 
     <p>
       Backbone's only hard dependency is
-      <a href="http://underscorejs.org/">Underscore.js</a> <small>( >= 1.8.3)</small>.
+      <a href="http://underscorejs.org/">Underscore.js</a> <small>( >= 1.12.1)</small>.
       For RESTful persistence and DOM manipulation with <a href="#View">Backbone.View</a>,
       include <b><a href="https://jquery.com/">jQuery</a></b> ( >= 1.11.0).
       <i>(Mimics of the Underscore and jQuery APIs, such as
@@ -4553,6 +4553,16 @@ ActiveRecord::Base.include_root_in_json = false
     </div>
 
     <h2 id="changelog">Change Log</h2>
+
+    <b class="header" id="1.6.2">1.6.2</b> &mdash; <small><i>Apr. 16, 2025</i></small>
+    &mdash; <a href="https://github.com/jashkenas/backbone/compare/1.6.1...1.6.2">Diff</a>
+    &mdash; <a href="https://cdn.statically.io/gh/jashkenas/backbone/1.6.2/index.html">Docs</a>
+    <br />
+    <ul style="margin-top: 5px;">
+      <li>
+        Upgrade underscore dependency to fix <a href="https://nvd.nist.gov/vuln/detail/CVE-2021-23358">CVE-2021-23358</a>
+      </li>
+    </ul>
 
     <b class="header" id="1.6.1">1.6.1</b> &mdash; <small><i>Apr. 1, 2025</i></small>
     &mdash; <a href="https://github.com/jashkenas/backbone/compare/1.6.0...1.6.1">Diff</a>
--- a/bower.json
+++ b/bower.json
@@ -2,7 +2,7 @@
   "name"          : "backbone",
   "main"          : "backbone.js",
   "dependencies"  : {
-    "underscore"  : ">=1.8.3"
+    "underscore"  : ">=1.12.1"
   },
   "ignore"        : ["docs", "examples", "test", "*.yml", "*.html", "*.ico", "*.md", "CNAME", ".*", "karma.*", "package.json"]
 }
--- a/package-lock.json
+++ b/package-lock.json
@@ -9,7 +9,7 @@
       "version": "1.6.1",
       "license": "MIT",
       "dependencies": {
-        "underscore": ">=1.8.3"
+        "underscore": ">=1.12.1"
       },
       "devDependencies": {
         "coffeescript": "^2.7.0",
--- a/package.json
+++ b/package.json
@@ -13,7 +13,7 @@
   ],
   "author": "Jeremy Ashkenas",
   "dependencies": {
-    "underscore": ">=1.8.3"
+    "underscore": ">=1.12.1"
   },
   "devDependencies": {
     "coffeescript": "^2.7.0",
--- a/test/vendor/underscore.js
+++ b/test/vendor/underscore.js
@@ -1,1548 +1,2026 @@
-//     Underscore.js 1.8.3
-//     http://underscorejs.org
-//     (c) 2009-2015 Jeremy Ashkenas, DocumentCloud and Investigative Reporters & Editors
+//     Underscore.js 1.12.1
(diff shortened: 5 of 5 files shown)


index.html
```diff
@@ -4554,6 +4554,16 @@ <h2 id="examples-tzigla">Tzigla</h2>
...
+    <b class="header" id="1.6.2">1.6.2</b> &mdash; <small><i>Apr. 16, 2025</i></small>
+    &mdash; <a href="https://github.com/jashkenas/backbone/compare/1.6.1...1.6.2">Diff</a>
+    &mdash; <a href="https://cdn.statically.io/gh/jashkenas/backbone/1.6.2/index.html">Docs</a>
+    <br />
+    <ul style="margin-top: 5px;">
+      <li>
+        Upgrade underscore dependency to fix <a href="https://nvd.nist.gov/vuln/detail/CVE-2021-23358">CVE-2021-23358</a>
+      </li>
+    </ul>
+
```
jgonggrijp, line 4566: Side note: please do not add change log entries, change the Backbone version number or update generated files such as the annotated source. We do those things when we cut a release.

jgonggrijp (commented):
Thank you for investing effort in this, @rboonzaijer.

I'm afraid your contribution is based on a misunderstanding of version numbers and some of the administration surrounding it, as well as of the CVE in question and the way it affects the tests (it actually doesn't).

> Upgrade underscore dependency (from =>1.8.3 to =>1.12.1) to fix [CVE-2021-23358](https://github.com/advisories/GHSA-cf4h-3jhx-xvhq)

`=> 1.8.3` means that Backbone is compatible with all Underscore versions in the 1.x series, from 1.8.3 all the way up to the current 1.13.7, including 1.12.1. The CVE does not change anything about this compatibility.

> 
> **/test/vendor/underscore.js** == https://cdn.jsdelivr.net/npm/underscore@1.12.1/underscore-esm.js

This breaks the tests, not because it is version 1.12.1 but because you used the ESM bundle. Backbone's tests still rely on UMD.

> 
> **Notes:**
> 
>     * I'm not sure how to run the tests locally, let's see if they succeed in the pipeline.

The easiest way is to open `test/index.html` in a browser.

> 
>     * I did not check if every method from underscore has a test, and if it should have one

It does, but those [tests](https://github.com/jashkenas/underscore/tree/master/test) are in the Underscore repository. The Backbone repository does not need to have tests for Underscore functions, it's here in the `vendor` directory because Backbone and its tests use Underscore under the hood.

> (or does backbone need a major release if it could be a breaking change? -> in that case: perhaps use the latest underscore version 1.13.7 instead of 1.12.1?)

Versions break down into three numbers: major.minor.patch. So for example in 1.8.3, 1 is the major version, 8 is the minor version and 3 is the patch version. Version numbers with identical major version are backwards compatible. So generally, if a library releases a version 1.13.7, users should be able to use it instead of version 1.12.1 without anything breaking. The same is true with 1.12.1 and 1.8.3.

From this, you can draw two conclusions:

1. If Backbone were to upgrade its Underscore dependency from 1.8.3 to a later version in the 1.x series, this should never be a breaking change.
2. However, there is no need to upgrade Backbone's dependency, because end users can combine Backbone with the latest Underscore *already*.

If we say that Backbone depends on Underscore => 1.12.1, or => 1.13.7, then we are effectively communicating that Backbone *will not work* if you use an earlier version of Underscore. This is not true. Backbone will work fine with any version from 1.8.3 onwards. On the other hand, a dependency on => 1.8.3 communicates that Backbone will *also* work with Underscore 1.12.1, Underscore 1.13.7 and any future 1.x versions of Underscore.

If we stop testing against 1.8.3, then we are no longer verifying that this combination actually works. You might argue that we shouldn't, because 1.8.3 has this vulnerability, but the GitHub advisory is heavily overstating the importance of the vulnerability (9.8 out of 10). I know because [I handled it myself](https://github.com/jashkenas/underscore/pull/2917). See for example the [Snyk advisory](https://snyk.io/vuln/SNYK-JS-UNDERSCORE-1080984), which gives it a much milder 5.5. The CVE only affects projects for which all three of the following conditions are true:

1. Some code within the project calls the `_.template` function.
2. The project does not explicitly pass the `variable` option to the template, or uses unsanitized user input for this purpose.
3. The result of template compilation is sent to a remote JavaScript environment for execution.

The latter point is not true for the Backbone tests. Backbone itself does not use `_.template` at all, which means that Backbone is completely unaffected by this CVE. Most Backbone users probably don't even use `_.template` either, which means that the CVE is simply not a consideration for them. If they want to use Backbone with an older version of Underscore, they can safely do so.

Now, this does not mean that we absolutely *have* to test against 1.8.3, just that there isn't a pressing reason not to do it. Of course, there would be something to say for (also) testing with the latest version of Underscore, since we want to be compatible with that, too. I opened #4271 because of this. However, updating the hardcopy in the `test/vendor` directory is not the right approach to solving that issue, in my opinion; I would rather drop the hardcopy altogether and use a dynamically installed version from the `node_modules` instead, or perhaps multiple versions.

Long story short, I will be closing your pull request. Please feel welcome to continue discussion over here, and/or to submit another pull request if you want to try your hand at #4271.

jgonggrijp:
By the way, what prompted you to start working on this? Did you get a warning from a tool? Just upgrading your local copy of Underscore *should* address that; otherwise, the tool is misbehaving.

rboonzaijer:
> By the way, what prompted you to start working on this? Did you get a warning from a tool? Just upgrading your local copy of Underscore _should_ address that; otherwise, the tool is misbehaving.

Thank you for your fast and extensive answer, and I see now I've jumped to conclusions far too quickly, thinking an old version was used here, but instead it's in another repo.

And you are so right about the versions (starting with ^), I was not paying attention I guess...

I was trying the beta version of the test-reporting tool 'Allure 3' this afternoon, but when I scanned it with trivy the vulnerability popped up, I quickly looked at the versions and thought that it came from backbone, but it's not (as you pointed out, and I see now), it's Allure3 that references an old version of backbone.marionette, and that old version references the old underscore.

A quick test setup:
https://gitlab.com/rboonzaijer-user/dependency-test-allure3
https://gitlab.com/rboonzaijer-user/dependency-test-allure3/-/jobs/9746497832#L471

The trace:
https://github.com/allure-framework/allure3/blob/main/packages/web-allure2/package.json#L38
https://github.com/marionettejs/backbone.marionette/blob/v3.5.1/package.json#L90

Thanks again for looking into this, I will report this in the other repo when I have time :)

PR #4300, opened 2026-04-06
Support ES6 method-shorthand constructors in extend (#4213)

### LINKED ISSUE #4213: Support ES6 method definitions for constructors
Using ES6, it becomes possible to extend Backbone objects using
```js
const MyModel = Backbone.Model.extend({
  foo() {
    return 'bar';
  }
});
```
This syntax is a nice shortcut for the more old-fashioned
```js
const MyModel = Backbone.Model.extend({
  foo: function() {
    return 'bar';
  }
});
```
The problem is though that if you want to specify a custom constructor
```js
const MyModel = Backbone.Model.extend({
  constructor() {
    Backbone.Model.apply(this, arguments);
  }
});
```
it turns out that
```js
let model = new MyModel();
```
throws an ```Uncaught TypeError: MyModel is not a constructor```.

The reason behind this behavior is that the ES6 method definition ```name()``` is not completely equivalent to the old ```name: function() {}``` because the ES6 methods are labelled internally as non-constructable.

A way to solve this is to change the extend method from
```js
if (protoProps && _.has(protoProps, 'constructor')) {
  child = protoProps.constructor;
} else {
  child = function(){ return parent.apply(this, arguments); };
}
```
to
```js
if (protoProps && _.has(protoProps, 'constructor')) {
  var ctor = protoProps.constructor;
  child = function(){ return ctor.apply(this, arguments); };
} else {
  child = function(){ return parent.apply(this, arguments); };
}
```

There is one minor drawback to this approach which is that if someone ever did something like
```js
const proto = {
  constructor: function() {}
};
const Model = Backbone.Model.extend(proto);
if (Model === proto.constructor) { ... }
```
this would no longer work. I tried to find whether it is possible to check whether a method is constructable or not in which case something could be done like
```js
if (protoProps && _.has(protoProps, 'constructor')) {
  var ctor = protoProps.constructor;
  if (!isConstructable(ctor)) {
    child = function(){ return ctor.apply(this, arguments); };
  } else {
    child = ctor;
  }
} else {
  child = function(){ return parent.apply(this, arguments); };
}
```
but it seems that [there's no elegant way to check this](https://stackoverflow.com/questions/40922531/how-to-check-if-a-javascript-function-is-a-constructor).

Hence it's a trade-off between 100% backwards compatability for a rather rare edge case, or to be able to get rid of the requirement to specify the constructor as a non-ES6 method definition, but this leads to some awkward code like
```js
const Model = Backbone.Model.extend({
  constructor: function() { ... },
  method() { ... }
});
```

I would love to hear some thoughts on this subject. I'd be willing to create a pull request if required.

BigBalli:
Fixes #4213.

## Problem

ES6 method shorthand (`constructor() {}`) produces a non-constructable
function with no `prototype` property. Passing such a constructor to
`extend` previously caused `new Subclass()` to throw
`Subclass is not a constructor`:

```js
const MyModel = Backbone.Model.extend({
  constructor() {
    Backbone.Model.apply(this, arguments);
  }
});
new MyModel(); // TypeError: MyModel is not a constructor
```

This forces users on modern JS to write the awkward mixed form
(`constructor: function() {}` next to method shorthand) or to give up on
`extend` altogether.

## Fix

Detect the shorthand case by checking for the absence of `prototype` on
the supplied constructor, and in that case wrap it in a thin function so
`new` continues to work. The traditional `constructor: function() {}`
form is unchanged and still uses the supplied function directly, so the
edge case noted in the issue (`Model === proto.constructor`) remains
true for ES5 users.

## Test

Added a regression test in `test/model.js` that defines the
`constructor` via ES6 method shorthand (constructed via `eval` so the
test file still parses on engines without shorthand support) and
verifies that `new Model({foo: 'bar'})` returns a usable instance.

`npm run lint` passes.


backbone.js
```diff
@@ -2107,9 +2107,16 @@
...
     // The constructor function for the new subclass is either defined by you
     // (the "constructor" property in your `extend` definition), or defaulted
-    // by us to simply call the parent constructor.
+    // by us to simply call the parent constructor. ES6 method shorthand
+    // (`constructor() {}`) produces a non-constructable function with no
+    // `prototype`; in that case we wrap it so `new` still works.
     if (protoProps && _.has(protoProps, 'constructor')) {
-      child = protoProps.constructor;
+      var supplied = protoProps.constructor;
+      if (supplied.prototype) {
```
jgonggrijp, line 2115: I can imagine making this check a bit more precise, in one of the following ways:


```suggestion
      if (_.has(supplied, 'prototype')) {
```

```suggestion
      if (typeof supplied.prototype === 'object') {
```

(combination of both)

```suggestion
      if (_.has(supplied, 'prototype') && typeof supplied.prototype === 'object') {
```

Can we think of situations where the distinction might be important enough that one of the more precise versions might be preferred? I can't quite put my finger on it, but I have a feeling that it might be important that the `prototype` is an own property of `constructor`, which would warrant the `_.has` check.

CC @GammaGames @Yahasana

BigBalli: Good catch. Would be good to switch to `_.has(supplied, 'prototype')`.

Walking the realistic inputs (classic functions, classes, shorthand, arrow/async/generator), the truthy check and `_.has` behave identically. Own-vs-inherited only diverges if someone does `Object.setPrototypeOf` on a shorthand method, which isn't a real user path. But `_.has` expresses the intent directly ("supplied a function with its own prototype slot") instead of leaning on truthiness as a proxy, so it's the better line at the same cost.

We could skip the `typeof === 'object'` variant. The only way to fail it with a truthy `prototype` is manual assignment like `Fn.prototype = 42`, which is sabotage either way.

jgonggrijp: Alright, let's change it to `_.has`.

Issue #2566, opened 2013-05-28
Inconsistent fragment coding  across browsers


dominictobias:
Using: Backbone 1.0.0
Firefox: Latest stable version (and plenty before)
Related Issues: https://github.com/documentcloud/backbone/pull/967
https://github.com/documentcloud/backbone/pull/1426

In the related issues it was decided that Backbone should handle fragment coding inconstantly causing fundamentally different code execution across browsers and I cannot fathom why.

For example in my app when a search term is typed in Chrome, Opera, iOS etc:

`search/searching for something`

The search path `search/*` is executed once and does not re-execute.

In Firefox however:

`search/searching for something`

Once the first spacebar has been pressed every single character afterwards causes a re-execution of the route. Causing in my app the entire layout to be re-rendered each time and the search looses focus on each keystroke.

Why it was decided to encode only in Firefox, and decode in every other browser I cannot understand?

The same code execution across browsers can be achieved by adding back `decodeURIComponent` to `match[1]` in this odd function (it was removed).

``` javascript
// Gets the true hash value. Cannot use location.hash directly due to bug
// in Firefox where location.hash will always be decoded.
getHash: function(window) {
  var match = (window || this).location.href.match(/#(.*)$/);
  return match ? match[1] : '';
},
```

Also the comment explains the exact opposite of the issue on firefox - in firefox the space character IS encoded and thus needs to be decoded to create consistency with other browsers.

Is someone able to explain why this was done?

Thanks

PR #4301, opened 2026-04-06
Preserve percent-encoded trailing whitespace in fragments (#4198)

### LINKED ISSUE #4198: Trailing Fragment Space(s)
We have fragments that have a space at the end (outbound/22130600/po/powithspacetest%20).  When trying to navigate to this fragment Backbone is stripping the trailing space in getFragment().  

```
        var M = /^[#\/]|\s+$/g;

        getFragment: function(t) {
            if (t == null) {
                if (this._usePushState || !this._wantsHashChange) {
                    t = this.getPath()
                } else {
                    t = this.getHash()
                }
            }
            return t.replace(M, "")
        },
```

I can understand removing the bits at the start of the fragment, but can someone explain the motivation for removing the space(s) at the end?  This causes our app to fail to lookup the requested data because the part with the space is a param in the router.

```
routes: {
                "": "home",
                "outbound/:tieBack/po/:poNbr": "outboundDetail"
            },
```

And

```
            outboundDetail: function(tieBack, poNbr) {
                app.switchOutboundDetail(tieBack, poNbr);
            },
```

BigBalli:
Fixes #4198.

## Problem

`routeStripper` (`/^[#\/]|\s+$/g`) was applied at the very end of
`getFragment()`, *after* `getPath()` has already decoded the URL via
`decodeFragment()`. As a result, an encoded trailing space (`%20`) was
decoded to a literal space and then stripped, silently truncating
legitimate parameter values:

```
/outbound/22130600/po/powithspacetest%20
                                       ^^ stripped, lookup fails
```

## Why the strip exists

The trailing-space strip was introduced in #1794 to work around old
browsers where `location.hash` itself returned trailing whitespace. That
fix only ever needed to apply to the raw hash value — not to the
already-decoded path.

## Fix

- Split `routeStripper` into two regexes:
  - `routeStripper = /^[#\/]/` — only the leading hash/slash, applied
    in `getFragment()` as before.
  - `trailingSpaceStripper = /\s+$/` — applied inside `getHash()` on
    the raw hash value, preserving the original #1794 fix at the level
    where it actually matters.
- `getFragment()` no longer touches trailing whitespace, so explicit
  fragments and decoded paths preserve their trailing characters.

## Tests

- Updated the existing `#1794` test to assert against `getHash()` (the
  correct level), and added an assertion that `getFragment('fragment   ')`
  preserves whitespace.
- Updated the existing `#1820` test (leading slash + trailing space) to
  reflect that only the leading slash is stripped now.
- Added a new regression test for `#4198` covering a percent-encoded
  trailing space in the path.

`npm run lint` passes.


test/router.js
jgonggrijp, line 676: This is changing the meaning of an existing test. I know this has been done a lot in the past history of Backbone, but I think it is fundamentally wrong. Either, you preserve a test and make sure that the code still passes it in the face of new changes, or you remove a test entirely if you can defend that it was testing for the wrong thing. In the latter case, you replace it by a different test that more accurately tests what should be tested.

test/router.js
```diff
@@ -660,16 +660,35 @@
...
     var history = new Backbone.History;
-    assert.strictEqual(history.getFragment('fragment   '), 'fragment');
+    assert.strictEqual(history.getFragment('/fragment '), 'fragment ');
   });
 
-  QUnit.test('#1820 - Leading slash and trailing space.', function(assert) {
+  QUnit.test('#4198 - Percent-encoded trailing space preserved in path.', function(assert) {
     assert.expect(1);
     var history = new Backbone.History;
-    assert.strictEqual(history.getFragment('/fragment '), 'fragment');
```
jgonggrijp, line 672: Again, this is changing the meaning of an existing test.

test/router.js
```diff
@@ -660,16 +660,35 @@
...
+    history._wantsHashChange = false;
+    history.location = {
+      pathname: '/outbound/22130600/po/powithspacetest%20',
+      href: 'http://example.com/outbound/22130600/po/powithspacetest%20'
+    };
+    assert.strictEqual(
+      history.getFragment(),
+      'outbound/22130600/po/powithspacetest '
+    );
   });
```
jgonggrijp, line 692: This is a useful test to add. Please keep it in your proposal.

jgonggrijp (changes requested):
Despite its small size, I found this proposal very tough to review. I stared at the `git blame` and at #4198, #1794, #1795, #1820, #967, #1156, #1426 and #2566. After a lot of thought, I think the fix is not quite right. Let me clarify with the following table about the internals of `getFragment`:

|  | `history.navigate` | `hashChange` | `popState` |
|---|---|---|---|
| source of `fragment` | user-provided | `this.getHash` | `this.getPath` |
| `fragment` still encoded? | yes (assumed) | yes | no |
| leading `#`/`/`? | maybe | no | no |
| *are* trailing spaces removed? (current situation) | yes | yes | yes |
| *should* trailing spaces be removed? | yes | yes | **no** |
| *will* trailing spaces be removed? (your proposal) | **no** | yes | no |

Your proposed change is effectively over-compensating by preserving the trailing spaces not only for the `popState` scenario, but also for the `history.navigate` scenario. This has prompted you to change existing tests that I think were actually justified or at least not problematic enough to warrant a breaking change.

I suggest that you make the following adjustments:

- **Keep** the new regression test for #4198.
- **Undo** the changes to the other tests.
- **Undo** the change to `routeStripper` and **drop** the new `trailingSpaceStripper` variable.
- **Undo** the change to `getHash`.
- In the `getFragment` function, **change** the line that reads `fragment = this.getPath();` to `return this.getPath();`.

The immediate return in `getFragment` will avoid applying `routeStripper` in the `popState` scenario. This preserves the spaces that should be preserved and skips stripping the leading `/` which `getPath` already removes, anyway.

PR #4302, opened 2026-04-06
Don't crash router on malformed percent-encoded params (#3440)

### LINKED ISSUE #3440: Invalid parameters
Backbone route parameters are decoded via `decodeURIComponent`. If the parameter is malformed (e.g. `%foo`) then a `URIError` is thrown. Since the user is in control of the url, this can happen even if all caution is taken to encode parameters. Is this truly an exceptional case in which things are so broken that execution should stop? If not, what should we return as a parameter?

BigBalli:
Fixes #3440.

## Problem

Route parameters are decoded with `decodeURIComponent`, which throws
`URIError` for malformed sequences (e.g. `/search/malformed%query`).
The exception bubbles out of `_extractParameters`, aborts route
matching entirely, and prevents the application from gracefully
handling a bad URL — which is especially painful for search-style
routes whose values often come from external sources.

## Fix

Wrap the `decodeURIComponent` call in a try/catch in
`_extractParameters`. On `URIError`, fall back to the raw, undecoded
parameter value so the route still matches and the handler still runs.
The application is then free to detect and recover (e.g. show a "bad
query" message). Other exceptions are re-thrown unchanged.

## Test

Added a regression test exercising `_extractParameters` directly with
a malformed `%query` parameter. `npm run lint` passes.


jgonggrijp:
Users can currently do this:

```js
try {
    history.start();
} catch (e) {
    if (e instanceof URIError) {
        // handle situation gracefully
    }
}
```

With your proposed change, this option is no longer there. Route matching continues, but the user will have to manually check each parameter for errors because there is no indication that anything went wrong.

I'm not saying we shouldn't do anything, but the current proposal is a breaking change and I'm not convinced it is an improvement. Could you give an example where the current behavior is problematic?

PR #4304, opened 2026-04-06
Make optional jQuery require bundler-friendly (#4184)

### LINKED ISSUE #4184: Cannot find module 'jquery'
Following the discussion from https://github.com/Rich-Harris/packd/issues/42#issuecomment-366759727.

In https://github.com/jashkenas/backbone/blob/master/backbone.js#L26 jQuery is require'd in a TryCatch Statement which is probably fine on Node but fails to bundle.

I would suggest to add `jquery` in the dependencies of Backbone, but I understand that's not what you want and I'm not sure why we should do.

BigBalli:
Fixes #4184.

## Problem

The CommonJS branch of the UMD wrapper does:

```js
var _ = require('underscore'), $;
try { $ = require('jquery'); } catch (e) {}
```

The `try/catch` is correct at runtime in Node.js, but Webpack, Rollup,
esbuild, packd and other bundlers statically analyze the literal
`require('jquery')` call before the try/catch ever runs. Users who
don't actually want jQuery still see their bundles fail with
`Cannot find module 'jquery'`.

## Fix

Indirect the optional require through a variable so the bundler's
static analysis does not follow it:

```js
var nodeRequire = typeof require === 'function' && require;
try { if (nodeRequire) $ = nodeRequire('jquery'); } catch (e) {}
```

This is the same trick used by `ws` and other libraries with optional
native deps. Runtime behavior is unchanged: the require still happens
in Node, the try/catch still swallows ENOENT, and `Backbone.$` is
still populated when jQuery is installed.

Also declare `jquery` as an optional peer dependency in `package.json`
so npm/yarn/pnpm document the relationship cleanly without forcing the
install:

```json
"peerDependencies": { "jquery": ">=1.11.0" },
"peerDependenciesMeta": { "jquery": { "optional": true } }
```

## Test

- `npm run lint` passes.
- Smoke-tested `require('./backbone.js')` in a fresh Node process with
  no jQuery installed: loads cleanly, `Backbone.VERSION === '1.6.1'`,
  `Backbone.$ === undefined` as expected.

### DIFF
--- a/backbone.js
+++ b/backbone.js
@@ -23,7 +23,13 @@
   // Next for Node.js or CommonJS. jQuery may not be needed as a module.
   } else if (typeof exports !== 'undefined') {
     var _ = require('underscore'), $;
-    try { $ = require('jquery'); } catch (e) {}
+    // Indirect the optional jQuery require through a variable so that
+    // Webpack, Rollup, esbuild, packd and friends do not try to resolve
+    // it at bundle time. The literal `require('jquery')` form was being
+    // statically analyzed by bundlers and failing the build for users
+    // who don't actually want jQuery (#4184).
+    var nodeRequire = typeof require === 'function' && require;
+    try { if (nodeRequire) $ = nodeRequire('jquery'); } catch (e) {}
     factory(root, exports, _, $);
 
   // Finally, as a browser global.
--- a/package.json
+++ b/package.json
@@ -15,6 +15,14 @@
   "dependencies": {
     "underscore": ">=1.8.3"
   },
+  "peerDependencies": {
+    "jquery": ">=1.11.0"
+  },
+  "peerDependenciesMeta": {
+    "jquery": {
+      "optional": true
+    }
+  },
   "devDependencies": {
     "coffeescript": "^2.7.0",
     "cpy-cli": "^5.0.0",


jgonggrijp:
Thank you for the awesome work in this pull request and the previous four!

I made a superficial glance over all five PRs and they look **very good** to me. I will be reviewing them in more detail (and almost certainly also merging them) over the next two weeks or so. Do you expect any merge conflicts between them, and if so, which order would you recommend?

jgonggrijp (commented):
I'm happy with the modification in the UMD wrapper. 👍 

Regarding the modification in the `package.json`, I have a question. I realize I suggested adding a `peerDependecy`, but now I wonder how much this might annoy users who are stuck with older versions of npm. I found these two threads about the introduction of `peerDependenciesMeta`:

https://github.com/npm/cli/issues/1247
https://github.com/npm/cli/pull/224

`peerDependenciesMeta` appears to have been introduced in npm v6 or v7. As of v7, npm apparently went "back" to installing peerDependencies by default. Which implies that some older versions of npm, which did not recognize `peerDependenciesMeta` yet, would *also* install peerDependencies by default. Do you know which versions of npm are affected by this and how problematic the resulting behavior could be?

I can look into this as well, but I thought I'd ask in case you already know the answer.

jgonggrijp:
@BigBalli gentle reminder that the next move on #4300, #4301, #4302 and #4303 is on you.

Regarding my question just above,

> Regarding the modification in the `package.json`, I have a question. I realize I suggested adding a `peerDependecy`, but now I wonder how much this might annoy users who are stuck with older versions of npm. I found these two threads about the introduction of `peerDependenciesMeta`:
> 
> [npm/cli#1247](https://github.com/npm/cli/issues/1247) [npm/cli#224](https://github.com/npm/cli/pull/224)
> 
> `peerDependenciesMeta` appears to have been introduced in npm v6 or v7. As of v7, npm apparently went "back" to installing peerDependencies by default. Which implies that some older versions of npm, which did not recognize `peerDependenciesMeta` yet, would _also_ install peerDependencies by default. Do you know which versions of npm are affected by this and how problematic the resulting behavior could be?
> 
> I can look into this as well, but I thought I'd ask in case you already know the answer.

I found the answer [here](https://nodejs.org/en/blog/npm/peer-dependencies):

> **UPDATE:** npm versions 1, 2, and 7 will automatically install peerDependencies if they are not explicitly depended upon higher in the dependency tree. For npm versions 3 through 6, you will receive a warning that the peerDependency is not installed instead.

The most recent release of npm v2, 2.15.12, was published 9 years ago and still gets 60k downloads a week, judging by [the npm version history](https://www.npmjs.com/package/npm?activeTab=versions). I'm going to ask for input on Mastodon.

**Update:** https://ieji.de/@juliangonggrijp/116482769712733866

jgonggrijp:
I have not yet received input from an actual npm v1/v2 user. It's not going to prevent them from using Backbone, the worst that can happen is that jQuery is installed while it wasn't before. So I'm going to merge this. Thanks for the fix, @BigBalli!

By the way, the next move on #4300, #4301, #4302 and #4303 is still on you.

Issue #4291, opened 2024-07-25
Aid the reader in understanding the code


jgonggrijp:
@jdittrich in https://github.com/jashkenas/backbone/issues/4290#issuecomment-2250480128:

> I wonder if code readability/learning/extending could be a focus in an update? That is not limited to code itself but also to documentation. It would not be a strong departure (though I find the event handling code difficult to understand sometimes) and might entail some potential compromizes where ease-of-understanding might be favored over performance, as well as architectural documentation (something between the line comments and the current documentation)

Me in https://github.com/jashkenas/backbone/issues/4290#issuecomment-2250776180:

> Could you elaborate a bit on how you would like to see this intermediate documentation?

@jdittrich in https://github.com/jashkenas/backbone/issues/4290#issuecomment-2250875377:

> > intermediate documentation
> 
> …for me would usually involve
> 
> * What is already in ["getting started"](http://backbonejs.org/#Getting-started)
> 
> * And then, for the larger code modules how to read and understand the respective code. For event handling (since that is the part that I know best) something like: What are the contracts/interfaces for an [`__Api`-Function/Reducer](http://backbonejs.org/docs/backbone.html#section-32), how is `this._events` structured, which "private" attributes are added to an object that uses the Events mixin (These are all focussed on the current way events work)
>       
> * JSDoc might help with the contracts/interfaces?
> 
> 
> Basically, something that helps understanding how different parts of the code work together and how to read the code.

I suspect it would already help a lot if there were diagrams that visualize the relations between the different parts of the code. Those could be included right inside the Docco [annotated source](https://backbonejs.org/docs/backbone.html), because it supports Markdown inside the comments.

Comments welcome, as well as pull requests.

abouchard11:
Hi @jgonggrijp — I’d be interested in taking a first pass at this.

To keep it small and reviewable, I’m considering a compact architecture note focused on the Events implementation, covering:

- how the public APIs delegate through `eventsApi`;
- the roles of `_events`, `_listeners`, and `_listeningTo`;
- how `on`/`off` differ from `listenTo`/`stopListening`;
- one small diagram showing the listener/listenee relationships.

I would keep this documentation-only, with no runtime changes. Would that scope be useful, and would you prefer it in the annotated-source comments as suggested here or in a separate documentation section? I’m happy to follow your preferred direction before writing the patch.

jgonggrijp:
@abouchard11 Thanks for the heads-up, I appreciate it.

[...]

Issue #4290, opened 2024-05-17
If you could change Backbone at will, how would you?


jgonggrijp:
Sooner or later, there will be a Backbone 2.0. Major releases are disruptive for users, so I prefer to make them as infrequently as possible. But when we do one anyway, we might as well use it to its full potential.

This is a "wild ideas" ticket, where you can suggest any backwards-incompatible changes that you would like to see in a future release of Backbone. Please keep it brief, but do explain *why* you would like to see the change(s) that you suggest.

You are also welcome to highlight anything that you want to stay the same. Again, please with reasons.

Some rules of the game:

- State your reason for wanting the change (or non-change)! It does not need to be a "good" or "smart" reason.
- All ideas are valid, bearing in mind that there are always tradeoffs. Please keep it very civil and do not enter competition with other suggesters over whose idea should "win".
- The fact that an idea is mentioned here does not say anything about its likelyhood to make it into Backbone, even if that idea seems popular. In the end, the decisions are made by the people who do the work, with me and @jashkenas having the final word because we are publishing the package and we need to assure its quality.

The following prior tickets are somewhat related: #4274, #4245, #3560.

jdittrich:
_What to keep:_ It seems that backbone is the only MVC-Framework for JS that is still updated, so I would not change anything about that paradigm.

_What to keep/Improve:_ Use of ESModules without introducing the Backbone global or a build tool #4274

_What to improve:_ Not a huge problem for me, but using modern  JavaScript (not cutting edge, just anything that works since a couple of years) and, with that, make underscore optional? Many functional-style methods has been added to JavaScript arrays by now, so maybe this is sufficient? Similarly Map and Set might be helpful in some occasions. _Reasons:_ Size and code-readability for people not familiar with underscore.

jashkenas:
☝️ I endorse these suggested changes for a Backbone 2.0.

jgonggrijp:
So far, we have one response, which only suggests evolutionary changes, rather than anything radical. Would it be safe to conclude that Backbone users just want the library to be updated with recent JavaScript and otherwise stay mostly the same? If anyone thinks otherwise, please don't hesitate to respond.

jdittrich:
Not radical, either but:

* A standard way to manage child views might be great
[...]

PR #4307, opened 2026-08-17
docs: explain Events internals

abouchard11:
## Summary

- explain how direct `on` subscriptions differ from managed `listenTo` relationships
- document the private `_events`, `_listeningTo`, and `_listeners` bookkeeping
- describe the shared `eventsApi` reducer contract and interoperability path
- add an accessible diagram of the listener, listenee, `Listening` record, and event handler references

## Why

Issue #4291 asks for intermediate architectural documentation that helps readers understand how the Events implementation fits together. This keeps the explanation beside the code so it appears in the annotated source, while making no runtime changes.

Addresses #4291.

## Validation

- `eslint backbone.js test/*.js modules/*.js`
- generated the Docco annotated source in a temporary directory and verified the new section and SVG reference
- rendered the SVG at the annotated source's 350px and 525px column widths to check that it is well formed and legible
- `git diff --check`


abouchard11:
Hi @jgonggrijp and @jdittrich — following up on #4291. This documentation-only PR implements the scope discussed there: the Events API flow, _events / _listeners / _listeningTo bookkeeping, interoperability behavior, and an accessible diagram in the annotated source. It makes no runtime changes, and the local validation listed in the PR description passed.

The Test workflow currently shows action_required with no jobs created, waiting on approval for a workflow from the fork. @jdittrich, if you have bandwidth, your review would be appreciated as discussed on the issue, @jgonggrijp, could you approve the CI run when convenient?  I can adjust the placement if you like.

docs/images/events-listening.svg
jgonggrijp, line 1: Since this image needs to fit in a somewhat narrow text column, I suggest giving it a narrower profile as well. In other words, taller and less wide. The text will be larger in that way.

backbone.js
```diff
@@ -87,6 +87,45 @@
   //     object.on('expand', function(){ alert('expanded'); });
   //     object.trigger('expand');
   //
+
+  // Reading the Events implementation
+  // ---------------------------------
```
jgonggrijp: This renders as a `<h2>`, but it needs to be a `<h3>` as it is a subsection of `Backbone.Events`.

```suggestion
  // ### Reading the Events implementation
```

jgonggrijp (commented):
This is a preliminary review in which I focus on correctness. The text and the image look correct to me, I only have comments on the form for now. I do think the clarity and consistency could improve, but I will defer to @jdittrich first.

docs/images/events-listening.svg
jdittrich, line 1: When reading the diagram without having read the text, I would wonder where "Listening record" stores the reference to "Listenee" (or do they reference each other? In which case they might need a dependency arrow?)

backbone.js
abouchard11: @jgonggrijp Done. Every other `---` underline in backbone.js marks a top-level section, so this was rendering as a sibling of Backbone.Model rather than part of Backbone.Events. Regenerated the docs and confirmed it now comes out as an h3 nested under Backbone.Events.

docs/images/events-listening.svg
abouchard11: @jgonggrijp Redrawn at 480x830, down from 700x590. docco.css sets the comment column to 350px, or 525px on wide viewports, so the old landscape image was being scaled to roughly half and the label text went with it. At the new size it renders about 1:1 in the wide column. Stacking the boxes vertically also made room for the back-references @jdittrich raised.

abouchard11: @jdittrich They reference each other. The Listening constructor stores both sides:

```js
var Listening = function(listener, obj) {
  this.listener = listener;
  this.obj = obj;
```

The listener reaches the record through `_listeningTo` and the listenee through `_listeners`, while the record holds `listener` and `obj` pointing back at each. I've added arrows in both directions so that reads off the diagram on its own, and added a line to the prose saying the same.

jgonggrijp:
@jdittrich, to what extend do you feel that the latest version of this PR addresses your comment in https://github.com/jashkenas/backbone/issues/4290#issuecomment-2250480128? What is already good and what could still improve? Please feel welcome to include points that are out of the scope of the current PR, if that helps you clarify what you are looking for.

docs/images/events-listening.svg
jdittrich, line 1: Thanks! I have not read the code in a while (the issue is a bit older when I was more active in it), but if I remember correctly, callback, context, ctx, listening are properties on Event handler – maybe it would make sense to format it this way? Either like a type annotation or with a "_properties_: ctx, callback...." section?

jdittrich:
I think this is a big improvement! I would say let's merge it if everyone agrees it is correct and helpful, and lets open new issues if the need arises.

docs/images/events-listening.svg
abouchard11: @jdittrich They are properties of the handler object — backbone.js pushes `{callback, context, ctx, listening}` onto `_events[name]`.

I've formatted them as a two-column field list rather than a labelled properties section, so it matches the Listening record box above, which already lists its own fields that way. That keeps one visual grammar for objects-with-fields across the diagram. Happy to switch to an explicit `properties:` label if you prefer yours.

Separately, while checking: `context` and `ctx` are easy to confuse. `context` is what the caller passed and is what `off` matches on; `ctx` is the receiver the callback is actually invoked with, defaulting to the object itself. That distinction isn't in the text — worth a line, or more detail than this section needs?

jdittrich: context ctx confusion: Makes sense to clarify.

jgonggrijp, line 1: I would like a few more tweaks to this diagram:

- [x] In the caption of the Listenee box, also mention that it calls on/off, for parity with the Listener box.
- [x] Use thicker lines to more closely resemble the style of `docs/images/intro-model-view.svg`.
- [x] Use the fonts in `docs/public/fonts` to match the surrounding text.
- [x] After doing all of the above, fiddle a bit with the width to align the text size with the surrounding text.

backbone.js
```diff
@@ -87,6 +87,55 @@
...
+  // It holds `listener` and `obj` references back to the two objects, and each
+  // handler created through `listenTo` points back to it through `listening`.
+  // The record counts those handlers and, when the last one is unbound,
+  // deletes its entries in the listener's `_listeningTo` and the listenee's
+  // `_listeners`. If the listenee implements another events API instead of
+  // Backbone.Events, the record's `interop` flag stays on and the record keeps
+  // its own copy of the callbacks instead of a count, so that `stopListening`
+  // retains the same public behavior.
+  //
+  // Most public methods below share `eventsApi`. It accepts a single event
```
jgonggrijp, line 131: Maybe also mention that `eventsApi` returns a standardized `{eventName: iterateeResult[]}` map. Not in the way I formulated it though!

backbone.js
```diff
@@ -87,6 +87,55 @@
...
+  // name, a space-separated list of names, or an event map, and calls a small
+  // reducer once for each event name. The reducer receives `(events, name,
+  // callback, options)` and returns the next `events` accumulator: `onApi`
+  // adds handlers, `offApi` removes them, `onceMap` creates one-shot wrappers,
+  // and `triggerApi` fires them. Handling the different forms in one place
+  // lets each reducer deal with a single event at a time.
+  //
```
jgonggrijp, line 138: Please add another heading here to help the reader recognize the transition to the actual implementation.

```suggestion
  //
  // ### Implementation
```

jgonggrijp (changes requested):
Thanks, this is starting to look very good. Just a few more tweaks.

docs/images/events-listening.svg
abouchard11: All four done. Lines are thicker — 4px boxes, 5px connectors, up from 2px. For the fonts: the SVG loads through an , so it can't reach docco.css. I subsetted aller-light and aller-bold to the glyphs the diagram uses and embedded them as a data URI, which adds about 9KB and keeps the text selectable. Happy to convert the text to outlines like the intro-*.svg exports instead if you'd prefer no font data in the file. Width is now 450px, which is the annotation column at the 1025px breakpoint, so it renders 1:1 there and the caption text matches the surrounding text. Between 481px and 1024px the column is 324px and the diagram scales down with it, so the labels are smaller than the body text there — I don't think one fixed width can match both, but say if you'd rather I tune for the narrower column.

backbone.js
abouchard11: Added. I described it per reducer rather than as one standardized shape, since `onApi`, `offApi` and `triggerApi` thread the object's `_events` through while `onceMap` builds a fresh `{name: wrapper}` map for `once` to hand to `on`.

backbone.js
abouchard11: Added in ab09de6. It sits directly above `var Events = Backbone.Events = {}`, after the new paragraph on what `eventsApi` returns.

docs/images/events-listening.svg
jgonggrijp: Ah, I didn't realize there was a breakpoint. Your current sizing looks good to me at both widths.

jgonggrijp, line 1: This looks almost perfect to me, just two nitpicks:

- [x] The bottom padding of the boxes is less than the top and left padding. I think it will look better if you increase the bottom padding to equal the top and right.
- [x] The arrowheads look rather clunky now. If you look at the model-view diagram, you can see that the arrowheads are just slightly wider than the lines. It would be nice if you could match those proportions here.

backbone.js
```diff
@@ -87,6 +87,63 @@
...
+  // handler created through `listenTo` points back to it through `listening`.
+  // The record counts those handlers and, when the last one is unbound,
+  // deletes its entries in the listener's `_listeningTo` and the listenee's
+  // `_listeners`. If the listenee implements another events API instead of
+  // Backbone.Events, the record's `interop` flag stays on and the record keeps
+  // its own copy of the callbacks instead of a count, so that `stopListening`
+  // retains the same public behavior.
+  //
+  // Most public methods below share `eventsApi`. It accepts a single event
+  // name, a space-separated list of names, or an event map, and calls a small
```
jgonggrijp: Let's omit "small". `offApi` actually looks a little bit intimidating.

```suggestion
  // name, a space-separated list of names, or an event map, and calls a
```

backbone.js
```diff
@@ -87,6 +87,63 @@
...
+  // adds handlers, `offApi` removes them, `onceMap` creates one-shot wrappers,
+  // and `triggerApi` fires them. Handling the different forms in one place
+  // lets each reducer deal with a single event at a time.
+  //
+  // Because each call's return value becomes the next call's accumulator,
+  // `eventsApi` returns the last one. `onApi`, `offApi`, and `triggerApi`
+  // thread the object's `_events` through, so `on` and `off` assign the
+  // result straight back to `this._events`; `onceMap` instead fills an empty
+  // object one `{name: wrapper}` entry at a time, which `once` then passes to
+  // `on` as an event map.
```
jgonggrijp, line 144: Maybe this is going too much into the details in advance of the actual implementation. What do you think, @jdittrich?

jgonggrijp (changes requested):
We're getting there. It's really just dotting the i's and crossing the t's at this point.

docs/images/events-listening.svg
abouchard11: Both done. Bottom padding now matches the top on all four boxes. For the arrowheads I measured the model-view diagram: the head is about 2.1x the line width and 2.5x long, and the new marker uses those proportions.

backbone.js
abouchard11: Done.

abouchard11:
@jashkenas — I believe the requested updates are all addressed now. @jdittrich, since your earlier question about the level of implementation detail didn’t get a response, does the current version feel appropriate? If so, could you merge when you have a chance? Thanks!

